PR #4, opened 2022-02-25
add a condition code that is always true.

arvidn:
The actual integer value can be debated, I don't have a strong opinion.
I also took the opportunity to assert `cost` in the existing unit tests (in addition to the new one).

The change to enable negative op-codes is a pretty big portion of this patch, and maybe it's not worth it. The patch would be smaller if we pick a positive no-op code.

### DIFF
--- a/src/gen/conditions.rs
+++ b/src/gen/conditions.rs
@@ -3,7 +3,7 @@ use super::condition_sanitizers::{
     parse_amount, parse_height, parse_seconds, sanitize_announce_msg, sanitize_hash,
 };
 use super::opcodes::{
-    parse_opcode, ConditionOpcode, AGG_SIG_COST, AGG_SIG_ME, AGG_SIG_UNSAFE,
+    parse_opcode, ConditionOpcode, AGG_SIG_COST, AGG_SIG_ME, AGG_SIG_UNSAFE, ALWAYS_TRUE,
     ASSERT_COIN_ANNOUNCEMENT, ASSERT_HEIGHT_ABSOLUTE, ASSERT_HEIGHT_RELATIVE, ASSERT_MY_AMOUNT,
     ASSERT_MY_COIN_ID, ASSERT_MY_PARENT_ID, ASSERT_MY_PUZZLEHASH, ASSERT_PUZZLE_ANNOUNCEMENT,
     ASSERT_SECONDS_ABSOLUTE, ASSERT_SECONDS_RELATIVE, CREATE_COIN, CREATE_COIN_ANNOUNCEMENT,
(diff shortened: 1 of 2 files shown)


src/gen/conditions.rs
```diff
@@ -304,6 +304,12 @@ fn parse_args(
                 flags,
             )?))
         }
+        ALWAYS_TRUE => {
+            if (flags & STRICT_ARGS_COUNT) != 0 {
+                check_nil(a, rest(a, c)?)?;
```
richardkiss, line 309: What does this do?

src/gen/opcodes.rs
```diff
@@ -33,6 +33,9 @@ pub const ASSERT_SECONDS_ABSOLUTE: ConditionOpcode = 81;
 pub const ASSERT_HEIGHT_RELATIVE: ConditionOpcode = 82;
 pub const ASSERT_HEIGHT_ABSOLUTE: ConditionOpcode = 83;
 
+// no-op condition
+pub const ALWAYS_TRUE: ConditionOpcode = 90;
```
richardkiss, line 37: How about using `0`?

richardkiss: Or if that's not possible, use `1` (so it looks like `q`).

richardkiss (changes requested):
I request change `90` to `0`.

src/gen/opcodes.rs
arvidn: sure. that's a single byte string containing `0`, right? as all opcodes are currently one byte.

src/gen/conditions.rs
arvidn: this still requires the condition to have 0 arguments (in mempool mode). I'm still not sure how this always-true condition is meant to be used, so being strict on the argument count is a conservative approach. But perhaps the whole point of this is to allow any arguments, and just always ignore them.

arvidn: in either case, there should be a few test cases covering the arguments

src/gen/opcodes.rs
richardkiss: I was actually thinking the `0` number, ie. null, the empty string. If you're using a "first byte" decoding rather than a "convert to clvm int" mechanism, we can use `1` (ie. the single byte `0x01`).

src/gen/conditions.rs
richardkiss: Yeah, just ignore any arguments. The idea is to allow for "remarks".

richardkiss (changes requested):
Let's go with `1` instead of `0` to eliminate potential confusion.

richardkiss (approved):

richardkiss (approved):

PR #7, opened 2022-03-01
Get `chia_rs` building with stripped-down `clvmr`.

richardkiss:


wheel/src/api.rs
```diff
@@ -17,6 +17,8 @@ pub const MEMPOOL_MODE: u32 =
 pub fn chia_rs(_py: Python, m: &PyModule) -> PyResult<()> {
     m.add_function(wrap_pyfunction!(a_test_function, m)?)?;
     m.add_function(wrap_pyfunction!(run_generator, m)?)?;
+    m.add_class::<PySpendBundleConditions>()?;
+    m.add_class::<PySpend>()?;
```
richardkiss, line 21: I believe that you don't actually need to do this. There's not much harm in doing it, but there's also not much benefit since they don't contain any static methods or class methods.

arvidn: surely accessing members end up calling into rust code though, that needs to be exported somehow, no?

arvidn: otherwise the data would have to be copied into python when returned

richardkiss: The idea is that you can return `#[pyclass]` objects to python even if you don't export them in the module.

PR #8, opened 2022-03-09
additional test for conditions

arvidn:
checking that extra arguments are ignored


richardkiss (approved):
Didn't look too close, but more tests is always good.

arvidn:
these tests are asking to be refactored to be more compact and have less duplicated boilerplate

PR #9, opened 2022-03-10
in mempool mode, also check the number of arguments to conditions

arvidn:
probably best reviewed with whitespace ignored.

### DIFF
--- a/src/gen/conditions.rs
+++ b/src/gen/conditions.rs
@@ -14,6 +14,7 @@ use super::sanitize_int::sanitize_uint;
 use super::validation_error::{first, next, rest, ErrorCode, ValidationErr};
 use crate::gen::flags::COND_ARGS_NIL;
 use crate::gen::flags::NO_UNKNOWN_CONDS;
+use crate::gen::flags::STRICT_ARGS_COUNT;
 use crate::gen::validation_error::check_nil;
 use clvmr::allocator::{Allocator, NodePtr, SExp};
 use clvmr::cost::Cost;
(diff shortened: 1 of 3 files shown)


src/gen/conditions.rs
```diff
@@ -143,21 +144,29 @@ fn parse_args(
...
-                        // pull out the first item (param)
-                        if let SExp::Atom(b) = a.sexp(param) {
-                            if a.buf(&b).len() <= 32 {
-                                return Ok(Condition::CreateCoin(puzzle_hash, amount, param));
-                            }
+            // ignored. (unless we're in mempool mode, and the STRICT_ARGS_COUNT
+            // flag is set)
+
+            // we always expect one more item, even if it's the zero-terminator
+            c = rest(a, c)?;
```
richardkiss, line 151: This seems like a new error case. If it's intentional, that's okay. I guess it means a condition must be a list/can't be an atom?

richardkiss (approved):
I don't completely get it, but I'm assuming the tests cover it.

src/gen/conditions.rs
arvidn: I don't think so.

`c` at the start of this match-block must be a cons-box where "first" is  puzzle-hash. Then `c` is updated to be the rest, which again must be a cons-box where "first" is the amount. Since `first()` already succeeded, `rest()` must also succeed without an error here. Whether we get back an atom, a nil terminator or another cons-box however, we don't know (yet).

arvidn:
yeah, the tests need a bit of an overhaul to be more succinct and easier to read.
The motivation for the patch is to enable checking that condition argument lists are terminated after the arguments we expect, in mempool mode.

PR #10, opened 2022-03-16
make the python module also export key functionality from clvm_rs

arvidn:
This makes it possible for chia-blockchain to only depend on the chia_rs module.

There's some duplicated code with `clvm_rs/wheel` here, I can't think of a way to avoid it, but I think it's still worth it.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -57,3 +57,10 @@ jobs:
         . ./activate
         cd tests
         ./test-generators.py
+
+    - name: Run cost checks
+      run: |
+        . ./activate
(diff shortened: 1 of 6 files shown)


richardkiss (approved):
Looks good. Do you want to copy the tests over too?

richardkiss:
What would you think about putting the `clvm_rs` stuff into a submodule? So top level module is `chia_rs`, and clvm-specific stuff is in `chia_rs.clvm`. I'm sure `pyo3` supports it.

arvidn:
regarding the tests; yes, it would probably be a good idea to move those over too

PR #12, opened 2022-05-04
function to compute a chia merkle tree root

arvidn:
It treats the leaves as a list to be sorted by a recursive radix sort and implements the hashing along the return path.
This gives a speed-up of about 44 compared to the existing python implementation.

For chia additions, we insert both puzzle hashes and their coins, so 2 merkle values per addition (assuming every addition use a different puzzle hash). With full blocks having ~2000 additions, 4000 merkle values represents a full merkle set.

Tested on an M1

Generating 500 merkle sets of 4000 values:

```
python time: 46.96s
rust time: 1.06s
```

Generating 500 merkle sets with only 100 values:

```
python time: 0.57s
rust time: 0.03s
```

### DIFF
--- a/src/merkle_set.rs
+++ b/src/merkle_set.rs
@@ -63,7 +63,7 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
     let mut left: i32 = 0;
     let mut right = range.len() as i32 - 1;
 
-    // move 0 bits to the left, and 1 bits to the rigth
+    // move 0 bits to the left, and 1 bits to the right
     while left <= right {
         let left_bit = get_bit(&range[left as usize], depth);
         let right_bit = get_bit(&range[right as usize], depth);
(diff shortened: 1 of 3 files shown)


src/merkle_set.rs
```diff
@@ -0,0 +1,554 @@
...
+            hash(left_type, right_type, &left_hash, &right_hash),
+            node_type,
+        )
+    }
+}
+
+pub fn compute_merkle_root(leafs: &mut [[u8; 32]]) -> [u8; 32] {
+    // There's a special case for empty sets
+    if leafs.is_empty() {
+        return [
```
arvidn, line 139: this should return `BLANK`

src/merkle_set.rs
```diff
@@ -0,0 +1,554 @@
...
+
+    // first sort the range based on the bit at "depth" (starting with the most
+    // significant bit). It also sorts the two resulting ranges recursively by
+    // the next bit. The return value is the SHA256 digest of the resulting
+    // merkle tree. Any node that only has a children on one side, is a no-op,
+    // where that child's hash is forwarded up the tree.
+    let mut left: i32 = 0;
+    let mut right = range.len() as i32 - 1;
+
+    // move 0 bits to the left, and 1 bits to the rigth
```
arvidn, line 66: typo: "rigth"

arvidn:
> Pretty sure this isn't consensus-critical.

The block header has a merkle root for additions and removals, so pretty sure it's consensus critical

richardkiss (approved):
I'm not really familiar enough with the original code to know if this is the same. I will testify that it seems to be pretty straightforward recursion, so as long as it matches on the edge cases (and the python has the same edge cases), it's likely to be compatible.

arvidn:
mariano suggested renaming the function to `compute_merkle_set_root()` to make it more clear what kind of root it is. So I will update this PR.

PR #14, opened 2022-05-10
use ByteN types instead of plain arrays (and simplify those types a bit)

arvidn:


src/streamable/mod.rs
```diff
@@ -1,4 +1,4 @@
 pub mod bytes;
-pub mod coin;
-pub mod fullblock;
-pub mod program;
+// pub mod coin;
+// pub mod fullblock;
+// pub mod program;
```
arvidn, line 4: I don't want to pull these in until they're done

tests/run_gen.py
```diff
@@ -81,7 +81,7 @@ def print_spend_bundle_conditions(result) -> str:
         if s.seconds_relative > 0:
             ret += f"  ASSERT_SECONDS_RELATIVE {s.seconds_relative}\n"
         for a in sorted(s.create_coin):
-            if len(a[2]) > 0:
+            if a[2] is not None and len(a[2]) > 0:
```
arvidn, line 84: this API changed back now, to have the hint be `None` if there is none

wheel/src/run_generator.rs
```diff
@@ -42,52 +42,48 @@ pub struct PySpendBundleConditions {
...
     for c in spend.create_coin {
         create_coin.push((
-            PyBytes::new(py, &c.puzzle_hash).into(),
+            c.puzzle_hash,
             c.amount,
-            node_to_pybytes(py, a, c.hint),
+            if c.hint != a.null() {
+                Some(a.atom(c.hint).into())
+            } else {
+                None
```
arvidn, line 63: This change from empty bytes to `None` is also reflected in the test here: https://github.com/Chia-Network/chia_rs/pull/14/files#diff-f584716cd024cb6e7ff78c2fbc0136c01f9742ab26c49f24642262b5b61c68beR84

wheel/src/run_generator.rs
```diff
@@ -42,52 +42,48 @@ pub struct PySpendBundleConditions {
...
         agg_sig_me: agg_sigs,
     }
 }
 
-fn convert_spend_bundle_conds(
-    py: Python,
-    a: &Allocator,
-    sb: SpendBundleConditions,
-) -> PySpendBundleConditions {
+fn convert_spend_bundle_conds(a: &Allocator, sb: SpendBundleConditions) -> PySpendBundleConditions {
```
arvidn, line 78: these type no longer contain python-allocated types (like `PyBytes`), so we don't need a `Python` object to allocate them

src/gen/coin_id.rs
```diff
@@ -6,12 +7,12 @@ pub fn compute_coin_id(
     parent_id: NodePtr,
     puzzle_hash: NodePtr,
     amount: &[u8],
-) -> [u8; 32] {
+) -> Bytes32 {
```
richardkiss, line 10: I think it's better to avoid non-native types where possible. Let the caller of this function use `.into`.

richardkiss: (This should be weighed against lots of `.into` calls that might otherwise not be needed.)

arvidn: what's your rationale for that? I would think having even more specific types (like, `CoinId`, `PuzzleHash` etc.) would improve type safety.

I suppose the main purpose of `BytesN` types is to provide python conversions, and this function isn't exposed to python, so technically doesn't need it.

It's called exactly once in production code (to construct an `Arc<Bytes32>`) + a few times in the unit test. Does this preference also extend to the `coin_id` member of the `Spend` class?

richardkiss: I'm pretty sure `pyo3` knows how to convert `&[u8]` to `PyBytes` already, so re-implementing for `Bytes32` probably isn't even necessary. My concern is a very mild one about polluting namespace and dependency-space here with a custom `Bytes32` type in this file when it probably isn't strictly necessary.

richardkiss (approved):

src/gen/coin_id.rs
arvidn: It's necessary for supporting serialization, to have a type to implement the `Serialize` and `Deserialize` traits. This specific return value isn't covered by that either, but I believe the members it initialized are.

PR #15, opened 2022-05-11
add Chia Streamable serialization/deserialization support via serde

arvidn:
Make the PySpend and PySpendBundleConditions serializeable in this format.

This shaves off some CPU time from block validation in the main thread, where we deserialize the `PySpendBundleConditions` object. Those objects are sent across the process boundary from the workers.

This brings the time spent in `pre_validate_blocks_multiprocessing` from 7.47% -> 0.94%. See profiles below.

Benchmark before this patch:
slow-batch-00448

Benchmark after this patch:

slow-batch-00448


wheel/src/run_generator.rs
```diff
@@ -28,8 +33,50 @@ pub struct PySpend {
...
+            height_relative,
+            seconds_relative,
+            create_coin,
+            agg_sig_me,
+        }
+    }
+}
+
+#[pyproto]
+impl PyObjectProtocol for PySpend {
```
arvidn, line 59: this is duplicated twice, and I suspect will be duplicated more times going forward. I'm kind of surprised pyo3 doesn't have a macro to enable the obvious `__str__` and `__repr__` implementations as well as the comparisons.

I considered writing a derive macro for this, but as I understand it, that would have to live in a separate crate, which seemed like a steep hurdle right now.

wheel/src/run_generator.rs
```diff
@@ -47,15 +94,77 @@ pub struct PySpendBundleConditions {
...
+    }
+
+    #[staticmethod]
+    pub fn from_bytes(blob: &[u8]) -> PyResult<Self> {
+        py_from_bytes::<Self>(blob)
+    }
+
+    // returns the type as well as the number of bytes read from the buffer
+    #[staticmethod]
+    pub fn parse_rust(blob: &[u8]) -> PyResult<(Self, u32)> {
```
arvidn, line 125: one problem I have here is that this won't accept a `memoryview` object. So on the python side, I have to force the `memoryview` to `bytes` which isn't great. Any ideas?

richardkiss: My understanding is the `memoryview` api is not available when we use the `abi` bindings, which allow us to create just a single wheel that works across python 3.7/8/9/10. If we go to version-specific wheels, then we get `memoryview` back (I guess that API isn't guaranteed to remain the same across versions). So it's a big trade-off in build complexity vs performance. My guess is copying bytes one extra time won't be too much of a performance hit as long as the users of the rust wheel understand that accessing rust fields is expensive so don't do it repeatedly.

src/streamable/deserialize.rs
```diff
@@ -0,0 +1,519 @@
+use super::error::{Error, Result};
+use std::convert::TryInto;
+
+pub struct ChiaDeserializer<'storage> {
```
richardkiss, line 4: I think modules like this are traditionally named `ser.rs`.

richardkiss: As you know, I forked `bincode` and made some tweaks that let it support "chia" serialization (where `usize` is encoded as `u32` instead of `u64`). I think an ideal future would be getting it work a bit more nicely and contributing it back to `bincode` so we can use it instead of rolling our own. But for sure, this is fine for now. Maybe we put it in a separate module called `chia_bincode` or `bincode_chia` or `bincode_u32` or something?

src/streamable/serialize.rs
```diff
@@ -0,0 +1,491 @@
+use super::error::{Error, Result};
```
richardkiss, line 1: How much of this is ported from `bincode` vs home-grown? If a lot is borrowed, maybe a comment to this effect? "Based on <url to source>"

wheel/src/run_generator.rs
richardkiss: By the way, since you're moving both serialization and deserialization from python to rust, you don't actually need the format to be backward compatible. You could really use straight `bincode` here rather than the weird `chia_bincode`. [Of course, we'll need the `chia_bincode` to parse blockchain structures that can't change, so maybe we may as well use it everywhere.]

richardkiss (approved):
Seems great! Not sure if any of the ideas I wrote are worth pursuing. If not, feel free to merge as is.

wheel/src/run_generator.rs
arvidn: > Another option is to allow an optional "cursor" argument, defaulting to 0, so you can use the same bytes objects across multiple parse calls

Good point. I haven't spent too much time thinking of clever ways to integrate this with the existing `Streamable` framework. Right now I just pull out the buffer from `BytesIO` and then advance its cursor when rust is done parsing. I'm not sure it's worth the trouble to try to pass something more stream-like down, as that would be quite complicated. Especially since we only ever use it with `BytesIO` anyway.

I think copying the bytes each time is fine for now. At least it doesn't show up in the profiles. Once everything else is faster it might and we could address it then. (e.g. `FullBlock` objects may be large enough for it to matter).

> By the way, since you're moving both serialization and deserialization from python to rust, you don't actually need the format to be backward compatible

Some of the other expensive objects to serialize and deserialize that I'm hoping to address down the road do need to be backwards compatible. like `FullBlock` and `CoinRecord`.

src/streamable/deserialize.rs
arvidn: The reason I went for a from-scratch implementation is that `bincode` supports a *lot* of options (most of them aren't relevant to us), and is quite a lot of code. It's a lot easier (I think) to review and be confident in the simpler implementation I made, that it does in fact implement our protocol.

`bincode` also supports constructs that `Streamable` doesn't support, and I think it would be risky to allow the opportunity to serialize such constructs, into invalid `Streamable` format. My implementation fails with `NotSupported` instead. In an ideal world, it would be a compile-time error, but I can't think of a way to do that, since traits must be fully implemented, even parts that are never used.

src/streamable/serialize.rs
arvidn: bincode was my main source of inspiration, for sure. I'll add a comment.

arvidn:
here's the chia-blockchain part of this: https://github.com/Chia-Network/chia-blockchain/pull/11503

src/streamable/de.rs
```diff
@@ -116,11 +116,11 @@ impl<'de, 'a> serde::Deserializer<'de> for &'a mut ChiaDeserializer<'de> {
         visitor.visit_i8(self.read_u8()? as i8)
     }
 
-    fn deserialize_unit<V>(self, visitor: V) -> Result<V::Value>
+    fn deserialize_unit<V>(self, _visitor: V) -> Result<V::Value>
     where
         V: serde::de::Visitor<'de>,
     {
-        visitor.visit_unit()
+        Err(Error::NotSupported)
```
arvidn, line 123: this was an inconsistency. The serializer doesn't support this, so I made the deserializer also fail (and the type blow as well)

PR #16, opened 2022-05-12
Python 3.11

altendky:
Draft for:
- [x] Wait for https://github.com/Chia-Network/chia_rs/pull/30
- [ ] Needs a blspy wheel for 3.11


arvidn:
iirc, there was also some other trick @richardkiss had to pull to get `maturin` to work on windows.

richardkiss:
Note that a single binary is built for all python versions for a given platform. Because of this, I don't think it's actually necessary to check all versions of python.

richardkiss:
As for the Windows-specific stuff, I'd recommend looking at how the windows build works elsewhere.

Ideally, we should serialize these process – create a wheel, export it as an artifact, then import that artifact by dependent testing and benchmark actions. This would let us build the wheel just once, and also test just that particular binary.

It would take a Github Actions wizard to do this for sure, but it's the right way.

altendky:
I've done exactly that sequence in other projects and have commented on it at various points here, but I haven't taken it on myself to actually switch over here.  Let alone for a dozen or whatever repos...

https://github.com/altendky/ciborg/actions/runs/363450564 is an example.
image

As to testing more Python versions, I personally default to testing the matrix we claim to support, until it gets onerous.  I think that our support repos like this are low enough traffic and the tests are quick enough that the CI runtime overhead is negligible.  In some sense there is added overhead in the CI setup complexity, but if we maintain it then we get the benefit of consistency across projects which is an alternative benefit.

richardkiss:
> I've done exactly that sequence in other projects and have commented on it at various points here, but I haven't taken it on myself to actually switch over here. Let alone for a dozen or whatever repos...
> 
> https://github.com/altendky/ciborg/actions/runs/363450564 is an example. image
> 
> As to testing more Python versions, I personally default to testing the matrix we claim to support, until it gets onerous. I think that our support repos like this are low enough traffic and the tests are quick enough that the CI runtime overhead is negligible. In some sense there is added overhead in the CI setup complexity, but if we maintain it then we get the benefit of consistency across projects which is an alternative benefit.

A comment here... I think ideally you want to build the `sdist` first, and use *just* that as an input to build each wheel. That way, we have a high confidence that the sdist is being built and is legit.

.github/workflows/benchmark.yml
```diff
@@ -52,31 +57,32 @@ jobs:
...
 
     - name: python tests
       run: |
-        . ./activate
         pytest tests
 
     - name: test generators
+      if: matrix.os == 'ubuntu-latest'
+      # TODO: This should happen everywhere but the benchmarking part is not passing
+      #       on all platforms so this can be revisited separately.
```
arvidn, line 76: we should probably run this on our dedicated benchmarking runners

arvidn (approved):

.github/workflows/benchmark.yml
altendky: The benchmark aspect I suspect should be made optional, probably default off, and it should be run with that on the benchmarks runners, sure.  But I wouldn't want to lose regular test coverage of it and would like coverage across platforms to be added here.  Anyways, I'm good with leaving this for another PR.  My next round with our workflows might be some sort of reusable library mechanism to help with the cross-repo duplication and remaining variances.

cmmarslender (approved):

PR #17, opened 2022-05-13
implement PyObjectProtocol and Streamable hooks as a derive macro

arvidn:
this factors out the implementation of `PyObjectProtocol`, for comparisons and string conversions into a derive-macro . It also implements the functions we use to expose the "Chia Streamable" serialization to python.

This reduces code duplication, but also lays the foundation for a future patch to implement `to_json_dict()`, which is another aspect we rely on in order to print streamable objects.


richardkiss (approved):
Seems like a great idea!

PR #18, opened 2022-05-14
add ToJsonDict trait and generic implementation for Streamable types

arvidn:
converting these two type, `PySpend` and `PySpendBundleConditions`, into json isn't critical, but we rely on it for `__str__()` on streamable types, which is used in tests (and one error log).

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -35,22 +35,32 @@ jobs:
       with:
         toolchain: stable
 
+    - name: setup venv
+      run: |
+        python -m venv venv
+        ln -s venv/bin/activate
+
     - name: Install dependencies
       run: |
+        . ./activate
         python -m pip install maturin
         rustup target add x86_64-unknown-linux-musl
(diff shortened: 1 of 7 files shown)


arvidn:
the tests ran here: https://github.com/Chia-Network/chia_rs/runs/6433750450?check_suite_focus=true#step:9:13

richardkiss:
This code is excellent, and it's great you've figured out macros. I understand the motivation behind this, but I've always hated putting the JSON serialization code in with `Streamable` and thought it a huge mistake. The reason is because I've always dreamed about being able to create a minimal library that validates consensus, and extra code needed to create APIs and manage mempool (which is not part of consensus! Just part of chain growth) built on top. If there's any way to separate this out into a module that wouldn't have to be included for a consensus-only library, I would be very grateful.

richardkiss (approved):

arvidn:
I agree. I kind of consider this a stop-gap until we can separate out the json serialization in chia-blockchain

PR #20, opened 2022-05-20
Add type hints

arvidn:
also rename `PySpend` and `PySpendBundleConditions` to just `Spend` and `SpendBundleCondition`


richardkiss (approved):
Great change!

PR #21, opened 2022-05-23
Add from_json_dict()

arvidn:
This one `Streamable` macro is getting crowded, however, you can only have a single `impl` block marked with `#[pymethods]` so they all have to be implemented at the same time.

I tried the build configuration to allow more than one, but it failed with a mysterious error message.


richardkiss (approved):
I didn't get too much into the details but this doesn't seem too critical and the rust compiler seems like it will be a lot more pedantic than I could ever be.

PR #22, opened 2022-05-23
fix typo in type stubs

arvidn:
Fix return type of to_bytes() to be a bytes object instead of a list of integers


wheel/chia_rs.pyi
```diff
@@ -20,9 +20,9 @@ class Spend:
     def __repr__(self) -> str: ...
     def __richcmp__(self) -> Any: ...
     @staticmethod
-    def from_bytes(self, bytes) -> Spend: ...
+    def from_bytes(bytes) -> Spend: ...
     @staticmethod
-    def parse_rust(self, bytes) -> Tuple[Spend, int]: ...
+    def parse_rust(bytes) -> Tuple[Spend, int]: ...
```
richardkiss, line 25: You might consider making this a "classmethod" rather than static methods since they generally use the class declaration.

```
def from_bytes(cls, bytes) -> Spend: ...
```

richardkiss (approved):

PR #23, opened 2022-05-24
support copy

arvidn:


richardkiss:
Why bother with this? Isn't `Spend` immutable?

richardkiss (approved):

arvidn:
even though `Spend` is immutable, we have at least one python test in `chia-blockchain` that calls `copy.deepcopy()` on types that we want to be rust native. Without direct support for `__copy__` and `__deepcopy__`, python will try to pickle the object. Supporting pickling is much more complicated.

PR #25, opened 2022-05-24
replace custom slice() with AsRef and Deref traits

arvidn:
to clean up usage of BytesImpl

### DIFF
--- a/src/streamable/bytes.rs
+++ b/src/streamable/bytes.rs
@@ -1,8 +1,10 @@
 use core::fmt::Formatter;
 use serde::ser::SerializeTuple;
 use serde::{Deserialize, Serialize};
+use std::convert::AsRef;
 use std::fmt;
 use std::fmt::Debug;
+use std::ops::Deref;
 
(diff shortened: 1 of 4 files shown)


src/streamable/bytes.rs
```diff
@@ -115,14 +117,22 @@ impl<'a, const N: usize> From<&'a BytesImpl<N>> for &'a [u8] {
...
 
+impl<const N: usize> AsRef<[u8]> for BytesImpl<N> {
+    fn as_ref(&self) -> &[u8] {
+        &self.0
+    }
+}
+impl<const N: usize> Deref for BytesImpl<N> {
+    type Target = [u8];
+    fn deref<'a>(&'a self) -> &'a [u8] {
+        &self.0
```
richardkiss, line 133: Can you fix this clippy warning?

arvidn: I thought clippy warnings would fail CI. I think github might be using a newer version than we use in CI.

richardkiss (approved):

PR #26, opened 2022-05-24
Coin type

arvidn:
This patch implements `Coin` in rust. Apart from supporting "Streamable" serialization/deserialization, from_json, to_json etc. `Coin` has one function: "name()", which computes the coin name.

Using a rust implementation for `Coin` provides two main improvements:

1. The `name()` function can be implemented efficiently (without allocating any any heap space)
2. The data structure of the coin is compact and more efficient to access because of that (all fields are fixed size)

My benchmarks suggests that this shaves off about 22% CPU when validating blocks. Since sync time is dominated by disk I/O, the wall-clock time for the benchmark has only a modest improvement.

benchmark: M1, SSD, sync full block

| test | before | after | after / before |
| --- | --- | --- | --- |
| `receive_block_batch()` CPU | 17.46 % | 13.57 % | 0.777 |
| wall-clock time |  109.84s | 104.46s | 0.95 |

It's also noteworthy that `Coin.name()` CPU usage goes from 3.45 % -> 0.68% a speedup of 5x.

before:
slow-batch-00448

after:
slow-batch-00448


arvidn:
I suspect this might help the wallet as well, where `Coin.name()` appears to be called a lot.

wheel/chia_rs.pyi
```diff
@@ -87,3 +88,26 @@ class LazyNode:
     def atom() -> bytes: ...
 
 def serialized_length(program: bytes) -> int: ...
+
+class Coin:
+    parent_coin_info: bytes32
+    puzzle_hash: bytes32
```
arvidn, line 94: The choice to use `bytes32` here deserves some scrutiny.

1. This adds a dependency onto `chia-blockchain`
2. These types are actually just regular `bytes`, but guaranteed to be 32 bytes.
3. Specifying `bytes` as the type here causes a *lot* of mypy issues in `chia-blockchain`, when (afaik) `bytes32` doesn't have any important behaviour that differs from a `bytes` type of size 32. (as opposed to `uint64`)

richardkiss: Have you considered moving `bytes32` to `chia_rs`?

richardkiss:
Wow, that's an impressive speed-up.

I agree that maybe we need to consider moving away from these fancy types. I suspect it could end up bubbling to many places

tests/test_coin.py
```diff
@@ -0,0 +1,330 @@
...
+    )
+    c = Coin(parent_coin, puzzle_hash, 0x7FFFFFFFFFFFFFFF)
+    assert (
+        c.name()
+        == sha256(
+            parent_coin
+            + puzzle_hash
+            + bytes([0x7F, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF])
+        ).digest()
+    )
```
richardkiss, line 128: Maybe some `0x80` prefixes too? Since that's where the extra-byte wraparound happens.

richardkiss (approved):

wheel/chia_rs.pyi
arvidn: it would still be a python type though, right? I'm not sure it's possible to make `bytes32` a rust type and still appear to be derived from the built-in `bytes` type.

So, no, I didn't consider that. It sounds a bit scary.

tests/test_streamable.py
```diff
@@ -41,97 +44,118 @@ def test_json_spend() -> None:
...
 
-    b = Spend.from_json_dict({
-        "coin_id": "0x" + coin.hex(),
-        "puzzle_hash": "0x" + ph.hex(),
-        "height_relative": 1337,
-        "seconds_relative": 0,
-        "create_coin": [["0x" + ph2.hex(), 1000000, None]],
-        "agg_sig_me": [["0x" + sig.hex(), "0x6d7367"]],
-    })
+    b = Spend.from_json_dict(
```
richardkiss, line 69: Are these just `black` fixes? You could probably get away with reverting them.

wheel/chia_rs.pyi
richardkiss: Yeah, you'd probably have to either create a `bytes32` type in rust (that subclasses `bytes`), or do it in python and call out from rust to python to create it. I'd say let's get this in as-is for now and fight with `pyo3` another day.

richardkiss (approved):

tests/test_streamable.py
arvidn: yeah, I think so. reverted

PR #32, opened 2022-07-25
improve the parse_rust() function to accept a buffer

arvidn:
This avoids a copy of the entire serialized buffer when calling `parse_rust()`. It especially speeds up wallet sync:

main:
```
INFO     Successfully subscribed and updated 103906 coin ids
INFO     Sync (trusted: True) duration was: 148.95257115364075
```

with this patch:
```
INFO     Successfully subscribed and updated 103906 coin ids
INFO     Sync (trusted: True) duration was: 123.46990084648132
```

The CPU profile of the wallet sync looks like this:

main:

chia-hotspot-2

this patch:

chia-hotspot-2

### DIFF
--- a/.github/workflows/build-arm64-wheels.yml
+++ b/.github/workflows/build-arm64-wheels.yml
@@ -51,7 +51,7 @@ jobs:
             if [ ! -f "activate" ]; then ln -s venv/bin/activate; fi && \
             . ./activate && \
             pip install maturin && \
-            CC=gcc maturin build --no-sdist --release --manylinux 2014 -m wheel/Cargo.toml \
+            CC=gcc maturin build --release --manylinux 2014 -m wheel/Cargo.toml \
           '
 
     - name: Upload artifacts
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -16,12 +16,12 @@ py-bindings = ["dep:pyo3"]
 
 [dependencies]
 serde = { version = "1.0.130", features = ["derive", "rc"] }
-clvmr = "=0.1.20"
+clvmr = "=0.1.21"
```
richardkiss, line 19: You're going to want 0.1.22 here (as you probably know).

.github/workflows/build-arm64-wheels.yml
```diff
@@ -17,6 +17,7 @@ jobs:
     strategy:
       matrix:
         os: [[ARM64, Linux]]
+        python: ["cp37-cp37m", "cp38-cp38", "cp39-cp39"]
```
richardkiss, line 20:
```suggestion
        python: ["cp37-cp37m", "cp38-cp38", "cp39-cp39", "cp310-cp310"]
```

richardkiss:
With `maturin` removing `--no-sdist` and adding `--sdist` and changing the default, you may need to add an `--sdist` to the place that builds & uploads the `sdist`.

richardkiss (approved):

PR #33, opened 2022-07-26
add script to generate chia_rs.pyi type stubs

arvidn:
this is a step towards lowering the hurdle to move more `Streamable` python types into this repo


richardkiss (approved):
Excellent

PR #35, opened 2022-07-28
use Sha256 wrapper from clvm_rs

arvidn:


richardkiss:
My main complaint is that we moving from the generally-accepted standard [Hasher](https://doc.rust-lang.org/std/hash/trait.Hasher.html) trait to a home-grown one. We should probably modify the `clvm_rs` version to use a more standard API. But I'll approve this for now and let you decide how best to proceed.

I'm happy to make the `clvm_rs` change if you like... just let me know.

richardkiss (approved):

arvidn:
that's a good point. our `Sha256` "wrapper" should implement the `Hasher` trait. It would need some more wrapping though, and this patch doesn't really make it worse, we don't implement that trait today.

If you make that change in `clvm_rs` I'll make the corresponding change in `chia_rs`

arvidn:
rebased on top of the test fix

PR #37, opened 2022-07-28
move error.rs and bytes.rs to chia_rs root

arvidn:
I don't think these really belong in `streamable` module. This shortens `use` statements.
This is also preparing for a great simplification (and code deletion) of the streamable implementation.


richardkiss:
I'm fine with this re-org. I'd put all this stuff under the `streamable` module so it was easier to disable, since it was experimental at the time.

richardkiss (approved):

richardkiss:
The benchmark is failing, but I'm not familiar enough with it and the expect log file output to know if this is a legitimate failure or not.

arvidn:
what's really needed is for the benchmarks to run on our dedicated benchmark runner. it requires some coordination with IPS to whitelist this repo I believe.

PR #40, opened 2022-07-29
replace serde with a simple trait + macro for Streamable

arvidn:
this patch drops the dependency on serde and implements *just* the Streamable support we need (and that the Streamable protocol supports).

This has the following advantages:

* We need a macro to generate the python binding tying the serde serializer and deserializer. This code is about as large as the whole Streamable-support after this patch. So, this patch removes a *lot* of code (mostly boiler plate).
* Implementing a serializer and deserializer for serde requires compile-time support for a lot of types that we don't support in the Streamable format. e.g. floating point and dictionaries. Running serialization through serde turns errors that *should* be compile-time errors into runtime errors (i.e. the `NotSupported` run-time error is produced in a bunch of  places). With this patch, marking a class as `Streamable` containing unsupported types is a compile error with this patch
* Going forward, we'll need one more "serializer" that computes the sha256 hash of the `Streamable` representation. With this simpler streamable implementation, that's simple to add and doesn't take a lot of code, whereas with serde we would need to duplicate a lot of the boilerplate code from the `ChiaSerializer` for all unsupported types.
* Using a simpler input buffer type and output buffer type (slice + position and `Vec<u8>`) makes a number of operations more efficient. Especially byte-buffers. With serde sequences need to be serialized one item at a time. With this patch we can just memcopy the buffer directly into the `Vec<u8>` and out of the slice.
* Using a simpler input buffer also allows us to reuse the existing function that computes the length of a CLVM serialized buffer. This is especially handy for creating a rust version of the `FullBlock` type (which I believe would be an important step towards making the full node more efficient).

One, perhaps, controversial part of this patch is that I made a custom `InputBuffer` type, that's just a slice + cursor rather than using the `Reader` trait. Likewise for the output buffer, I use a `Vec<u8>` instead of the `Writer` trait. The reason to not use the `Reader` is explained in the bullet list above, to get direct access to the bytes. The main difficulty with the `Writer` trait is that its errors are `std::io::Error`, which is a non-trivial type. It can't be compared equal for instance. Using a simpler output buffer (that cannot fail, since memory allocation isn't an error) both simplifies error handling and also the error type. See the change to the `Error` enum in this patch.


src/gen/conditions.rs
```diff
@@ -314,7 +313,7 @@ fn parse_args(
     }
 }
 
-#[derive(Serialize, Deserialize, Debug)]
```
arvidn, line 317: These types were marked `Serialize` and `Deserialize` by mistake (as far as I can tell). It was exposed when changing this to `Streamable` because `NodePtr` does not implement the `Streamable` trait. If it worked previously, it would not have produced any useful results (since `NodePtr` is an `int` under the hood).

src/streamable.rs
```diff
@@ -0,0 +1,625 @@
...
+        Ok((
+            T::parse(input)?,
+            U::parse(input)?,
+            V::parse(input)?,
+            W::parse(input)?,
+        ))
+    }
+}
+
+// ===== TESTS ====
```
arvidn, line 186: all of these tests were copied over from the `de.rs` and `ser.de`. But the `NotSupported` tests were removed (because those are compile-time errors now)

src/streamable.rs
```diff
@@ -0,0 +1,625 @@
+use crate::chia_error::{Error, Result};
+use std::convert::TryInto;
+use std::mem::size_of;
+
+pub struct InputStream<'a> {
```
richardkiss, line 5: Any reason to not use https://doc.rust-lang.org/std/io/struct.Cursor.html here?

src/streamable.rs
```diff
@@ -0,0 +1,625 @@
...
+        // attacks
+        let mut ret = Vec::<T>::new();
+        for _ in 0..len {
+            ret.push(T::parse(input)?);
+        }
+        Ok(ret)
+    }
+}
+
+impl Streamable for String {
```
richardkiss, line 86: Maybe a little more challenging, but you might be able to use `Cursor` here too. You'd have to convert `chia_error::Error` to `std::io::Error` (and downcasting might lose some diagnostics, although I'd guess at this point all you get is a programmer-readable error). Might be no tangible benefit.

richardkiss (approved):
Looks good! Maybe consider `Cursor` for reading and possibly writing, as I wrote above.

arvidn:
yeah, I think using `Cursor` makes sense.

arvidn:
I rebased this on top of latest `main` to get the test fix in. I also squashed-in the switch to use `std::io::Cursor` instead of a home-grown class.

wheel/py_streamable/src/lib.rs
```diff
@@ -63,8 +63,8 @@ pub fn py_streamable_macro(input: TokenStream) -> TokenStream {
 
                     #[staticmethod]
                     pub fn from_bytes(blob: &[u8]) -> PyResult<Self> {
-                        let mut de = ChiaDeserializer::from_slice(blob)?;
-                        Self::deserialize(&mut de).map_err(|e| e.into())
+                        let mut input = std::io::Cursor::<&[u8]>::new(blob);
+                        Self::parse(&mut input).map_err(|e| <chia::chia_error::Error as Into<PyErr>>::into(e))
```
richardkiss, line 67: Note that the `?` operator does a `Err(From::from(e))` implicitly, so you might be able to simplify this if you implement `From<chia_error::Error>` for `PyErr`. Maybe this isn't possible, but might be worth looking into.

richardkiss (approved):

wheel/py_streamable/src/lib.rs
arvidn: yeah, I'm aware of that (but probably fail to take advantage of it at times). I'm not 100% sure, but I believe the reason I made this change was because a file using this macro probably hadn't imported `chia::chia_error::Error`, so it wasn't in scope. I think it's a bit problematic when the macro requires the user to import a bunch of types so I tried to make the macro generate as many fully qualified types as possible. So I believe fully qualifying the type here makes the macro simpler to use.

I think it's worth looking into simplifying it still, but in a separate PR.

PR #42, opened 2022-08-01
add support for get_hash() to the Streamable trait and the derive macro

arvidn:
This mirrors the python `Streamable` protocol, which has a `get_hash()` function. Although, the one I added here nests by taking the hashing context by mutable reference. The python binding wraps it to give it the same behavior as in python though.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1,5 +1,6 @@
 use crate::chia_error;
 use crate::streamable::{read_bytes, Streamable};
+use clvmr::sha2::Sha256;
 use core::fmt::Formatter;
 use std::convert::AsRef;
 use std::convert::TryInto;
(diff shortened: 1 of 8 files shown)


richardkiss:
Naming nit – I don't like to use the `get_` prefix for things that aren't "fast". Maybe consider `calculate_hash` or `generate_hash` or some other verb which suggests that this isn't a cheap operation.

src/bytes.rs
```diff
@@ -17,6 +18,10 @@ use pyo3::types::PyBytes;
 pub struct Bytes(Vec<u8>);
 
 impl Streamable for Bytes {
+    fn get_hash(&self, digest: &mut Sha256) {
```
richardkiss, line 21: Now that I see it... maybe `update_digest`?

richardkiss (changes requested):
I think we need a new name for `get_hash`, especially since it doesn't return anything.

richardkiss:
Patch looks good otherwise.

src/bytes.rs
arvidn: yeah, I'll change it to `update_digest()`

richardkiss (approved):
Looks good.

PR #43, opened 2022-08-01
remove RangeSet

arvidn:
now that redundant leading zeros are invalid in conditions, there's no need to jump through hoops to support malicious programs.

The back story of the `RangeSet` type is that, according to the original consensus rules, condition arguments were allowed to be arbitrarily long, padded with leading zeros. This was fine as long as the integer we parsed out from the atom still were within the limits.

This opened up a DoS attack where malicious generator programs could produce very large (but valid) outputs that would take a very long time to parse. It would be especially effective if the same leading zeros were reused in multiple conditions. That would make the full nodes parse the same zeros over and over again.

To mitigate this DoS attack, this I wrote this `RangeSet` class which remembered where large ranges of zero-bytes have already been found. When parsing an integer, any reused leading zeros could just be skipped.

In the soft-fork at height 2300000, we restricted the condition arguments to not be allowed to ever have any redundant leading zeros. That means all this extra complexity to support this attack without falling over is no longer necessary. We can just fail the condition immediately.

This patch removes the `RangeSet` and the associated flag to require canonical integer representation (i.e. no redundant leading zeros).


richardkiss (approved):
Nice

PR #51, opened 2022-08-12
Fix type stubs

arvidn:
I also added a CI job to ensure the generated file is up-to-date.


wheel/chia_rs.pyi
```diff
@@ -31,7 +31,7 @@ class LazyNode:
     def atom() -> bytes: ...
 
 def serialized_length(program: bytes) -> int: ...
-def tree_hash(program: bytes) -> bytes32: ...
+def tree_hash(program: ReadableBuffer) -> bytes32: ...
```
richardkiss, line 34: I wasn't able to find `ReadableBuffer` declared or imported. Does this work with mypy?

richardkiss (approved):
Seems good, except for question about `ReadableBuffer`.

wheel/chia_rs.pyi
arvidn: it's on line 9 in this file. It's a poor-man's definition of a python buffer.

arvidn: the reason this changed here was that I forgot to regenerate this file when merging this change

richardkiss: Ah, got it, sorry, not sure how I missed it.

PR #55, opened 2022-08-31
First crack at compressor.

richardkiss:


src/compression/bytes32.rs
```diff
@@ -0,0 +1,34 @@
+use std::fmt::{Debug, Formatter};
```
richardkiss, line 1: @arvid I know you already have a `Bytes32` type in this crate, so this is far from ideal. I'll try to migrate over to yours, but it's not a drop-in replacement as this one is basically a tuple with one entry `[u8; 32]`, and the existing one is a typedef.

src/compression/compressor.rs
```diff
@@ -0,0 +1,88 @@
+use super::serialize::node_to_bytes_backrefs;
```
richardkiss, line 1: There are more functions here than we really need.

src/compression/read_cache_lookup.rs
```diff
@@ -0,0 +1,250 @@
+/// When deserializing a clvm object, a stack of deserialized child objects
```
richardkiss, line 1: I might be able to import some utility functions from `clvm_rs` rather than pasting them here. A separation in `clvm_rs::serialize` into two files (one with serialization and the other with deserialization functions) might be helpful here.

src/compression/serialize.rs
```diff
@@ -0,0 +1,247 @@
+use std::io;
```
richardkiss, line 1: I might be able to import some utility functions from `clvm_rs` rather than pasting them here. A separation in `clvm_rs::serialize` into two files (one with serialization and the other with deserialization functions) might be helpful here.

src/compression/bytes32.rs
richardkiss: Done.

src/compression/compressor.rs
```diff
@@ -0,0 +1,56 @@
...
+    let compressed_block = allocator.new_pair(quote_node, node_ptr)?;
+    let program = allocator.new_pair(quote_node, program)?;
+    let list = allocator.null();
+    let list = allocator.new_pair(compressed_block, list)?;
+    let list = allocator.new_pair(program, list)?;
+    let list = allocator.new_pair(apply_node, list)?;
+    Ok(list)
+}
+
+pub fn decompress(allocator: &mut Allocator, blob: &[u8]) -> Result<NodePtr, std::io::Error> {
```
arvidn, line 28: if we ever end up needing this in the node or wallet, I would expect we'd want a rust implementation.

src/compression/compressor.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+pub fn create_autoextracting_clvm_program(input_program: &[u8]) -> std::io::Result<Vec<u8>> {
+    let mut allocator = Allocator::new();
+    let node_ptr = decompress(&mut allocator, input_program)?;
+    let node = Node {
+        allocator: &allocator,
+        node: node_ptr,
+    };
+    let compressed_block = node_to_bytes_backrefs(&node).expect("can't compress");
```
arvidn, line 51: I take it this only fails if there's a bug, right?

arvidn (commented):
I would expect the whole `serialize.rs` file can be replaced by pulling in those functions from `clvm_rs`, or are there any non-obvious hurdles to do that?

I would like to see unit tests for virtually all functions (including the trivial ones like converting a `NodePtr` to an index.

The compression function needs round-trip tests of a few examples of trees (ideally to cover all the cases in the compression).

There should be a fuzzer, both that feeds garbage into the compressor and decompressor. But also a fuzzer that exercises round-trip on a random tree (I recently added some fuzzing utility functions to create a tree from the fuzzer input buffer).

I'm also curious about what kind of support or convention we have to test clvm code. It would be really nice to be able to unit test all your functions too.

src/compression/compressor.rs
```diff
@@ -0,0 +1,56 @@
...
+        allocator.null(),
+        max_cost,
+        None,
+    )?;
+    Ok(reduction.1)
+}
+
+pub fn create_autoextracting_clvm_program(input_program: &[u8]) -> std::io::Result<Vec<u8>> {
+    let mut allocator = Allocator::new();
+    let node_ptr = decompress(&mut allocator, input_program)?;
```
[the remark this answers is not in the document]
richardkiss: This was originally for a command-line tool. It's not used at the moment.

src/compression/compressor.rs
richardkiss: Correct. Every tree should be compressible. I'm less sure there are no DOS attacks. But that's just the farmer's problem for now.

src/compression/compressor.rs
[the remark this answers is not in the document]
arvidn: but the idea is that `create_autoextracting_clvm_program()` will be used by the farmer, right? So this function needs to exist (and make sense).

As far as I understand, if anyone would pass a compressed program into this function, it would be indicative of a bug. If I understand this correctly, I would expect that to be an error rather than a silent recompress operation.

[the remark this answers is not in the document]
richardkiss: Sorry, my comment is wrong. You are right, this is the main API.

That said, I don't see any harm in allowing it to accept trees serialized with backrefs, as it's just an extension of regular deserialization.

src/compression/compressor.rs
```diff
@@ -0,0 +1,55 @@
...
+        &dialect,
+        program,
+        allocator.null(),
+        max_cost,
+        None,
+    )?;
+    Ok(reduction.1)
+}
+
+pub fn create_autoextracting_clvm_program(input_program: &[u8]) -> std::io::Result<Vec<u8>> {
```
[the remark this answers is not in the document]
richardkiss: I'll bring over the fuzzer from `clvm_rs`.

arvidn (dismissed):

arvidn:
I was about to press "approve". But this looks disconcerting: https://github.com/Chia-Network/chia_rs/actions/runs/3833857129/jobs/6525695310#step:7:47

Why was that fuzzing cancelled? and why did it not cause the CI to fail?

arvidn (approved):

arvidn:
rerunning the job made it pass

PR #56, opened 2022-08-31
implement get_puzzle_and_solution_for_coin() in chia_rs

arvidn:
The motivation for the function is that running a generator and pulling out the coin puzzle and solution is currently quite expensive. This speeds it up about 7 folds.

profile before:
chia-hotspot-25-33-main

profile after:
chia-hotspot-14-24


mariano54:
Can we skip the get_full_block call, and access the generator bytes directly?

mariano54:
Or does that not work due to previous generators?

arvidn:
> Can we skip the get_full_block call, and access the generator bytes directly?

We can't skip the DB lookup, but we can skip the expensive part, parsing the `FullBlock`. I'm working on that patch now

richardkiss (approved):
Seems pretty straightforward.

The linear search gives a speed advantage to coins spent earlier in a block. Without indexing, we're stuck though. Maybe we add a column to `CoinRecord` to indicate index in block.

Can we create a test that runs this on [this code](https://gist.github.com/richardkiss/e4a0e6ceb97489f7a9f1d3938b1a0d43) ? It's a "copy-protection" (or "fork detector") generator that ensures the correct (and broken) deserialization code is there.

arvidn:
> Can we create a test that runs this on [this code](https://gist.github.com/richardkiss/e4a0e6ceb97489f7a9f1d3938b1a0d43) ? It's a "copy-protection" (or "fork detector") generator that ensures the correct (and broken) deserialization code is there.

yes, but we would have to do that at the `chia-blockchain` level. the deserializer argument is still expected to be passed in as an argument.

PR #57, opened 2022-09-01
add fuzzers

arvidn:
this patch adds fuzz targets to parsing and checking conditions from a block generator, as well as the recently added `get_puzzle_and_solution_for_coin()`

### DIFF
--- a/src/fuzzing_utils.rs
+++ b/src/fuzzing_utils.rs
@@ -1,0 +1,70 @@
+use clvmr::allocator::Allocator;
+use clvmr::allocator::NodePtr;
+
+pub struct BitCursor<'a> {
+    data: &'a [u8],
+    bit_offset: u8,
+}
+
+fn mask(num: u8) -> u8 {
+    0xff >> num
+}
+
+impl<'a> BitCursor<'a> {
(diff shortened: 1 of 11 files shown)


src/fuzzing_utils.rs
```diff
@@ -0,0 +1,68 @@
...
+        if bits >= 8 {
+            self.data = &self.data[(bits / 8) as usize..];
+        }
+        self.bit_offset = (bits % 8) as u8;
+    }
+}
+
+const BUFFER: [u8;63] = [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0];
+
+pub fn make_tree(a: &mut Allocator, cursor: &mut BitCursor) -> NodePtr {
```
arvidn, line 51: make_tree() turns the fuzzer input and turns it into a SExp tree

richardkiss (approved):
Seems good.

PR #59, opened 2022-09-02
turn tree_hash() into non-recursive implementation

arvidn:
harden the `tree_hash()` function by moving away from a recursive implementation.
extend unit test
add fuzzer

### DIFF
--- a/src/gen/get_puzzle_and_solution.rs
+++ b/src/gen/get_puzzle_and_solution.rs
@@ -5,24 +5,43 @@ use clvmr::op_utils::u64_from_bytes;
 use clvmr::sha2::{Digest, Sha256};
 use std::convert::AsRef;
 
-// TOOD: make this not recursive
+enum TreeOp {
+    SExp(NodePtr),
+    Cons,
+}
+
 // TODO: perhaps this should move to clvm_rs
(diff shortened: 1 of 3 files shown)


src/gen/get_puzzle_and_solution.rs
```diff
@@ -5,24 +5,43 @@ use clvmr::op_utils::u64_from_bytes;
...
+                SExp::Pair(left, right) => {
+                    ops.push(TreeOp::Cons);
+                    ops.push(TreeOp::SExp(left));
+                    ops.push(TreeOp::SExp(right));
+                }
+            },
+            TreeOp::Cons => {
+                let mut sha256 = Sha256::new();
+                sha256.update([2_u8]);
+                sha256.update(hashes.pop().unwrap());
```
richardkiss, line 36: As a policy, better to use `.expect` rather than `.unwrap`. Shouldn't matter here though since unwrap can't really fail.

richardkiss (approved):
Looks good

src/gen/get_puzzle_and_solution.rs
arvidn: good point

PR #61, opened 2022-09-08
make python types sendable

arvidn:
This allows them to be used across threads in python


richardkiss (approved):
Seems pretty straightforward.

PR #64, opened 2022-10-07
run benchmarks on dedicated CI runner

arvidn:


.github/workflows/benchmark.yml
```diff
@@ -70,18 +70,57 @@ jobs:
...
+    - name: setup venv
+      run: |
+        python -m venv venv
+
+    - uses: chia-network/actions/activate-venv@main
+
+    - name: Install dependencies
+      run: |
+        python -m pip install maturin
+        python -m pip install colorama
```
altendky, line 109: pip does better with fewer calls so that it can consider the whole dependency tree at one time when selecting versions.

```suggestion
        python -m pip install maturin colorama
```

.github/workflows/benchmark.yml
```diff
@@ -70,18 +70,57 @@ jobs:
...
+        fetch-depth: 1
+
+    - uses: chia-network/actions/setup-python@main
+      name: Install Python ${{ matrix.python-version }}
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Update pip
+      run: |
+          python -m pip install --upgrade pip
```
altendky, line 93: I would probably do this inside the venv below  with the dependency installation.  I think this will be inherited into the env, but may as well just update it where we actually care.  Kind of counter to the comment there, _this_ would be a separate call two pip.  So you end up with two.  One to update install-related stuff like pip and maybe setuptools or wheel or build, and another to install every other package.

richardkiss (approved):

richardkiss:
I'm not a CI expert, but this seems unobjectional.

PR #65, opened 2022-10-07
support Tuples (unnamed members)

arvidn:
as well as 128 bit integral types in the rust streamable macro.

Plain tuples were already supported (up to a few fields). This adds support for arbitrary types with unnamed members. i.e. when you define a type like this:

```
struct G1Element([u8;48]);
```

This is a step towards being able to have more chia types in rust. The wallet-protocol types are next on my list.

### DIFF
--- a/src/streamable.rs
+++ b/src/streamable.rs
@@ -62,6 +62,8 @@ streamable_primitive!(u32);
 streamable_primitive!(i32);
 streamable_primitive!(u64);
 streamable_primitive!(i64);
+streamable_primitive!(u128);
+streamable_primitive!(i128);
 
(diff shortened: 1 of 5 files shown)


src/streamable.rs
```diff
@@ -472,6 +505,16 @@ fn test_parse_struct() {
     );
 }
 
+#[cfg(test)]
+#[derive(Streamable, PartialEq, Debug)]
+struct TestTuple(String, u32);
```
arvidn, line 510: here's the test demonstrating it working

richardkiss (approved):
Looks good

PR #67, opened 2022-10-07
Clippy fmt workspace

arvidn:
run `cargo fmt` and `cargo clippy` on every crate in the workspace, not just the root


richardkiss (approved):

ChiaMineJP (approved):
lgtm

PR #68, opened 2022-10-10
add new crate chia-bls

arvidn:
with support for all Chia key operations, from BIP39, hardened and unhardened key derivation, signatures, aggregation and verification.

The overview of the functions added in this crate, see the [README](https://github.com/Chia-Network/chia_rs/blob/28fac7629e71fa9c68600bc2c36c412f0479514d/chia-bls/README.md).

The unit tests include some regression tests against `blspy`, with manual python code snippets.

There are two fuzz targets, one that loads `blspy` and validates results against it ([here](https://github.com/Chia-Network/chia_rs/pull/68/files#diff-28956f0dc0211fe179419220378cbaf20127022210f6bb8a53c8868c45e02582)) and one that checks internal consistency of these functions ([here](https://github.com/Chia-Network/chia_rs/pull/68/files#diff-8e0b9cdf9ef27d60d0026f266b36fd03dd11ac9208211c4f4a28b8571b4f4d72)). e.g. unhardened derivation of secret key and public key independently still validate the signature.

There are 3 benchmark targets. This implementation takes about 2.3 times longer to verify a signature compared to blspy/bls-signatures.

```
     Running benches/verify.rs
verify, small msg       time:   [2.3635 ms 2.3666 ms 2.3706 ms]
                        change: [-0.2317% -0.0294% +0.1766%] (p = 0.80 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  2 (2.00%) high mild
  8 (8.00%) high severe

verify, 4kiB msg        time:   [2.3765 ms 2.3797 ms 2.3838 ms]
                        change: [-0.5150% -0.2241% +0.0461%] (p = 0.12 > 0.05)
                        No change in performance detected.
Found 14 outliers among 100 measurements (14.00%)
  3 (3.00%) high mild
  11 (11.00%) high severe
```

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -70,7 +70,7 @@ jobs:
       run: |
         pytest tests
 
-  benchmarks:
+  generator-benchmarks:
     name: Generator performance
     runs-on: benchmark
     strategy:
@@ -121,3 +121,23 @@ jobs:
         cd tests
         ./generate-programs.py
         ./run-programs.py
+
+  benchmarks:
+    name: rust benchmarks
+    runs-on: benchmark
+    strategy:
+      fail-fast: false
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
+
+    - name: Set up rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+
+    - name: cargo bench
+      run: |
+        cargo bench --all
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -310,7 +310,7 @@ jobs:
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-features --all
+          args: --all-features --workspace
 
   fuzz_targets:
     runs-on: ubuntu-latest
@@ -338,7 +338,7 @@ jobs:
         with:
             toolchain: stable
       - name: cargo test
-        run: cargo test
+        run: cargo test --workspace --all-features
 
   upload:
     name: Upload to PyPI - ${{ matrix.os.name }} ${{ matrix.python.major-dot-minor }} ${{ matrix.arch.name }}
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -220,7 +220,7 @@ jobs:
       uses: actions/upload-artifact@v3
       with:
         name: packages
-        path: ./target/wheels/
+        path: ./wheel/target/wheels/
 
   check-typestubs:
     name: Check chia_rs.pyi
@@ -280,7 +280,7 @@ jobs:
       uses: actions/upload-artifact@v3
       with:
         name: packages
-        path: ./target/wheels/
+        path: ./wheel/target/wheels/
 
   fmt:
     runs-on: ubuntu-latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,8 @@
+# the "wheel" crate is excluded from the workspace because pyo3 has problems with
+# "cargo test" and "cargo bench"
 [workspace]
-members = ["wasm", "wheel", "chia_streamable_macro"]
+members = ["wasm", "chia_streamable_macro", "chia-bls"]
+exclude = ["wheel"]
 
 [package]
 name = "chia"
--- a/chia-bls/Cargo.toml
+++ b/chia-bls/Cargo.toml
@@ -1,0 +1,31 @@
+[package]
+name = "chia-bls"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+tiny-bip39 = "=1.0.0"
+anyhow = "=1.0.65"
+# the newer sha2 crate doesn't implement the digest traits required by hkdf
+sha2 = "=0.9.9"
+bls12_381_plus = "=0.7.0"
+num-bigint = "=0.4.3"
+hkdf = "=0.11.0"
+group = "=0.12.0"
+
+[dev-dependencies]
+hex = "^0.4.3"
+rand = "^0.8.5"
+criterion = "^0.4"
+
+[[bench]]
+name = "derive_key"
+harness = false
+
+[[bench]]
+name = "sign"
+harness = false
+
+[[bench]]
+name = "verify"
+harness = false
--- a/chia-bls/README.md
+++ b/chia-bls/README.md
@@ -1,0 +1,76 @@
+Library providing building blocks for a Chia wallet.
+
+BIP39 mnemonic handling:
+
+```
+fn entropy_to_mnemonic(entropy: &[u8; 32]) -> String
+fn mnemonic_to_entropy(mnemonic: &str) -> Result<[u8; 32], Error>
+fn entropy_to_seed(entropy: &[u8; 32]) -> [u8; 64]
+```
+
+SecretKey
+
+```
+impl SecretKey {
+    pub fn from_seed(seed: &[u8; 64]) -> SecretKey
+    pub fn from_bytes(bytes: &[u8; 32]) -> Option<SecretKey>
+    pub fn to_bytes(&self) -> [u8; 32]
+
+    pub fn public_key(&self) -> PublicKey
+
+    pub fn derive_unhardened(&self, idx: u32) -> SecretKey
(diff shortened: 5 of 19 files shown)


.github/workflows/benchmark.yml
```diff
@@ -121,3 +121,23 @@ jobs:
...
+        fetch-depth: 1
+
+    - name: Set up rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+
+    - name: chia-keys benchmark
+      run: |
+        cargo bench --all
```
arvidn, line 143: there's no upper limit imposed by CI right now, perhaps there should be, to catch performance regressions

.github/workflows/build-test.yml
```diff
@@ -310,7 +310,7 @@ jobs:
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-features --all
+          args: --all-features --workspace
```
arvidn, line 313: apparently `--all` is deprecated in favor of `--workspace`

.github/workflows/build-test.yml
```diff
@@ -338,7 +338,7 @@ jobs:
         with:
             toolchain: stable
       - name: cargo test
-        run: cargo test
+        run: cargo test --workspace --all-features
```
arvidn, line 341: there's still a problem with the pyo3 wheel failing when running tests (and benchmarks I believe). I haven't figured out a way around this yet. Assistance would be appreciated.

richardkiss: I ran `git bisect` and the change that breaks this is 625e51fc63c0ea1082a4b1f0d5a1a85d343841e0.

richardkiss: See also https://github.com/PyO3/pyo3/issues/340 and https://github.com/PyO3/pyo3/issues/341

richardkiss: Incidentally, https://pyo3.rs/main/features.html#multiple-pymethods suggests `multiple-pymethods` may break wasm builds. :(

chia-keys/src/derive_keys.rs
```diff
@@ -0,0 +1,44 @@
...
+fn derive_path_hardened(key: &SecretKey, path: &[u32]) -> SecretKey {
+    let mut derived = key.derive_hardened(path[0]);
+    for idx in &path[1..] {
+        derived = derived.derive_hardened(*idx);
+    }
+    derived
+}
+
+pub fn master_to_wallet_unhardened_intermediate<Key: DerivableKey>(key: &Key) -> Key {
+    derive_path_unhardened(key, &[12381_u32, 8444, 2])
```
richardkiss: You might consider simply exporting these numbers (ie. [12381, 8444, 2]) from another function rather than hard-coding them inside a function. Call them `chia_default_wallet_path` or something more accurate.

chia-keys/src/derive_keys.rs
```diff
@@ -0,0 +1,44 @@
...
+pub fn master_to_wallet_hardened(key: &SecretKey, idx: u32) -> SecretKey {
+    derive_path_hardened(key, &[12381_u32, 8444, 2, idx])
+}
+
+pub fn master_to_pool_singleton(key: &SecretKey, pool_wallet_idx: u32) -> SecretKey {
+    derive_path_hardened(key, &[12381_u32, 8444, 5, pool_wallet_idx])
+}
+
+pub fn master_to_pool_authentication(key: &SecretKey, pool_wallet_idx: u32, idx: u32) -> SecretKey {
+    assert!(pool_wallet_idx < 10000);
```
richardkiss: ? why?

chia-keys/src/public_key.rs
```diff
@@ -0,0 +1,109 @@
...
+        let digest = hasher.finalize();
+
+        // in an ideal world, we would not need to reach for the sledge-hammer of
+        // num-bigint here. This would most likely be faster if implemented in
+        // Scalar directly.
+
+        // interpret the hash as an unsigned big-endian number
+        let mut nounce = BigUint::from_bytes_be(digest.as_slice());
+
+        let q = BigUint::from_bytes_be(&[
```
richardkiss: Consider `lazy_static` (as in `clvm_rs`) or `once_cell` here to avoid calling this function multiple times.

chia-keys/src/public_key.rs
```diff
@@ -0,0 +1,109 @@
...
+        assert_eq!(derived_pk.to_bytes(), derived_sk.public_key().to_bytes());
+    }
+}
+
+#[cfg(test)]
+use rand::Rng;
+
+#[test]
+fn test_from_bytes() {
+    let mut rng = rand::thread_rng();
```
richardkiss: Seems a little scary. You could get a failure and never be able to reproduce it. I suggest seeding the rng with a fixed value, or at least printing the seed (maybe using the current timestamp).

chia-keys/src/secret_key.rs
```diff
@@ -0,0 +1,302 @@
...
+        let mut bytes = self.0.to_bytes();
+        bytes.reverse();
+        bytes
+    }
+
+    pub fn public_key(&self) -> PublicKey {
+        PublicKey(G1Projective::generator() * self.0)
+    }
+
+    fn to_lamport_pk(&self, idx: u32) -> [u8; 32] {
```
richardkiss: Okay, this stuff I am not familiar with.

chia-keys/src/secret_key.rs
```diff
@@ -0,0 +1,302 @@
...
+
+        // in an ideal world, we would not need to reach for the sledge-hammer of
+        // num-bigint here. This would most likely be faster if implemented in
+        // Scalar directly.
+
+        // interpret the hash as an unsigned big-endian number
+        let mut scalar = BigUint::from_bytes_be(digest.as_slice());
+
+        let q = BigUint::from_bytes_be(&[
+            0x73, 0xed, 0xa7, 0x53, 0x29, 0x9d, 0x7d, 0x48, 0x33, 0x39, 0xd8, 0x08, 0x09, 0xa1,
```
richardkiss: Refactor into a `group_order.rs` file or something. We really should consider forking bls12_381 and moving this into there.

Hmm, actually, might be worth trying the already existing `from_u512` method [here](https://github.com/mikelodder7/bls12_381_plus/blame/main/src/scalar.rs#L326). Even though it has a lot of padding, it uses some clever techniques that might made it faster than classical mod. (We could also just implement a `from_u256` method.)

richardkiss: (Although I get that at this point, you're just trying to get something that works even if non-optimized.)

chia-keys/src/derive_keys.rs
arvidn: I don't see the upside of doing that. I want to keep the API as simple as possible, and as hard as possible to use the wrong derivation path. If we introduce another derivation path, I'd rather just add another function here that includes it.

chia-keys/src/derive_keys.rs
arvidn: because the last derivation index combines both `pool_wallet_idx` and `idx` into a single entry. Which means each pool wallet is only allowed to make 10000 transitions I imagine. This is how this function is defined in python, I'm not familiar with the rationale for doing this vs. just using a deeper derivation path.

chia-keys/src/public_key.rs
arvidn: using a lazy_static caused a performance regression.
```
     Running benches/derive_key.rs (/Users/arvid/Documents/dev/chia_rs/target/release/deps/derive_key-7791798e019a8a44)
secret key, unhardened  time:   [323.28 µs 326.14 µs 329.63 µs]
                        change: [+2.8152% +3.9835% +5.4750%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

One reason might be that the `%=` operator is only implemented for where the right-hand side is moved into it. So even with a constant `Q`, I still ave to `clone()` it when calling `%=`.

chia-keys/src/secret_key.rs
arvidn: this is more or less a verbatim port from `bls-signatures`

chia-keys/src/secret_key.rs
arvidn: yeah, I hope to revisit this in the future. I think I've spent more than its fair share to get this to work at all.

richardkiss (approved):

richardkiss (approved):

PR #73, opened 2022-10-31
Add some consistency to create_coin by converting it from a HashSet into a BTreeSet

AmineKhaldi:


src/gen/conditions.rs
```diff
@@ -280,7 +280,7 @@ pub fn parse_args(
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Ord, PartialOrd)]
```
arvidn, line 283: this is actually problematic. the `NodePtr` is not deterministic, but is essentially a pointer to the hint, not the hint itself

arvidn: I'll have to figure something out

arvidn:
I think this PR is outdated. we still don't preserve all conditions so we won't be able to identify identical spends this way.

PR #79, opened 2022-11-10
move tree_hash() into its own crate (clvm-utils)

arvidn:
this is a pure refactoring. I'm moving the `tree_hash()` function into its own crate `clvm-utils`. This is in preparation of adding more clvm utilities, such as curry/uncurry and computing puzzle-hashes.

### DIFF
--- a/.github/workflows/build-crate-and-npm.yml
+++ b/.github/workflows/build-crate-and-npm.yml
@@ -43,7 +43,10 @@ jobs:
     - name: build
       run: cargo build --release
     - name: dry-run of `cargo publish`
-      run: cargo publish --dry-run
+      run: |
+        cargo publish --dry-run
+        cd clvm-utils
+        cargo publish --dry-run
 
     - name: Upload crate artifacts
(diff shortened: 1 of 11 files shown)


arvidn:
CI is red now because `cargo publish --dry-run` fails, because it requires the new crate to have been uploaded to crates.io.

```
no matching package named `clvm-utils` found
  location searched: registry `crates-io`
  required by package `chia v0.1.[14](https://github.com/Chia-Network/chia_rs/actions/runs/3438238969/jobs/5734039912#step:9:15) (/home/runner/work/chia_rs/chia_rs/target/package/chia-0.1.14)`
```

If it looks good I'll upload it and re-trigger the CI

richardkiss (approved):
Looks good.

richardkiss:
Sounds to me like you want to create a crate that the wallet can use without dragging in unnecessary consensus stuff like the block running ROM (so a wallet base library). I support the idea of keeping the wallet attack surface minimal.

arvidn:
yes, that's my hope.

PR #82, opened 2022-11-12
move protocol types into its own crate

arvidn:
This is the first step. I intend to move the other protocol types from the `wheel` too, but they are a bit more complicated since they have python bindings directly


richardkiss (approved):

richardkiss:
Looks good.

PR #85, opened 2022-11-17
support unnamed fields in PyStreamable

arvidn:
This patch adds support for types with unnamed fields to the `PyStreamable` macro. fields in a tuple are unnamed fields.

Tuples are still limited in that they don't support JSON serialization/deserialization, except for a special case of 1-tuples, where they just behave as the underlying type.

In order to demonstrate that this works in CI, I also added two new types `G1Element` and `G2Element`. These are just fixed-size byte blobs that can be used as place holders in types that contain these fields. Using "dumb" storage for these punts on a gnarly problem of using the rust implementation of bls Signature and PublicKey vs. the C++ blspy implementation. The rust implementation is quite a lot slower, so we can't use it across the board. So this provides an option that can be used together with both implementations, depending on the situation.

Additionally, the `generate_type_stubs.py` script is updated to support tuples. Although, it adds phantom fields to the python object, mostly to get the constructors right. We might want to change this in the future.


chia_py_streamable_macro/src/lib.rs
```diff
@@ -38,72 +39,30 @@ pub fn py_streamable_macro(input: TokenStream) -> TokenStream {
...
-                    }
-
-                    pub fn __deepcopy__<'p>(&self, memo: &pyo3::PyAny) -> PyResult<Self> {
-                        Ok(self.clone())
-                    }
-
-                    pub fn __copy__<'p>(&self) -> PyResult<Self> {
-                        Ok(self.clone())
-                    }
-
```
arvidn, line 106: this block was moved out and down, to be reused between the two cases of named and unnamed fields. (it's called `streamable`)

richardkiss:
The obvious way to serialize `(A, B)` to json is as `[serialize(A), serialize(B)]`. That said, I don't think we should hold this patch up for this functionality if it's not needed now.

richardkiss (approved):

arvidn:
yeah, I was considering that too. I hesitated because it would also make sense to serialize `(A)` as `[serialize(A)]`, but it's a really handy feature to define a custom 1-tuple type as a "strong-type" (and I believe this is a bit of a convention in rust). i.e. acting just like the underlying type.

Maybe you could have both and just live with the irregularity.

PR #86, opened 2022-11-18
Fix python fields

arvidn:


tests/test_coin.py
```diff
@@ -389,3 +389,10 @@ def test_coin_hash() -> None:
 
     assert hash(c1) == hash(c4)
     assert type(hash(c1)) is int
+
+def test_coin_fields() -> None:
+
+    c1 = Coin(parent_coin, puzzle_hash, 1000000)
+    assert c1.parent_coin_info == parent_coin
+    assert c1.puzzle_hash == puzzle_hash
+    assert c1.amount == 1000000
```
arvidn, line 398: this test failed prior to this patch

chia_py_streamable_macro/src/lib.rs
```diff
@@ -71,6 +71,12 @@ pub fn py_streamable_macro(input: TokenStream) -> TokenStream {
                     pub fn from_json_dict(o: &pyo3::PyAny) -> PyResult<Self> {
                         <Self as FromJsonDict>::from_json_dict(o)
                     }
+
+                    #(#[getter]
+                        fn #fnames(&self, py: Python) -> PyResult<#ftypes> {
+                            Ok(self.#fnames.clone())
+                        }
+                    )*
```
arvidn, line 79: this is the part of the macro that creates the getter functions, replacing the need for `#[pyo3(get)]`

aqk: Nice. It looks like this creates getter functions for all members. You might want to add a note about expected use of mutable members if you are keeping an API guidelines doc.

aqk (approved):

chia_py_streamable_macro/src/lib.rs
arvidn: all these types we export (so far) are use the same way our `dataclass(frozen=True` are in python, i.e. immutable.
I don't expect we'll have a type where python communicates back to rust by mutating members either.

PR #87, opened 2022-11-19
improve test coverage of edge case in merkle_tree_set

arvidn:
This patch adds tests for the edge case where two hashes are only separated by bit 255. The tests are converted from the tests in `chia-blockchain`.

### DIFF
--- a/src/merkle_set.rs
+++ b/src/merkle_set.rs
@@ -91,6 +91,8 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
             // if every bit is identical, we have a duplicate value
             // duplicate values are collapsed (since this is a set)
             // so just return one of the duplicates as if there was only one
+            debug_assert!(range.len() > 1);
+            debug_assert!(range[0] == range[1]);
             (range[0], NodeType::Term)
         } else {
             // this means either the left or right bucket/sub tree was empty.
@@ -114,6 +116,22 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
                 (child_hash, child_type)
             }
         }
+    } else if depth == 255 {
+        // this is an edge case where we make it all the way down to the
+        // bottom of the tree, and split the last pair. This has the same
+        // effect as the else-block, but since we use u8 for depth, it would
+        // overflow
+        debug_assert!(range.len() > 1);
+        debug_assert!(left < range.len() as i32);
+        (
+            hash(
+                NodeType::Term,
+                NodeType::Term,
+                &range[0],
+                &range[left as usize],
+            ),
+            NodeType::MidDbl,
+        )
     } else {
         let (left_hash, left_type) = radix_sort(&mut range[..left as usize], depth + 1);
         let (right_hash, right_type) = radix_sort(&mut range[left as usize..], depth + 1);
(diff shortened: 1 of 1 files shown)


src/merkle_set.rs
```diff
@@ -91,6 +91,8 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
             // if every bit is identical, we have a duplicate value
```
aqk, line 91: Question about this comment: Why does depth 255 having only a single child imply that we have an all-right or all-left tree ("if every bit is identical") ?

Are we expected to have returned already in the other cases?

arvidn: we build the tree by comparing the hash, bit-by-bit, from the left to right. As long as a node has more than 1 node, we keep "splitting" them based on the next bit. If you end up at the *last* bit (i.e. bit 255) and they are the same, we know the hashes are duplicates.

I believe it really should be an error when this happens, because it suggests that we're trying to build a block with duplicate spends (which is forbidden by other rules). However, the original implementation in python allowed (and ignored) duplicates, so to be compatible with existing tests, I also ignore them here

src/merkle_set.rs
```diff
@@ -114,6 +116,22 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
...
+            hash(
+                NodeType::Term,
+                NodeType::Term,
+                &range[0],
+                &range[left as usize],
+            ),
+            NodeType::MidDbl,
+        )
     } else {
         let (left_hash, left_type) = radix_sort(&mut range[..left as usize], depth + 1);
```
aqk, line 136: Since we now have two implementations of `depth == 255`, it might be nice to add an assert to the `else` case to distinguish it, as well.

src/merkle_set.rs
```diff
@@ -114,6 +116,22 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
                 (child_hash, child_type)
             }
         }
+    } else if depth == 255 {
+        // this is an edge case where we make it all the way down to the
+        // bottom of the tree, and split the last pair. This has the same
+        // effect as the else-block, but since we use u8 for depth, it would
+        // overflow
```
aqk, line 123:
```suggestion
        // overflow. Thus, this case must be tested before the code in the below else-block.
```

This is a great comment. It is implied, but it might be nice to explicitly state that this block needs to come before the `else` case

src/merkle_set.rs
aqk: Ah, right. Hash equality is like an escape condition for this type of search. Thanks.

> I believe it really should be an error when this happens, because it suggests that we're trying to build a block with duplicate spends (which is forbidden by other rules). However, the original implementation in python allowed (and ignored) duplicates, so to be compatible with existing tests, I also ignore them here

You should put this in a comment, since this merkle set implementation is embedded in this specific project, (and not a general library). We might want to update the code when the Python implementation is no longer used.

src/merkle_set.rs
arvidn: I'm not quite sure I follow your addition. The edge case is run instead of the else-block, the order doesn't really matter. you could put the else block first and have the condition `depth < 255`. But perhaps I'm misunderstanding

src/merkle_set.rs
aqk: Yes, I guess that's the only difference. Not needed :)

src/merkle_set.rs
aqk: No, you are correct; what I meant is equivalent to putting `debug_assert!(depth < 255);` in the else block.

Totally not needed.

aqk (approved):

PR #88, opened 2022-11-19
add support for remaining primitive types to json serialization/deserialization

arvidn:
there were some primitives missing, implementing the traits. This patch also makes a macro to mitigate repetition.

Additionally, another earlier oversight was to not implement the `AsRef` trait for `Bytes` (but just `BytesImpl<n>`). So that's also added.

### DIFF
--- a/chia-protocol/src/to_json_dict.rs
+++ b/chia-protocol/src/to_json_dict.rs
@@ -6,23 +6,28 @@ pub trait ToJsonDict {
     fn to_json_dict(&self, py: Python) -> PyResult<PyObject>;
 }
 
-impl ToJsonDict for bool {
-    fn to_json_dict(&self, py: Python) -> PyResult<PyObject> {
-        Ok(self.to_object(py))
-    }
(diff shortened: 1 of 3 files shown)


chia-protocol/src/to_json_dict.rs
```diff
@@ -6,23 +6,28 @@ pub trait ToJsonDict {
...
-    fn to_json_dict(&self, py: Python) -> PyResult<PyObject> {
-        Ok(self.to_object(py))
-    }
-}
-
-impl ToJsonDict for u32 {
-    fn to_json_dict(&self, py: Python) -> PyResult<PyObject> {
-        Ok(self.to_object(py))
-    }
+macro_rules! from_json_primitive {
```
aqk, line 9:
```suggestion
macro_rules! to_json_primitive {
```

Did you intend to use this macro name?

arvidn: oops :)

aqk (approved):

PR #90, opened 2022-11-22
instead of using a bool field, expose a flags field

arvidn:
where one flag indicates eligible for dedup


AmineKhaldi (approved):
LGTM, thanks.

PR #93, opened 2022-11-30
Use macos-10.14 to build macOS Intel wheel

altendky:
Compare and contrast with https://github.com/Chia-Network/chia_rs/pull/94.


cmmarslender (approved):

altendky:
Going with https://github.com/Chia-Network/chia_rs/pull/94 instead I think.

PR #97, opened 2022-12-01
Forward amount parse validation errors as they are and introduce AmountExceedsMaximum

AmineKhaldi:


arvidn:
This test failure is because of the error message change. you'd need to update the test: https://github.com/Chia-Network/chia_rs/actions/runs/3596319974/jobs/6056852275#step:9:112

AmineKhaldi:
> This test failure is because of the error message change. you'd need to update the test: https://github.com/Chia-Network/chia_rs/actions/runs/3596319974/jobs/6056852275#step:9:112

Thanks to this insight I opted for adapting the code to keep the test's `ReserveFeeConditionFailed` expectation intact.

src/gen/sanitize_int.rs
```diff
@@ -32,7 +32,7 @@ pub fn sanitize_uint(
 
     // if there are too many bytes left in the value, it's too big
     if buf.len() > size_limit {
-        return Err(ValidationErr(n, code));
+        return Err(ValidationErr(n, ErrorCode::AmountExceedsMaximum));
```
arvidn, line 35: this needs a unit test

src/gen/conditions.rs
```diff
@@ -2670,7 +2683,7 @@ fn test_spend_amount_exceeds_max() {
         cond_test("((({h1} ({h2} (0x010000000000000000 ())))")
             .unwrap_err()
             .1,
-        ErrorCode::InvalidCoinAmount
+        ErrorCode::AmountExceedsMaximum
```
arvidn, line 2686: I'm thinking that it might be nice to preserve the error code for this case. This is the spend being invalid, because it's referring to a coin that cannot exist. We still convert this to `INVALID_CONDITION` when going back to python, but preserving this leaves the door open to record more precise errors.

I believe it would require mapping the error code on line 400.

src/gen/conditions.rs
```diff
@@ -2061,7 +2074,7 @@ fn test_create_coin_negative_amount() {
         cond_test("((({h1} ({h2} (123 (((51 ({h2} (-1 )))))")
             .unwrap_err()
             .1,
-        ErrorCode::InvalidCoinAmount
+        ErrorCode::NegativeAmount
```
arvidn, line 2077: +1 I would have thought this was already the error we got. This is a good improvement

src/gen/conditions.rs
```diff
@@ -2050,7 +2063,7 @@ fn test_create_coin_amount_exceeds_max() {
         cond_test("((({h1} ({h2} (123 (((51 ({h2} (0x010000000000000000 )))))")
             .unwrap_err()
             .1,
-        ErrorCode::InvalidCoinAmount
+        ErrorCode::AmountExceedsMaximum
```
arvidn, line 2066: +1

src/gen/conditions.rs
```diff
@@ -258,6 +268,9 @@ pub fn parse_args(
             match sanitize_uint(a, first(a, c)?, 4, ErrorCode::AssertHeightRelative) {
                 // Height is always positive, so a negative requirement is always true,
                 Err(ValidationErr(_, ErrorCode::NegativeAmount)) => Ok(Condition::Skip),
+                Err(ValidationErr(n, ErrorCode::AmountExceedsMaximum)) => {
+                    Err(ValidationErr(n, ErrorCode::AssertHeightRelative))
```
arvidn, line 272: Is this preserving the current behavior? If so, let's leave it like this. Otherwise I'm tempted to suggest we should turn this into an `InvalidCondition`, because it *is* an invalid condition rather than a valid condition that failed.

src/gen/conditions.rs
```diff
@@ -164,11 +164,16 @@ pub fn parse_args(
...
         }
         RESERVE_FEE => {
-            let fee = parse_amount(a, first(a, c)?, ErrorCode::ReserveFeeConditionFailed)?;
             if (flags & STRICT_ARGS_COUNT) != 0 {
                 check_nil(a, rest(a, c)?)?;
             }
-            Ok(Condition::ReserveFee(fee))
+            match parse_amount(a, first(a, c)?, ErrorCode::ReserveFeeConditionFailed) {
+                Err(ValidationErr(n, ErrorCode::NegativeAmount)) => {
+                    Err(ValidationErr(n, ErrorCode::ReserveFeeConditionFailed))
```
arvidn, line 172: to preserve the existing behavior, you would have to map `AmountExceedsMaximum` to `ReserveFeeConditionFailed` too, right?

arvidn (changes requested):
looks good. Two main comments:

1. The change to `sanitize_uint()` needs a unit test
2. I think we should preserve the error code for an invalid spend specification (i.e. where the amount of the coin we're claiming to spend exceeds maximum). I also think that the negative case for a spend specification should be translated into the `InvalidCoinAmount` error, but that's an unrelated change.

arvidn:
One way to bring down the size of this change might be to make `sanitize_uint()` take another error code indicating what error to return if the amount exceeds maximum.
That way you would not have to add so much code to map specific error messages at call sites.

arvidn:
going back to why we're making this change in the first place, I realized the *only* case we get the wrong (or an unintuitive) error code is in `CREATE_COIN` with an amount that exceeds maximum. That's the only thing that needs changing, nothing else.

One other place that I also noticed, though, is when parsing a spend. We identify which coin is being spent by its amount, if that amount is negative, we really shouldn't be getting an `AmountNegative` error, we should get an `InvalidAmount` error. The former is used for `CREATE_COIN` with a negative amount.

arvidn:
I think this is a smaller (and a bit simpler) change: https://github.com/Chia-Network/chia_rs/pull/99

AmineKhaldi:
> I think this is a smaller (and a bit simpler) change: #99
Agreed, thanks.

PR #98, opened 2022-12-03
Use updated clvm_rs.

richardkiss:


Cargo.toml
```diff
@@ -33,3 +33,6 @@ crate-type = ["rlib"]
 
 [profile.release]
 lto = true
+
+[patch.crates-io]
+clvmr = { git = "https://github.com/Chia-Network/clvm_rs.git", rev = "b69725d2" }
```
richardkiss, line 38: Once we've published to crates.io we can remove this patch.

fuzz/Cargo.toml
```diff
@@ -42,3 +42,6 @@ name = "parse-conditions"
 path = "fuzz_targets/parse-conditions.rs"
 test = false
 doc = false
+
+[patch.crates-io]
+clvmr = { git = "https://github.com/Chia-Network/clvm_rs.git", rev = "b69725d2" }
```
richardkiss, line 47: Ditto.

wheel/Cargo.toml
```diff
@@ -17,8 +17,11 @@ path = "src/lib.rs"
...
 chia-protocol = { path = "../chia-protocol", features = ["py-bindings"]  }
-clvmr = "=0.1.24"
+clvmr = "=0.2.0"
 pyo3 = { version = "=0.15.1", features = ["extension-module", "multiple-pymethods"] }
 chia_py_streamable_macro = { version= "0.1.2", path = "../chia_py_streamable_macro" }
 chia_streamable_macro = { version = "0.2.3", path = "../chia_streamable_macro" }
 hex = "=0.4.3"
+
+[patch.crates-io]
+clvmr = { git = "https://github.com/Chia-Network/clvm_rs.git", rev = "b69725d2" }
```
richardkiss, line 27: Tritto.

arvidn (dismissed):

arvidn:
it looks like there's still a version conflict of `clvm_rs`: https://github.com/Chia-Network/chia_rs/actions/runs/3606526215/jobs/6077924242#step:9:53

richardkiss:
> it looks like there's still a version conflict of `clvm_rs`: https://github.com/Chia-Network/chia_rs/actions/runs/3606526215/jobs/6077924242#step:9:53

I think this is because when you do `cargo publish`, it no longer gets sub-crates from the local file system, but instead forces a fetch from `crates.io` (because stuff uploaded to `crates.io` can only depend on other things from `crates.io` for transitive security). I expect a real `cargo publish` will succeed as long as the `clvm-utils` is published first.

However, I do have to update everyone's version number (or at least the things that depend on `clvmr` 0.20 now).

arvidn (approved):

PR #99, opened 2022-12-05
Introduce AmountExceedsMaximum error code

arvidn:
and make CREATE_COIN fail with that if it specifies a too large amount


AmineKhaldi (dismissed):
LGTM, thanks.

AmineKhaldi (approved):

PR #100, opened 2022-12-08
Wallet protocol

arvidn:
* add macros to making it more succinct to define new message types and streamable types
* add all protocol types required for the Wallet Protocol
* add chia protocol wrapper message

This patch set is suitable to be reviewed one commit at a time.


richardkiss:
Long, but seems pretty straightforward. This is the kind of code that problems will be uncovered as we try it (or via integration tests).

richardkiss (approved):

arvidn:
yes. we won't be using these for a while. I expect to have more tests by then

PR #102, opened 2022-12-09
MacOS deployment target on M1

arvidn:
lowering the deployment target across the board for all MacOS builds appears to have been a mistake. The M1 builds recently started failing by pip declaring the wheel built against the older MacOS version as not being supported.

Possibly caused by MacOS 10.4 is older than the M1 architecture.

It looks like this ticket may be related: https://github.com/actions/setup-python/issues/469

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -159,7 +159,7 @@ jobs:
     - name: Build MacOs with maturin on Python ${{ matrix.python }}
       if: matrix.os.matrix == 'macos'
       env:
-        MACOSX_DEPLOYMENT_TARGET: 10.14
+        MACOSX_DEPLOYMENT_TARGET: ${{ matrix.arch.matrix == 'intel' && '10.14' || '11.0' }}
       run: |
         python${{ matrix.python.major-dot-minor }} -m venv venv
         . venv/bin/activate


arvidn:
The remaining issue with `cargo publish --dry-run` is related to crate versions and their dependencies, which we haven't quite settled on and solved yet.

.github/workflows/build-test.yml
```diff
@@ -157,14 +157,21 @@ jobs:
           python -m pip install maturin
 
     - name: Build MacOs with maturin on Python ${{ matrix.python }}
-      if: matrix.os.matrix == 'macos'
+      if: matrix.os.matrix == 'macos' && matrix.arch.matrix == 'intel'
       env:
         MACOSX_DEPLOYMENT_TARGET: 10.14
```
altendky, line 162: How about this and avoid the duplication?  I mean, I hate this and/or form for picking values, but I'm getting used to it in CI.

```suggestion
        MACOSX_DEPLOYMENT_TARGET: ${{ matrix.arch.matrix == 'intel' && '10.14' || '' }}
```

altendky (changes requested):

.github/workflows/build-test.yml
altendky: Or maybe just always explicit?

```suggestion
        MACOSX_DEPLOYMENT_TARGET: ${{ matrix.arch.matrix == 'intel' && '10.14' || '11.0' }}
```

arvidn:
on CI, we build with `maturin` and then we do `pip install --no-index --find-links wheel/target/wheels/ chia_rs`, which I understand means we don't look at pypi at all, just in that directory. I ran a test where I instead specified the (one) wheel file that's in there. Then `pip` said it was unsupported by the platform.

.github/workflows/build-test.yml
arvidn: I'm not familiar with this syntax. It looks like it would all just evaluate to `True`. But I suppose these are bash-like operators.

arvidn: it's working

altendky: Similar to Python, you don't get a boolean out of a boolean operator.  You get one of the operands.  `"" or [1, 2, 3]` in Python results in `[1, 2, 3]` because `""` is falsey and it's an `or` so it checks the second operand which is truthy and uses that as the resulting value.  It's like the questionability of ternary operators, but more so.  But, it's either this or some JSON string games and indexing...

altendky (approved):

PR #104, opened 2022-12-12
Improve generator tooling

arvidn:
adds additional stress tests to the `test-generator` test suite. and improve the tests by using separate limits of mempool mode and consensus mode.


tests/test-generators.py
```diff
@@ -26,56 +26,80 @@ def parse_output(result, error_code) -> str:
...
 
         if sys.version_info[:2] == (3, 11):
             limit *= 1.3
 
-        if run_time > limit or run_time2 > limit:
+        if run_time > limit:
             print(f"run-time exceeds limit ({limit})!")
             failed = 1
+        if run_time2 > strict_limit:
+            print(f"run-time exceeds limit ({strict_limit})!")
```
prozacchiwawa, line 101: would it make sense to specify which limit was violated in this message?

tests/test-generators.py
```diff
@@ -26,56 +26,80 @@ def parse_output(result, error_code) -> str:
...
+    error_code2, result2 = run_gen(g, MEMPOOL_MODE, "generators/block-834752.hex")
+    run_time2 = perf_counter() - start_time
     output2 = parse_output(result2, error_code2)
 
     with open(g) as f:
         expected = f.read().split('\n', 1)[1]
         if not "STRICT" in expected:
             expected2 = expected
-            if not (result is None and result2 is None or result.cost == result2.cost):
+            if not (result is None or result2 is None or result.cost == result2.cost):
```
prozacchiwawa, line 46: I'd suggest:

    if result is not None and result2 is not None and result.cost != result2.cost:

It seems a bit easier to understand here.

tests/test-generators.py
```diff
@@ -26,56 +26,80 @@ def parse_output(result, error_code) -> str:
...
 
         compare_output(output2, expected2, "STRICT")
         print(f"  run-time: {run_time2:.2f}s")
 
+        # this is the ambition with future optimizations
         limit = 1.5
+        strict_limit = 1.5
 
         # temporary higher limits until this is optimized
         if "duplicate-coin-announce.clvm" in g:
```
prozacchiwawa, line 67: long(er) run, i might do:

    special_limits_per_test = {
        "duplicate-coin-announce.clvm": {"limit": 10, "strict_limit": 5},
        # ...
    }
    default_limits = { "limit": 1.5, "strict_limit": 1.5 }

    # ...

    filename = os.path.basename(g)
    use_limits = special_limits_per_test.get(filename, default_limits)
    limit, strict_limit = use_limits["limit"], use_limits["strict_limit"]

And since they're dict you can add more breadth.

arvidn: yeah, that's fair. I've been hoping that these exceptions are short lived (hoping to optimize `clvm` to be faster). It's not so short lived by now though

prozacchiwawa (approved):

PR #105, opened 2022-12-16
Streamable enums

arvidn:
This patch adds support for (simple) `enum`s to the `Streamable` macro (in `chia_streamable_macro`). The way it works is it iterates over all enum variants and ensures that:
* they all have an integral value assigned to them
* that the integral value fits in a `u8` variable (because that's what we use for message types)

With this new feature, it's possible to directly use the enums as fields of streamable types, so the `msg_type` and `node_type` are made proper enums.

Rust doesn't allow casting integers to enum variants (but it does support casting enums to ints), therefore, the generated `parse()` function for enums creates a large `match` statement for all integral values, returning the corresponding enum value.

The python binding macro implements py03 conversions for enums to and from python integers as well as implements the `ToJsonDict` and `FromJsonDict` to complete the support for inclusion in Stresmable types.

Since these enums are exposed to python as plain integers, so does the `generate_type_stubs.py`

### DIFF
--- a/chia-protocol/src/chia_error.rs
+++ b/chia-protocol/src/chia_error.rs
@@ -13,6 +13,7 @@ pub enum Error {
     InvalidString,
     InputTooLarge,
     SequenceTooLarge,
+    InvalidEnum,
     Custom(String),
 }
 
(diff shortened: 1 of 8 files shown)


richardkiss (dismissed):
Wow, very nice.

richardkiss (approved):

PR #108, opened 2023-01-03
make the fuzzing utility make_tree() more efficient

arvidn:
importantly not recursive, to avoid spurious stack-overflows when fuzzing


AmineKhaldi (approved):
LGTM thanks.

richardkiss (approved):

PR #112, opened 2023-01-05
bump version to 0.2.0

arvidn:


ChiaMineJP (commented):
Umm.
I've tried to build wasm but the wasm file created was only 256 bytes.
I checked `wasm/src/lib.rs` and found that the content was:
```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub fn a_test_function() -> u64 {
    500
}
```
So wasm code is not ready yet.
Do you think it blocks 0.2.0 release?

arvidn:
I don't think missing wasm bindings is a blocker for chia-blockchain 1.7.

I seem to recall you adding some bindings though, but perhaps that was for `clvm_rs`

ChiaMineJP (approved):

richardkiss (approved):

PR #117, opened 2023-01-08
Fix run_gen.py and issue in test block generator

arvidn:
The generator from block 834761, used as part of the test, had a bug. It was introduced when pulling the generator from the blockchain (in binary form) and converting it to clvm form. One of its atoms were rendered incorrectly because of this issue in `opd`: https://github.com/Chia-Network/clvm_tools/pull/97

Correcting the de-compiled clvm, produces the correct output, verified against the blockchain.


arvidn:
The change in the clvm code might not be so easy to spot. It's one of the atoms that used to be rendered as `":5)D"`, where the last byte (`0x0a`) had been truncated. The new atom is rendered as `0x3a3529440a` in the de-compiled program.

tests/run_gen.py
```diff
@@ -44,26 +44,6 @@ def run_gen(fn: str, flags: int = 0, args: Optional[str] = None):
         "ff02ffff04ff0dff8080808080ffff01ff0bffff0101ff058080ff0180ff0180"
         "80")
 
-    # this is the deserializer that's passed in to all blocks, from:
-    # https://github.com/Chia-Network/chia-blockchain/blob/main/chia/wallet/puzzles/chialisp_deserialisation.clvm.hex
-    deserializer = bytes.fromhex(
```
richardkiss, line 49: Why is this going away?

arvidn: this was a misunderstanding by me. I was confusing running the generators directly (where one would have to pass in the deserializer) versus running the generator ROM and passing in the generator. In the latter case, the ROM contains the deserializer, and is responsible for calling the generator correctly.

This test only run the generators via the generator ROM.

richardkiss (approved):

PR #118, opened 2023-01-11
in generator tests, print the full cost, including the cost-per-byte

arvidn:
This is in preparation for another patch that intends to raise the abstraction level of the `chia_rs` API to run blocks, where the generator ROM doesn't need to be passed in and where the resulting cost includes the cost-per-byte.

This prepares the tests to check the full cost, rather than just CLVM + conditions cost.


arvidn:
the key changes are in the python files. e.g. https://github.com/Chia-Network/chia_rs/pull/118/files#diff-f584716cd024cb6e7ff78c2fbc0136c01f9742ab26c49f24642262b5b61c68be

richardkiss (approved):

PR #121, opened 2023-01-12
improve error handling a bit

arvidn:


wheel/src/api.rs
```diff
@@ -90,21 +91,23 @@ pub fn get_puzzle_and_solution_for_coin<'py>(
...
+    let r = py.allow_threads(|| -> Result<(NodePtr, NodePtr), EvalErr> {
         let Reduction(_cost, result) =
             run_program(&mut allocator, dialect, program, args, max_cost)?;
         match parse_puzzle_solution(&allocator, result, find_parent, find_amount, find_ph) {
             Err(ValidationErr(n, _)) => Err(EvalErr(n, "coin not found".to_string())),
-            Ok((puzzle, solution)) => Ok((
-                node_to_bytes(&Node::new(&allocator, puzzle)).unwrap(),
-                node_to_bytes(&Node::new(&allocator, solution)).unwrap(),
-            )),
+            Ok((puzzle, solution)) => Ok((puzzle, solution)),
```
richardkiss, line 99:
```suggestion
            Ok(pair) => Ok(pair),
```

richardkiss (dismissed):

wheel/src/adapt_response.rs
```diff
@@ -8,11 +8,12 @@ use pyo3::types::PyDict;
...
 pub fn eval_err_to_pyresult<T>(py: Python, eval_err: EvalErr, allocator: Allocator) -> PyResult<T> {
     let node = Node::new(&allocator, eval_err.0);
-    let blob = node_to_bytes(&node).unwrap();
-    let msg = eval_err.1;
     let ctx: &PyDict = PyDict::new(py);
+    let msg = eval_err.1;
     ctx.set_item("msg", msg)?;
-    ctx.set_item("blob", blob)?;
+    if let Ok(blob) = node_to_bytes(&node) {
+        ctx.set_item("blob", blob)?;
```
richardkiss, line 15: You could add a `node` field that stores a `LazyNode` reference to the failed unserialized blob. You think this would be useful? It still wouldn't deserialize, but it could be peeled apart with `.pair` and `.atom`.

arvidn: yeah, that might make sense. In fact, we could probably just do that unconditionally. I don't think we use the `blob` field anywhere

richardkiss: It's just for diagnostics.

arvidn: making this a `LazyNode` seems reasonable. However, it's not trivial. The `LazyNode` code lives in the `clvm_rs` wheel. That's not published on crates.io so it would need to be moved into some common rust library to be able to pull it into `chia_rs`.

richardkiss: Then I think using the blob should be sufficient. I wasn't thinking about how this is `chia_rs`.

richardkiss (approved):

PR #122, opened 2023-01-12
run_block_generator()

arvidn:
Add a higher level function to run block generators. This includes the generator ROM program. Raising the abstraction level allows the function to return the full cost of the generator (including the byte-cost). Previously, we could only return the CLVM and conditions cost.

This patch is a step towards porting the generator ROM to rust, which would have these advantages:

1. We could run blocks with a lower memory footprint by resetting the allocator in between each puzzle
2. We could record the CLVM cost per puzzle (which helps with identical spend deduplication, which needs to track cost per spend)
3. We could fail invalid blocks faster, bu parsing and partially validating the conditions after each puzzle. i.e. we wouldn't need to run all puzzles first, and then parse and validate their output.


src/generator_rom.rs
```diff
@@ -0,0 +1,55 @@
+// the generator ROM from:
+// https://github.com/Chia-Network/chia-blockchain/blob/main/chia/wallet/puzzles/rom_bootstrap_generator.clvm.hex
+pub const GENERATOR_ROM: [u8; 737] = [
```
richardkiss, line 3: What about using https://doc.rust-lang.org/std/macro.include_bytes.html and creating a small binary file with the generator ROM?

richardkiss: Or even just including it as hex, like we did in python. (Requires `hex` library.)

wheel/src/api.rs
```diff
@@ -1,5 +1,5 @@
 use crate::compression;
-use crate::run_generator::{PySpend, PySpendBundleConditions, __pyo3_get_function_run_generator};
+use crate::run_generator::{PySpend, PySpendBundleConditions, __pyo3_get_function_run_generator, __pyo3_get_function_run_block_generator};
```
richardkiss, line 2: These imports are ugly. I take the blame for it. Ultimately (not now) we could create a function in `run_generator.rs` called `add_to_api` which takes a module and does the `m.add_function(wrap_pyfunction!(run_block_generator, m)?)?;` stuff there, so we don't have to resort to these name-mangled imports.

src/generator_rom.rs
arvidn: including it as hex would require making it a `lazy_static!`, right?

wheel/src/api.rs
richardkiss: We could also consider submodules. See https://github.com/Chia-Network/chia_rs/blob/main/wheel/src/compression.rs#L13 for an example (`import chia_rs.compression`).

wheel/src/run_generator.rs
```diff
@@ -4,6 +4,7 @@ use chia_protocol::to_json_dict::ToJsonDict;
 
 use chia::gen::conditions::{parse_spends, Spend, SpendBundleConditions};
 use chia::gen::validation_error::{ErrorCode, ValidationErr};
+use chia::gen::run_block_generator::run_block_generator as run_block_gen;
```
richardkiss, line 7: I like the name `native_run_block_generator` or `rust_run_block_generator`. But this is not a blocker.

wheel/src/run_generator.rs
```diff
@@ -142,11 +144,46 @@ pub fn run_generator(
...
         Ok((error_code, _)) => {
             // a validation error occurred
             Ok((error_code.map(|x| x.into()), None))
         }
         Err(eval_err) => eval_err_to_pyresult(py, eval_err, allocator),
     }
 }
+
+#[pyfunction]
+#[allow(clippy::borrow_deref_ref)]
```
richardkiss, line 157: This is good, gotta steal it.

richardkiss: Hmm, it didn't help me in my case. Does it actually work here?

richardkiss (approved):
I did make some comments, but I'll leave it up to you as to if you feel like implementing any of them.

src/generator_rom.rs
richardkiss: Maybe not... not sure if the compiler is smart enough to evaluate it as a constant at compile time. In any case, `include_bytes!` wouldn't need `lazy_static!` for sure.

wheel/src/run_generator.rs
arvidn: yeah, that's a better name

wheel/src/run_generator.rs
arvidn: iirc, this was to suppress a buggy clippy warning. I can't reproduce it now, so it's possible it's been fixed. I'll remove both of these

src/generator_rom.rs
arvidn: I don't think this blob will be here more than 6 months or so (perhaps I'm optimistic). We'll probably need it for the tests going forward though

richardkiss (approved):

PR #128, opened 2023-01-26
STRICT_ARGS_COUNT refactor

arvidn:
factor out the check that conditions have the right number of arguments into a function, to reduce duplicate code


richardkiss (approved):
Looks good.

PR #129, opened 2023-01-26
Parameterize tests

arvidn:
This patch set introduced a dependency on `rstest`, which comes with a few convenient macros to parameterize tests.
It then merges a few test categories into a few parameterized tests. This saves a significant amount of boiler plate code and simplifies the tests. It should also make the tests less error prone since the unique features of the tests are called out in the parameters.

The commits are suitable to be reviewed one at a time.

This patch is in preparation for adding new condition codes, avoiding the test code to blow up in size.

To double check that no test case was dropped by accident, I ran two sanity checks:

```
[main] $ cargo test | wc -l
    Finished test [unoptimized + debuginfo] target(s) in 0.25s
     Running unittests src/lib.rs (target/debug/deps/chia-d65d6a5ec578387a)
   Doc-tests chia
165
```

```
[parameterize-tests]$ cargo test | wc -l
    Updating crates.io index
   Compiling chia v0.2.1 (/home/arvid/dev/chia_rs)
    Finished test [unoptimized + debuginfo] target(s) in 4.30s
     Running unittests src/lib.rs (target/debug/deps/chia-2ddfd7de89e1d6af)
   Doc-tests chia
165
```

And:
```
cargo test | sort | tee main.log
```

comparing it to this patch:

```
cargo test | sort | tee params.log
```

After reordering rows, the files diff:

```diff
$ diff -u main.log param.log 
--- main.log    2023-01-26 18:24:37.241521597 +0100
+++ param.log   2023-01-26 18:21:13.434054127 +0100
@@ -51,39 +51,39 @@
 test gen::conditions::test_duplicate_agg_sig_me ... ok
 test gen::conditions::test_duplicate_agg_sig_unsafe ... ok
 test gen::conditions::test_duplicate_create_coin ... ok
-test gen::conditions::test_assert_coin_announce_extra_arg ... ok
-test gen::conditions::test_assert_puzzle_announces_extra_arg ... ok
-test gen::conditions::test_create_coin_announce_extra_arg ... ok
-test gen::conditions::test_create_puzzle_announces_extra_arg ... ok
-test gen::conditions::test_single_seconds_relative_extra_arg ... ok
-test gen::conditions::test_single_seconds_absolute_extra_arg ... ok
-test gen::conditions::test_single_assert_my_amount_extra_arg ... ok
-test gen::conditions::test_single_assert_my_coin_id_extra_arg ... ok
-test gen::conditions::test_single_assert_my_parent_coin_id_extra_arg ... ok
[...]

PR #130, opened 2023-01-26
parameterize more tests

arvidn:
Specifically, the tests that adds multiple instances of the same condition and tests how they are aggregated (it's either MAX or SUM right now).

Also the test that has a condition without any arguments, to ensure that error is caught.

This is a follow-up on https://github.com/Chia-Network/chia_rs/pull/129 with a few missed opportunities.

### DIFF
--- a/src/gen/conditions.rs
+++ b/src/gen/conditions.rs
@@ -1134,38 +1134,36 @@ fn test_single_condition(
     test(&conds, &spend);
 }
 
-#[test]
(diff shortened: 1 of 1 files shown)


aqk (approved):

richardkiss (approved):
Looks good

richardkiss:
One test failed (py3.11) so I'm rerunning it.

PR #131, opened 2023-01-27
add support for parsing ASSERT_BEFORE_

arvidn:
This adds support for parsing 4 new condition codes, `ASSERT_BEFORE_{HEIGHT|SECONDS}_{ABSOLUTE|RELATIVE}`.

These conditions are similar to the existing height/seconds asserts but instead of asserting the block height is above the limit, it asserts it's *below* the limit. This affects how multiple instances of the same condition are aggregated in the parsed state. These conditions preserve the *lowest* value, since that's the most restrictive.

There's a new flag, `ENABLE_ASSERT_BEFORE` which needs to be passed in to the conditions parser to enable support for the new conditions. This flag is necessary in order to soft-fork in the conditions while preserving existing behavior until it activates. e.g. if you include one of these conditions but with an invalid parameter, parsing will fail. The behavior prior to supporting these conditions, any unknown conditions would be ignored without parsing the argument.


arvidn:
The aspect of this patch I'm not entirely sure about is the flag to enable support (`ENABLE_ASSERT_BEFORE`). It adds a fair amount of complexity, but mostly test cases (not every combination of flags are tested today anyway).

The reason this flag is necessary is quite subtle, and it's about avoiding accidentally activating the soft-fork early. Even though this is just parsing, and we (in theory) could decide whether to honor the semantics of the conditions or not at a later point (to activate the soft-fork), the problem is that just parsing, and possibly failing to parse, is a change in behavior.

An alternative approach would be to have parse failures be ignored and the whole condition is just treated as if it's not there. However, after the soft-fork has activated, presumably we would want the parsing to be strict, so we would still need some flag to indicate wither we are before or after activation.

richardkiss:
Not relevant to this patch, but I worry that a not care hack to the mempool might lock up the chain if all farmers suffer from the same bug of not ejecting no-longer-valid spends from the mempool before trying to make a block.

src/gen/validation_error.rs
```diff
@@ -73,6 +77,11 @@ impl From<ErrorCode> for u32 {
             ErrorCode::InvalidCoinAmount => 10,
             ErrorCode::AssertHeightAbsolute => 14,
             ErrorCode::AssertHeightRelative => 13,
+            // TODO: introduce specific error codes on the python side
+            ErrorCode::AssertBeforeSecondsAbsolute => 117,
+            ErrorCode::AssertBeforeSecondsRelative => 117,
+            ErrorCode::AssertBeforeHeightAbsolute => 117,
+            ErrorCode::AssertBeforeHeightRelative => 117,
```
richardkiss, line 84: I think it's okay to just declare them all as new values here and then we can force python to be compliant later.

richardkiss (approved):
Seems reasonable to me.

src/gen/validation_error.rs
arvidn: ok, I'll do that. I was worried about forgetting to add those error codes to python later, but I've already started a "bump-chia_rs" branch, locally, where I'm collecting these kinds of things.

richardkiss (approved):

PR #132, opened 2023-01-27
factor out the LIMIT_HEAP logic into a function, to allow reusing it

arvidn:


wheel/src/api.rs
```diff
@@ -86,10 +86,10 @@ pub fn get_puzzle_and_solution_for_coin<'py>(
     find_amount: u64,
     find_ph: Bytes32,
 ) -> PyResult<(&'py PyBytes, &'py PyBytes)> {
-    let mut allocator = Allocator::new();
+    let mut allocator = make_allocator(LIMIT_HEAP);
     let program = node_from_bytes(&mut allocator, program)?;
     let args = node_from_bytes(&mut allocator, args)?;
-    let dialect = &ChiaDialect::new(NO_NEG_DIV);
+    let dialect = &ChiaDialect::new(NO_NEG_DIV | LIMIT_STACK);
```
aqk, line 92: Is this line that adds `LIMIT_STACK` a behaviour change?

arvidn: yes, it does technically change the behavior. It brings it line with what mempool and consensus will do after the upcoming softfork. The long term solution is to make the flags be passed in by the caller. I want to make that change (that's how I discovered this).

Also, this does not affect consensus, it's used in one of the full node requests.

aqk (approved):

PR #134, opened 2023-02-01
add support for ASSERT_CONCURRENT_PUZZLE

arvidn:
This patch adds a new (optional) condition code; `ASSERT_CONCURRENT_PUZZLE`. It ensures that one of the coins being spent in the spend bundle/block has the specified puzzle hash.

This condition is also enabled by the `ENABLE_ASSERT_BEFORE` flag, since it will be part of the same soft-fork.

This patch also enables these new conditions (conditioned by `ENABLE_ASSERT_BEFORE`) in the fuzzers, to include them in the coverage.


fuzz/fuzz_targets/parse-conditions.rs
```diff
@@ -5,12 +5,12 @@ use clvmr::allocator::Allocator;
 use chia::fuzzing_utils::{BitCursor, make_tree};
 use chia::gen::conditions::parse_spends;
 
-use chia::gen::flags::{COND_ARGS_NIL, STRICT_ARGS_COUNT, NO_UNKNOWN_CONDS};
+use chia::gen::flags::{COND_ARGS_NIL, STRICT_ARGS_COUNT, ENABLE_ASSERT_BEFORE};
```
richardkiss, line 8: Why remove `NO_UNKNOWN_CONDS`?

arvidn: good question. I believe I was thinking that it would increase the test surface of the fuzzer. Thinking about it again though, it probably just shifts it a little bit from the failure case to the ignore-case. I realize I ended up removing a fuzzing case. I'll restore this.

arvidn: I believe this was a copy-paste error actually. fixed now

richardkiss (approved):

PR #135, opened 2023-02-01
add more checks for impossible constraints

arvidn:
If a single spend bundle has an `ASSERT_BEFORE_HEIGHT_ABSOLUTE` that's lower than or equal to another `ASSERT_HEIGHT_ABSOLUTE`, the spend bundle will never be valid.

This patch checks for cases like this. Also for `_SECONDS_ABSOLUTE`.

The `_RELATIVE` checks can only be made within the same Spend, since we don't actually know any absolute timestamps or confirmation heights for the coins.

This is to catch a few obvious invalid cases early.

This still leaves non-obvious cases as successful (and must fail at a later state in the mempool pipeline).

For example, two *different* coins can have `ASSERT_HEIGHT_RELATIVE` and `ASSERT_BEFORE_HEIGHT_RELATIVE` that turn out to conflict, but we'll only know that once we look those coins up from the coin DB and know their confirmation heights.

Another, more subtle, case of invalid spend bundles is where an `ASSERT_SECONDS_ABSOLUTE` conflicts with an `ASSERT_BEFORE_HEIGHT_ABSOLUTE`. In `chia_rs` we don't have any mapping between block height and (predicted) timestamps. Clearly we couldn't reject spend bundles based on predictions anyway, but there may be cases where a `SECONDS` timestamp is known to refer to an old, existing, block. We won't catch those cases here either.

### DIFF
--- a/src/gen/conditions.rs
+++ b/src/gen/conditions.rs
@@ -543,6 +543,17 @@ fn parse_spend_conditions(
             Condition::AssertSecondsRelative(s) => {
                 // keep the most strict condition. i.e. the highest limit
                 spend.seconds_relative = max(spend.seconds_relative, s);
+                if let Some(bs) = spend.before_seconds_relative {
+                    if bs <= spend.seconds_relative {
+                        // this spend bundle requres to be spent *before* a
+                        // timestamp and also *after* a timestamp that's the
+                        // same or later. that's impossible.
+                        return Err(ValidationErr(
+                            c,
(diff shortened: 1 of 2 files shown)


richardkiss (approved):
Seems reasonable. The repeated code feels a bit wrong, but I suppose in this case it's okay. 

Alternative is we could just check every time through the loop, or just once at the end of conditions processing. I don't see a case where exiting early buys us much.

PR #138, opened 2023-02-08
add new function run_puzzle()

arvidn:
This function runs a puzzle given a solution, and parses the resulting conditions. This is meant to be used for `SpendBundle` and `CoinSpend` handling.

This PR is split up into 3 commits:

1. add implicit conversion of `ValidationErr` to `PyErr` to simplify the python binding
2. split up the `parse_spend_conditions()` function into `parse_single_spend()` and `parse_conditions()` functions, since for `run_puzzle()` we just want to parse the conditions. In order to keep the interface between these functions simpler, the internal `Spend` type was amended with parent-coin-id and amount. This commit also splits the fuzzer into two separate targets.
3. add the `run_puzzle()` function along with some (python) tests.


richardkiss (approved):
Looks good.

PR #140, opened 2023-02-15
fix edge cases in assert_before parsing

arvidn:
Specifically, assert_before a negative value is an automatic fail. assert_before a value exceeding the max is an automatic pass.

The two functions used to parse block height and timestamps (seconds) for the original `ASSERT_HEIGHT_*` and `ASSERT_SECONDS_*` has edge-case behavior tailored to those conditions. Those conditions requires the block height or timestamp to exceed the specified limit. If the limit is 0 or negative, the condition is always true. If the limit exceeds the max value however (uint32 and uint64 respectively), the conditions always fail.

Since the `ASSERT_BEFORE_*` conditions have the inverse semantics, the edge case behavior also need to be inverse. This patch introduces `parse_positive_height()` and `parse_positive_seconds()` with the inverse behavior for values outside the limits.

The bulk of this patch is to fix up the unit tests to explicitly cover these cases, both for the existing `parse_height()` and the new `parse_positive_height()` (as well as the "seconds" counterpart).

### DIFF
--- a/src/gen/conditions.rs
+++ b/src/gen/conditions.rs
@@ -1,7 +1,7 @@
 use super::coin_id::compute_coin_id;
 use super::condition_sanitizers::{
-    parse_amount, parse_create_coin_amount, parse_height, parse_seconds, sanitize_announce_msg,
-    sanitize_hash,
+    parse_amount, parse_create_coin_amount, parse_height, parse_positive_height,
+    parse_positive_seconds, parse_seconds, sanitize_announce_msg, sanitize_hash,
 };
 use super::opcodes::{
     parse_opcode, ConditionOpcode, AGG_SIG_COST, AGG_SIG_ME, AGG_SIG_UNSAFE, ALWAYS_TRUE,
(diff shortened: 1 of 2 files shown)


src/gen/conditions.rs
```diff
@@ -276,23 +276,31 @@ pub fn parse_args(
...
+            }
         }
         ASSERT_BEFORE_HEIGHT_ABSOLUTE => {
             maybe_check_args_terminator(a, c, flags)?;
-            let height = parse_height(a, first(a, c)?, ErrorCode::AssertBeforeHeightAbsolute)?;
-            Ok(Condition::AssertBeforeHeightAbsolute(height))
+            match parse_positive_height(a, first(a, c)?, ErrorCode::AssertBeforeHeightAbsolute)? {
+                None => Ok(Condition::Skip),
+                Some(height) => Ok(Condition::AssertBeforeHeightAbsolute(height)),
+            }
```
arvidn, line 303: these 4 hunks are the main changes to production code, altering the behavior of the new conditions

src/gen/conditions.rs
```diff
@@ -1325,52 +1333,20 @@ fn test_extra_arg(
...
-    );
-}
-
-#[cfg(test)]
-#[rstest]
-#[case(ASSERT_HEIGHT_ABSOLUTE, ErrorCode::AssertHeightAbsolute)]
-#[case(ASSERT_HEIGHT_RELATIVE, ErrorCode::AssertHeightRelative)]
-#[case(ASSERT_BEFORE_HEIGHT_ABSOLUTE, ErrorCode::AssertBeforeHeightAbsolute)]
-#[case(ASSERT_BEFORE_HEIGHT_RELATIVE, ErrorCode::AssertBeforeHeightRelative)]
-fn test_height_exceed_max(#[case] condition: ConditionOpcode, #[case] expected_error: ErrorCode) {
```
arvidn, line 1353: these two tests were replaced by the tests at the bottom of the file, and generalized. Not only do they cover the case of exceeding the upper limit, but also the lower limit

richardkiss (approved):
Seems reasonable.

PR #142, opened 2023-02-23
additions_amount and removals_amounts

arvidn:
since the conditions parser and validator already computes additions- and removals amounts, expose it to the caller in SpendBundleConditions. These values sometimes need to be recomputed otherwise.

By exposing these fields, it will be cheaper and simpler to query the `SpendBundleConditions` for the fee paid by a spend bundle, which currently is re-computed every time it's needed. It would also allow us to remove the `fee` field from `MempoolItem` ( related to https://github.com/Chia-Network/chia-blockchain/pull/14652 )


src/gen/run_puzzle.rs
```diff
@@ -29,16 +29,10 @@ pub fn run_puzzle(
...
-        spends: Vec::new(),
-        reserve_fee: 0,
-        height_absolute: 0,
-        seconds_absolute: 0,
-        before_height_absolute: None,
-        before_seconds_absolute: None,
-        agg_sig_unsafe: Vec::new(),
-        cost: 0,
+        removal_amount: amount as u128,
+        ..Default::default()
```
richardkiss, line 33: I'd never seen this syntax before.

richardkiss (approved):
Excellent.

src/gen/run_puzzle.rs
arvidn: me neither. clippy suggested it.

PR #144, opened 2023-03-01
add support for ASSERT_MY_BIRTH_HEIGHT and ASSERT_MY_BIRTH_SECONDS

arvidn:
These conditions ensure that the coin asserting were *confirmed* at a specific block height, or in a block with a specific timestamp.

The condition codes are:
```
pub const ASSERT_MY_BIRTH_SECONDS: ConditionOpcode = 74;
pub const ASSERT_MY_BIRTH_HEIGHT: ConditionOpcode = 75;
```

Since this validation can't be done in the parsing step, it's forwarded in the `Spend` object as two new optional integers.

There are a few cases that need to be validated however. If the same spend asserts *different* heights or *different* timestamps, that's an immediate failure.

If the same spend asserts the *same* height multiple times, that's fine, we just forward that one height (or timestamp).


src/gen/conditions.rs
```diff
@@ -653,6 +692,24 @@ pub fn parse_conditions(
                     return Err(ValidationErr(c, ErrorCode::AssertMyAmountFailed));
                 }
             }
+            Condition::AssertMyBirthSeconds(s) => {
+                // if this spend already has a birth_seconds assertion, it's an
+                // error if it's different from the new birth assertion. One of
+                // them must be false
+                if spend.birth_seconds.map(|v| v == s) == Some(false) {
```
richardkiss, line 699: Very clever

src/gen/opcodes.rs
```diff
@@ -107,196 +111,56 @@ fn opcode_tester_with_assert_before(a: &mut Allocator, val: &[u8]) -> Option<Con
...
-    assert_eq!(
-        opcode_tester_with_assert_before(&mut a, &[ASSERT_HEIGHT_ABSOLUTE, 0]),
-        None
-    );
-    assert_eq!(
-        opcode_tester_with_assert_before(&mut a, &[0, ASSERT_HEIGHT_ABSOLUTE]),
-        None
-    );
-    assert_eq!(opcode_tester_with_assert_before(&mut a, &[0]), None);
+#[cfg(test)]
```
richardkiss, line 117: Oh boy. I hope this is all the same meaning (I'm not that familiar with `rtest`, but it looks like you are!).

richardkiss (approved):
Seems reasonable.

src/gen/opcodes.rs
arvidn: I split out the parameterization into its own commit, to make it a bit cleaner at least.

PR #146, opened 2023-03-06
preserve assert_seconds_relative 0

arvidn:
(which is a no-op pre 1.8.0-softfork).

This patch also fixes an issue with including `ENABLE_ASSERT_BEFORE` in the `MEMPOOL_MODE` flags.


src/gen/conditions.rs
```diff
@@ -282,8 +282,15 @@ pub fn parse_args(
         }
         ASSERT_SECONDS_RELATIVE => {
             maybe_check_args_terminator(a, c, flags)?;
-            let seconds = parse_seconds(a, first(a, c)?, ErrorCode::AssertSecondsRelative)?;
```
arvidn, line 285: `parse_seconds()` returns a `u64` and will return 0 if the value is negative (to indicate a no-op). With this patch, we want to preserve a condition of 0, and distinguish it from a negative (no-op) value.

tests/run_gen.py
```diff
@@ -56,7 +56,7 @@ def print_spend_bundle_conditions(result) -> str:
 
         if s.height_relative:
             ret += f"  ASSERT_HEIGHT_RELATIVE {s.height_relative}\n"
-        if s.seconds_relative > 0:
+        if s.seconds_relative:
```
AmineKhaldi, line 59: These are optionals so we would check `is not None` here.

AmineKhaldi (commented):
Looks good, only one tiny change.

AmineKhaldi (dismissed):

src/gen/conditions.rs
```diff
@@ -282,8 +282,15 @@ pub fn parse_args(
...
-            Ok(Condition::AssertSecondsRelative(seconds))
+            match sanitize_uint(a, first(a, c)?, 8, ErrorCode::AssertSecondsRelative) {
+                // Height is always positive, so a negative requirement is always true,
```
aqk, line 286:
```suggestion
                // A negative argument for ASSERT_SECONDS_RELATIVE is always true
```

src/gen/conditions.rs
```diff
@@ -388,13 +395,12 @@ pub struct Spend {
...
     // conditions
-    // all these integers are initialized to 0, which also means "no
-    // constraint". i.e. a 0 in these conditions are inherently satisified and
-    // ignored. 0 (or negative values) are not passed up to the next layer
-    // One exception is height_relative, where 0 *is* relevant.
-    // the highest height/time conditions (i.e. most strict)
+    // all these integers are initialized to None, which also means "no
+    // constraint". A 0 may impose a constraint (post 1.8.0 soft-fork).
+    // for height_relative, 0 has always imposed a constraint (for ephemeral spends)
+    // negative values ignored and not passed up to the next layer.
```
aqk, line 401:
```suggestion
    // negative values are ignored and not passed up to the next layer.
```

src/gen/conditions.rs
```diff
@@ -388,13 +395,12 @@ pub struct Spend {
...
     pub coin_id: Arc<Bytes32>,
     // conditions
-    // all these integers are initialized to 0, which also means "no
-    // constraint". i.e. a 0 in these conditions are inherently satisified and
-    // ignored. 0 (or negative values) are not passed up to the next layer
-    // One exception is height_relative, where 0 *is* relevant.
-    // the highest height/time conditions (i.e. most strict)
+    // all these integers are initialized to None, which also means "no
+    // constraint". A 0 may impose a constraint (post 1.8.0 soft-fork).
+    // for height_relative, 0 has always imposed a constraint (for ephemeral spends)
```
aqk, line 400: This is probably a good place to drop a comment saying what happens during `(ASSERT_HEIGHT_RELATIVE 0)`

aqk (dismissed):

aqk (approved):

AmineKhaldi (approved):

PR #148, opened 2023-03-09
sanitize_uint improvement

arvidn:
this patch separates overflow errors from invalid integer representation errors in the `sanitize_uint()` function. The overflow behaviors are almost always subject to customizations, depending on who's calling it. The invalid integer representations on the other hand, should always just propagate up.

This simplifies the call sites and removes one intermediate layer of functions that currently customize the overflow behavior. Now it's done at each call site.

### DIFF
--- a/fuzz/fuzz_targets/sanitize-uint.rs
+++ b/fuzz/fuzz_targets/sanitize-uint.rs
@@ -1,0 +1,48 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+use clvmr::allocator::Allocator;
+use chia::gen::sanitize_int::{sanitize_uint, SanitizedUint};
+use chia::gen::validation_error::{ErrorCode, ValidationErr};
+
+fuzz_target!(|data: &[u8]| {
+    let mut a = Allocator::new();
+    let atom = a.new_atom(data).unwrap();
+    match sanitize_uint(&a, atom, 8, ErrorCode::InvalidCoinAmount) {
+        Ok(SanitizedUint::Ok(_)) => {
+            assert!(data.len() <= 9);
+            if data.len() == 9 {
+                assert!(data[0] == 0);
+            }
+        },
+        Ok(SanitizedUint::NegativeOverflow) => {
+            assert!(data.len() > 0 && (data[0] & 0x80) != 0);
+        },
+        Ok(SanitizedUint::PositiveOverflow) => {
+            assert!(data.len() > 8);
+        },
+        Err(ValidationErr(n, c)) => {
+            assert!(n == atom);
+            assert!(c == ErrorCode::InvalidCoinAmount);
+        }
+    }
+
+    match sanitize_uint(&a, atom, 4, ErrorCode::InvalidCoinAmount) {
+        Ok(SanitizedUint::Ok(_)) => {
+            assert!(data.len() <= 5);
+            if data.len() == 5 {
+                assert!(data[0] == 0);
+            }
+        },
+        Ok(SanitizedUint::NegativeOverflow) => {
+            assert!(data.len() > 0 && (data[0] & 0x80) != 0);
+        },
+        Ok(SanitizedUint::PositiveOverflow) => {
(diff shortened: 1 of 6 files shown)


src/gen/conditions.rs
```diff
@@ -253,94 +249,103 @@ pub fn parse_args(
...
-                    Err(ValidationErr(n, ErrorCode::AssertSecondsRelative))
-                }
-                Err(r) => Err(r),
-                Ok(r) => Ok(Condition::AssertSecondsRelative(u64_from_bytes(r))),
+            let node = first(a, c)?;
+            let code = ErrorCode::AssertSecondsRelative;
+            match sanitize_uint(a, node, 8, code)? {
+                SanitizedUint::PositiveOverflow => Err(ValidationErr(node, code)),
+                SanitizedUint::NegativeOverflow => Ok(Condition::Skip),
+                SanitizedUint::Ok(r) => Ok(Condition::AssertSecondsRelative(r)),
```
AmineKhaldi, line 277: In general, having `x` and `before x` next to each other helps highlighting the main difference in their arg parsing here (the positive and negative overflow negation) which would improve readability.

src/gen/conditions.rs
```diff
@@ -2104,7 +2117,7 @@ fn test_create_coin_amount_exceeds_max() {
         cond_test("((({h1} ({h2} (123 (((51 ({h2} (0x010000000000000000 )))))")
             .unwrap_err()
             .1,
-        ErrorCode::AmountExceedsMaximum
+        ErrorCode::InvalidCoinAmount
```
AmineKhaldi, line 2120: `ErrorCode::AmountExceedsMaximum` and `ErrorCode::NegativeAmount` map to some logic in `chia-blockchain` so it would be worth preserving them.

AmineKhaldi (commented):
Looks good! I left some comments.

src/gen/conditions.rs
arvidn: I restored these error codes

src/gen/conditions.rs
```diff
@@ -253,94 +249,103 @@ pub fn parse_args(
...
+                SanitizedUint::NegativeOverflow => Ok(Condition::Skip),
+                SanitizedUint::Ok(r) => Ok(Condition::AssertHeightAbsolute(r as u32)),
+            }
         }
         ASSERT_BEFORE_SECONDS_RELATIVE => {
             maybe_check_args_terminator(a, c, flags)?;
-            match parse_positive_seconds(a, first(a, c)?, ErrorCode::AssertBeforeSecondsRelative)? {
-                None => Ok(Condition::Skip),
-                Some(seconds) => Ok(Condition::AssertBeforeSecondsRelative(seconds)),
+            let node = first(a, c)?;
```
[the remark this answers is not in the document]
arvidn: I think the new `match` block is so much clearer that the comments are redundant

src/gen/conditions.rs
arvidn: I'm hesitant to move code in the same PR as I update it. but it's a good point that it would probably hightlight the differences better

fuzz/fuzz_targets/sanitize-uint.rs
```diff
@@ -0,0 +1,29 @@
...
+use libfuzzer_sys::fuzz_target;
+
+use clvmr::allocator::Allocator;
+use chia::gen::sanitize_int::{sanitize_uint, SanitizedUint};
+use chia::gen::validation_error::{ErrorCode, ValidationErr};
+
+fuzz_target!(|data: &[u8]| {
+    let mut a = Allocator::new();
+    let atom = a.new_atom(data).unwrap();
+    match sanitize_uint(&a, atom, 8, ErrorCode::InvalidCoinAmount) {
```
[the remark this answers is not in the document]
arvidn: done

aqk:
`Result<SanitizedUint, ValidationErr>` is so much better than `<&[u8], ValidationErr>`

aqk (approved):

PR #149, opened 2023-03-10
add assert_ephemeral condition

arvidn:


src/gen/conditions.rs
```diff
@@ -2412,6 +2461,17 @@ fn test_duplicate_create_coin() {
     );
 }
 
+#[test]
+fn test_duplicate_create_coin_with_hint() {
```
aqk, line 2465: Do we support more than one extra argument as "hints" ?

src/gen/conditions.rs
```diff
@@ -3311,3 +3371,119 @@ fn test_multiple_my_birth_assertions(
 
     test(spend);
 }
+
+#[test]
+fn test_assert_ephemeral() {
+    // ASSERT_EPHEMERAL
+    // the coin11 value is the coinID computed from (H1, H1, 123). The first
+    // coin we spend in this case
```
aqk, line 3379:
```suggestion
    // the coin11 value is the coinID computed from (H1, H1, 123). 
    // coin11 is the first coin we spend in this case.
```

What you had was correct - I find this easier to read because the sentence was running into the "51 is ..." comment below.

src/gen/conditions.rs
```diff
@@ -3311,3 +3371,119 @@ fn test_multiple_my_birth_assertions(
...
+           ((76 ) \
+           ))\
+       ))";
+    // we don't expect any error
+    let (a, conds) = cond_test(test).unwrap();
+
+    // just make sure there are no constraints
+    assert_eq!(conds.agg_sig_unsafe.len(), 0);
+    assert_eq!(conds.reserve_fee, 0);
+    assert_eq!(conds.cost, CREATE_COIN_COST);
```
aqk, line 3395: These 3 lines would make a great named test function if you use them elsewhere.

src/gen/conditions.rs
```diff
@@ -718,6 +737,13 @@ pub fn parse_conditions(
                 }
                 spend.birth_height = Some(h);
             }
+            Condition::AssertEphemeral => {
```
aqk, line 740: Hey, when we parse conditions and put them into the `state` set, do we store ancillary data like hints in another table? Does it depend on the `Condition`?

What I'm getting at is, is each of the conditions passed up (IF it is passed up), identical to the `Condition` that came out of the `Spend` ?

src/gen/conditions.rs
```diff
@@ -479,14 +487,21 @@ pub struct ParseState {
 
     // all coin IDs that have been spent so far. When we parse a spend we also
     // compute the coin ID, and stick it in this set. It's reference counted
```
AmineKhaldi, line 489: set => map?

src/gen/conditions.rs
arvidn: we do. This test is somewhat unrelated to the PR, but I noticed it was missing. This test is demonstrating that the hint field of `NewCoin` is not considered when looking for duplicate outputs. i.e. that we create two coins, where only one of them includes a hint.

arvidn: by "support" I mean: we allow them, but we only care about the first one (for now)

aqk: Gotcha!

src/gen/conditions.rs
arvidn: there is no other table. We store just what the full node needs. So the create coin hint is one of those things, but not memos. The conditions parser is quite full-node centric right now, it deliberately drops any information that's not strictly required for consensus (with the exception of the coin hint, which is required for a full-node service).

I don't think I understand what you mean. The `Condition` enum (aka variant) are the parsed-out conditions and their arguments. Consensus rules allows any number of extra arguments to the conditions, but we only keep the ones whose meaning is defined.

There's a next step where these conditions are interpreted and collated/deduplicated. Since many coins may have (different) `ASSERT_HEIGHT_ABSOLUTE` conditions, we really only care about the most restrictive one. So we gather up all the most restrictive conditions and return that back up (in `SpendBundleConditions`, which contains a list of `Spend` object with conditions for each coin we're spending).

src/gen/conditions.rs
arvidn: well, they aren't really that common. The comment is a bit misleading, suggesting that these lines ensure there are no constraints. I'll remove it.

arvidn:
@aqk This is the positive test: https://github.com/Chia-Network/chia_rs/pull/149/files#diff-f80b284fd6e11b143a06f56f735b48c7901525f11ff62bed4eb15fc1aa08a1ccR3376 (line 3376).

> but I don't see a test for a valid spend, but spent in a non-ephemeral context.

All the negative tests fits this description, except the `ASSERT_EPHEMERAL` conditions makes the spend invalid. Or are you talking about a positive test that doesn't use `ASSERT_EPHEMERAL`?

Keep in mind that at this point we don't have a blockchain database or the coin set. We can only assume that spends are valid at this point. Once we have the blockchain and the coins we validate this. There are a few ways we can *know* that a spend is invalid. e.g. if we encounter two spends of the same coin (`DOUBLE_SPEND`).

The way `ASSERT_EPHEMERAL` works is by ensuring that the coin ID, of the coin being spent, was created by some other spend part of the same spend bundle/block. A spend is specified by the components that the coin ID are computed from, (parent coin ID, puzzle hash, amount). For every spend, we compute its coin ID (we use it for checking double spends, `CREATE_COIN_ANNOUNCE` and `ASSERT_MY_ID`). To check whether a spend is of an ephemeral coin, we just need to use that same lookup table to look up the parent coin. If we don't find it, it's not ephemeral. If we find it, we look at all the coins that spend created to ensure our puzzle-hash and amount is in there. If we don't find it, it's not an ephemeral coin.

So, there are 3 ways to fail the ephemeral check:

1. our parent coint ID was not spent in this block
2. our parent coin ID *was* spent, but the coin it created doesn't match our puzzle-hash
3. our parent coin ID *was* spent, but the coin it created doesn't match our amount

In case 2 and 3, we can be certain that the spend bundle is invalid regardless of the `ASSERT_EPHEMERAL`, since the spend cannot be of a valid coin. In case 1, the coin it's attempting to spend may very well be valid, we cannot know. However, we can know that it's not ephemeral in this spend bundle, so the assertion fails.

aqk (approved):

AmineKhaldi (approved):

PR #150, opened 2023-03-15
assert not ephemeral

arvidn:
This patch adds a feature flag to disallow relative timelock conditions on ephemeral spends. As relative timelock, we count:
```
ASSERT_SECONDS_RELATIVE
ASSERT_HEIGHT_RELATIVE
ASSERT_BEFORE_SECONDS_RELATIVE
ASSERT_BEFORE_HEIGHT_RELATIVE
ASSERT_MY_BIRTH_SECONDS
ASSERT_MY_BIRTH_HEIGHT
```

These are condition that, indirectly, depend on the current block's height and timestamp. Since we're only considering ephemeral spends.

This patch is split up into 3 commits, best reviewed one at a time.

## commit 1

The first commit is really a clean-up of the previous `ASSERT_EPHEMERAL` condition, where instead of recording (parent, puzzle-hash, amount) for coins asserting they're ephemeral, we just record the *index* of that spend, in the list of spends we're building while parsing. This in the periphery of the core change, but since we'll need to also record coins that are not allowed to be ephemeral, having a simpler and more compact representation helps.

## commit 2

This first step records whether a spend contains *any* relative timelocks. One interesting case is when a condition is automatically true, and dropped. This requires that we record the *existence* of the condition some other way. We do this with a flag. This affects some tests that currently check the flag, that now need to take into the account that this new flag may be set.

## commit 3

This last step also adds the checking that spends with relative timelock conditions are in fact not ephemeral spends. For this we also need a new feature flag to enable this checking, since consensus rules may not be updated immediately, but needs a soft-fork height where it activates. We can activate it immediately in mempool mode however.


src/gen/conditions.rs
```diff
@@ -3530,3 +3564,104 @@ fn test_assert_ephemeral_wrong_parent() {
...
+#[case(ASSERT_BEFORE_SECONDS_ABSOLUTE, None)]
+#[case(
+    ASSERT_BEFORE_SECONDS_RELATIVE,
+    Some(ErrorCode::EphemeralRelativeCondition)
+)]
+fn test_relative_condition_on_ephemeral(
+    #[case] condition: ConditionOpcode,
+    #[case] mut expect_error: Option<ErrorCode>,
+    #[values(0, ENABLE_ASSERT_BEFORE)] enable_assert_before: u32,
+    #[values(0, NO_RELATIVE_CONDITIONS_ON_EPHEMERAL)] no_rel_conds_on_ephemeral: u32,
```
arvidn, line 3592: this is a feature of `rstest` which will test every combination of these parameters

AmineKhaldi (approved):
Looks reasonable.

aqk (approved):

PR #152, opened 2023-04-18
log block timestamps and delta time

arvidn:
This extends the chain analysis with block timestamps

### DIFF
--- a/chia-tools/graph.gnuplot
+++ b/chia-tools/graph.gnuplot
@@ -28,6 +28,11 @@ plot "chain-resource-usage.dat" using 1:11 with dots title "block decompress + p
 "chain-resource-usage.dat" using 1:12 with dots title "block reference lookup", \
 "chain-resource-usage.dat" using 1:14 with dots title "conditions`
 
+set output "block-wall-clock-delta.png"
+set title "transaction block timestamp delta"
+set xlabel "s"
+plot "chain-resource-usage.dat" using 1:16 with dots title "block timestamp delta", \
+
 set output "block-cost-vs-time.png"
 set title "block CLVM execution time versus CLVM cost"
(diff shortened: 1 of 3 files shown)


chia-tools/src/bin/analyze-chain.rs
```diff
@@ -94,6 +96,19 @@ fn main() {
             }
         };
 
+        let ftb = match block.foliage_transaction_block {
```
xdustinface, line 99: Im not a fan of abbreviations like this since it imo always just makes reading code less intuitive. Why not just name it `foliage_transaction_block`?

chia-tools/graph.gnuplot
```diff
@@ -67,3 +72,15 @@ set xrange [0:2000]
 plot "chain-resource-usage-cdf.dat" using 11:1 with lines title "block decompress + parse", \
 "chain-resource-usage-cdf.dat" using 12:1 with lines title "block reference lookup", \
 "chain-resource-usage-cdf.dat" using 14:1 with lines title "conditions"
+
+set output "block-wall-clock-delta-cdf.png"
```
xdustinface, line 76: Whats the meaning of the `cdf` postfix used in this file?

xdustinface (approved):
Looks good but one question and two nits, feel free to close them.

chia-tools/graph.gnuplot
arvidn: cumulative distribution function. The plot is essentially the integral of the frequency function of the samples

chia-tools/src/bin/analyze-chain.rs
arvidn: long variable names adds noise

PR #158, opened 2023-05-02
add new (simplified) generator ROM, enabled by the HARD_FORK flag

arvidn:
This introduces the alternate, hard-fork, generator ROM. It can be selected via the `HARD_FORK` flag.
The generator tests run all test generators using both ROMs ensuring identical output.


chia-tools/src/bin/analyze-chain.rs
```diff
@@ -8,7 +8,7 @@ use std::time::SystemTime;
 use sqlite::State;
 
 use chia::gen::conditions::parse_spends;
-use chia::gen::flags::MEMPOOL_MODE;
+use chia::gen::flags::{HARD_FORK, MEMPOOL_MODE};
```
richardkiss, line 11: This may be a bit pedantic, but you might consider calling it `HARD_FORK_1`. Unless you think this is the last hard fork.

src/gen/flags.rs
```diff
@@ -20,6 +20,10 @@ pub const ENABLE_ASSERT_BEFORE: u32 = 0x100000;
 // disallow relative height- and time conditions on ephemeral spends
 pub const NO_RELATIVE_CONDITIONS_ON_EPHEMERAL: u32 = 0x200000;
 
+// use the generator ROM that does not pass in the clvm parser as the first
+// argument
+pub const HARD_FORK: u32 = 0x400000;
```
richardkiss, line 25: Here I think the name should definitely be different. A hard fork may end up having other changes beyond the generator. So this flag should probably be `USE_V2_GENERATOR_ROM` or something like that, ie. just the hard fork features that are relevant at this scope.

chia-tools/src/bin/analyze-chain.rs
hoffmang9: As much as I know Bram wishes that we'll never have another hard fork... I remain a bit more skeptical...

richardkiss (approved):
Looks good, just a complaint about the name of the flag.

src/gen/flags.rs
arvidn: it's a good point. the name should indicate the actual change it makes, and if there are multiple changes as part of the hard fork, they should have different flags

src/generator_rom.rs
```diff
@@ -50,6 +50,28 @@ pub const GENERATOR_ROM: [u8; 737] = [
     0x80,
 ];
 
+// https://github.com/Chia-Network/chia-blockchain/blob/main/chia/wallet/puzzles/rom_bootstrap_generator2.clsp.hex
+
+pub const GENERATOR2_ROM: [u8; 265] = [
+    0xff, 0x02, 0xff, 0xff, 0x01, 0xff, 0x02, 0xff, 0x08, 0xff, 0xff, 0x04, 0xff, 0x02, 0xff, 0xff,
```
richardkiss, line 56: This construct has always bothered me and I just ran across this crate which we could use at compile-time only to tidy it up a bit:

https://docs.rs/hex-literal/latest/hex_literal/

Maybe even just import the `.hex` file and use it with `include_str!` or `include_bytes!`.

tests/test-generators.py
```diff
@@ -32,6 +32,17 @@ def parse_output(result, error_code) -> str:
     run_time = perf_counter() - start_time
     output = parse_output(result, error_code)
 
+    sys.stdout.write("running generator (hard-fork mode) ...\r")
+    start_time = perf_counter()
+    error_code3, result3 = run_gen(g, LIMIT_STACK | HARD_FORK, "generators/block-834752.hex")
+    run_time3 = perf_counter() - start_time
+    output3 = parse_output(result3, error_code3)
+    if output != output:
```
arvidn, line 40: this was a bad typo, causing the test not to catch the discrepancy in the cost of the new generator. I'll update `chia-blockchain first`

PR #159, opened 2023-05-04
use hex_literal for GENERATOR_ROM

arvidn:
this macro only takes a string literal, it cannot be combined with the include binary-macro


richardkiss (approved):

richardkiss:
I wonder if you could assign some other constant value the included value and then convert it with the `hex!` macro. I guess probably not.

arvidn:
yeah, I tried that. it seems the `hex_literal` macro only accepts string literals

PR #165, opened 2023-05-12
Opt for source-based code coverage

AmineKhaldi:


arvidn (commented):
would it be possible to build the python and wasm bindings with `-Cinstrument-coverage` and collect coverage when running those tests as well?
I think the coverage that's relevant is whatever is in the `src` directory in all crates in the workspace. it's possible that could be selected by `**/src/**`

arvidn:
I'm a bit hesitant about running all tests twice. once with coverage instrumentation and once without.

arvidn (approved):

arvidn (approved):

PR #167, opened 2023-05-24
run `cargo fmt` on fuzzers and wheel, in CI

arvidn:
This updates the CI workflow to run `cargo fmt` in `wheel` and `fuzz`. Those aren't part of the workspace, so they aren't covered by `cargo fmt --all`.
I also made it run `cargo clippy` in `wheel` but haven't figured out how to do that in `fuzz` yet, as it only builds with the `fuzzing` build configuration.


xdustinface (approved):
Seems fine. Not sure if all the reformatting, dropping the `mut` and changing the length checks is required for what the PR wants to do by its title and description though.

arvidn:
right, the title doesn't mention clippy

PR #168, opened 2023-05-24
add new SOFTFORK condition

arvidn:
Enabling this is a hard fork, and it allows future conditions to be added as soft-forks and have non-zero cost.

The new condition is `HARDFORK`. The opcode is 90. It requires at least one argument, which is the *cost* of the condition (specified in 10 thousands). All other parameters are ignored.

In memepool-mode (`NO_UNKNOWN_CONDS`), any `SOFTFORK` condition is an error, as expected.
In consensus mode (the default), any `SOFTFORK` condition is silently ignored, but the cost is applied.


src/gen/conditions.rs
```diff
@@ -3665,3 +3691,79 @@ fn test_relative_condition_on_ephemeral(
...
 }
+
+#[cfg(test)]
+#[rstest]
+// the scale factor is 10000
+#[case("((2 (1 )", 10000)]
+// 0 is OK (but does it make sense?)
+#[case("((2 (0 )", 0)]
+// the cost accumulates
+#[case("((2 (1 ) ((2 (2 ) ((2 (3 )", (1 + 2 + 3) * 10000)]
```
richardkiss, line 3702: What's the format for these conditions? It seems weird that the parens don't balance.

src/gen/opcodes.rs
```diff
@@ -50,6 +51,12 @@ pub const ASSERT_BEFORE_HEIGHT_ABSOLUTE: ConditionOpcode = 87;
 // no-op condition
 pub const REMARK: ConditionOpcode = 1;
 
+// takes its cost as the first parameter, followed by future extensions
+// the cost is specified in increments of 10000, to keep the values smaller
+// This is a hard fork and is therefore only available when enabled by the
+// ENABLE_SOFTFORK_CONDITION flag
+pub const SOFTFORK: ConditionOpcode = 2;
```
richardkiss, line 58: It's nice to have condition codes be large enough so they are in a different namespace than opcodes, which means a single dissassembler can work for programs representing clvm code and programs representing conditions without giving hints. Can we bump this up to, maybe say, 96 or 112?

Using `1` was sort of a mistake that maybe is was okay since `REM` and `q` are arguably kind of the same thing.

src/gen/conditions.rs
```diff
@@ -199,6 +203,20 @@ pub fn parse_args(
...
+        SOFTFORK => {
+            if (flags & NO_UNKNOWN_CONDS) != 0 {
+                // We don't know of any new softforked-in conditions, so they
+                // are all unknown
+                Err(ValidationErr(c, ErrorCode::InvalidConditionOpcode))
+            } else {
+                match sanitize_uint(a, first(a, c)?, 4, ErrorCode::InvalidSoftforkCost)? {
+                    // the first argument represents the cost of the condition.
+                    // We scale it by 10000 to make the argument be a bit smaller
+                    SanitizedUint::Ok(cost) => Ok(Condition::Softfork(cost * 10000)),
```
richardkiss, line 215: Do we really want to bias via a power of 10 rather than a power of 2?

Who's in charge here, the people or the computers? :)

src/gen/conditions.rs
arvidn: these tests use a simple parser where `(` means `0xff` and `)` means `0x80`, so all open-parentheses just take two items and are implicitly terminated

src/gen/opcodes.rs
arvidn: sure, I'll change this to a larger value

src/gen/conditions.rs
arvidn: well, all of our costs and cost limits are aligned with base 10 numbers.
I don't have a strong opinion on what this factor should be. I based this on the existing condition costs, which are `1800000` and `1200000`, so we might even get away with a precision of 100k.

richardkiss (approved):
Looks good.

PR #170, opened 2023-05-26
Unknown conditions with cost

arvidn:
# overview

This patch assign cost to some (currently unknown) condition codes, when the `ENABLE_SOFTFORK_CONDITION` flag is set. Setting this flag is a hard fork, and so is changing the costs of unknown conditions. This flag is intended to be set when the 2.0 hard fork activates.

The rationale for this change is to enable adding new conditions *with cost* as soft forks in the future. It is already possible to soft-fork in new conditions, but they must all have cost 0, since all unknown conditions currently have 0 cost.

# the details

Currently, all defined condition codes fit in 1 byte, so `ConditionOpcode` is an alias for `u8`. This patch changes this to be an alias for `u16`. This allows the conditions parser to handle 2-byte conditions.

When the flag `ENABLE_SOFTFORK_CONDITION` is enabled, any 2-byte condition opcode whose first byte is `0x01` is assigned a cost equal to the *second* bytes value AND 0x07 multiplied by 200000, plus 1'000'000.

# examples

| bye 0 | byte 1 | cost |
| --- | --- | --- |
| 1 | 0 | 1'000'000 |
| 1 | 1 | 1'200'000 |
| 1 | 2 | 1'400'000 |
| 1 | 3 | 1'600'000 |
| 1 | 4 | 1'800'000 |
| 1 | 5 | 2'000'000 |
| 1 | 6 | 2'200'000 |
| 1 | 7 | 2'400'000 |
| 1 | 8 | 1'000'000 |
| 1 | 2 | 1'200'000 |
| 1 | ... | ... |
| 1 | 253 | 2'000'000 |
| 1 | 254 | 2'200'000 |
| 1 | 255 | 2'400'000 |

### DIFF
--- a/src/gen/conditions.rs
+++ b/src/gen/conditions.rs
@@ -1,14 +1,14 @@
 use super::coin_id::compute_coin_id;
 use super::condition_sanitizers::{parse_amount, sanitize_announce_msg, sanitize_hash};
 use super::opcodes::{
-    parse_opcode, ConditionOpcode, AGG_SIG_COST, AGG_SIG_ME, AGG_SIG_UNSAFE,
-    ASSERT_BEFORE_HEIGHT_ABSOLUTE, ASSERT_BEFORE_HEIGHT_RELATIVE, ASSERT_BEFORE_SECONDS_ABSOLUTE,
-    ASSERT_BEFORE_SECONDS_RELATIVE, ASSERT_COIN_ANNOUNCEMENT, ASSERT_CONCURRENT_PUZZLE,
-    ASSERT_CONCURRENT_SPEND, ASSERT_EPHEMERAL, ASSERT_HEIGHT_ABSOLUTE, ASSERT_HEIGHT_RELATIVE,
-    ASSERT_MY_AMOUNT, ASSERT_MY_BIRTH_HEIGHT, ASSERT_MY_BIRTH_SECONDS, ASSERT_MY_COIN_ID,
-    ASSERT_MY_PARENT_ID, ASSERT_MY_PUZZLEHASH, ASSERT_PUZZLE_ANNOUNCEMENT, ASSERT_SECONDS_ABSOLUTE,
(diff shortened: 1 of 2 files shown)


src/gen/conditions.rs
```diff
@@ -217,6 +217,15 @@ pub fn parse_args(
                 }
             }
         }
+        256..=511 => {
```
richardkiss, line 220: Integer representations are not unique (for example, 0x0100 and 0x000100 are both 256). Does `sanitize_uint` require canonical representation? I'm just curious if we can use `0x000100` here. Either way is fine, just want to make sure there is intent.

richardkiss (approved):
Seems reasonable.

src/gen/opcodes.rs
```diff
@@ -60,53 +60,74 @@ pub const SOFTFORK: ConditionOpcode = 90;
...
-                    | ASSERT_CONCURRENT_SPEND
-                    | ASSERT_CONCURRENT_PUZZLE
-                    | ASSERT_MY_BIRTH_SECONDS
-                    | ASSERT_MY_BIRTH_HEIGHT
-                    | ASSERT_EPHEMERAL => Some(buf[0]),
-                    _ => None,
+    if buf.len() == 2 && (flags & ENABLE_SOFTFORK_CONDITION) != 0 {
+        let b0 = buf[0];
+        if b0 != 1 {
+            None
```
arvidn, line 83: @richardkiss in case the 2-byte condition code has a redundant leading zero, it's caught here and not interpreted as the 1-byte condition code. For codes > 2 bytes, they are caught on line 130, and also not interpreted as a code whose canonical form has fewer bytes

src/gen/conditions.rs
arvidn: I don't think `sanitize_uint` is involved in the parsing of the condition codes, but codes with redundant leading zeroes are not interpreted as the codes in canonical form. See my other comment

PR #171, opened 2023-06-01
add new flag, LIMIT_OBJECTS

arvidn:
to restrict the number of atoms and pairs we allow allocating. This overlaps with `LIMIT_HEAP` which also limits the total heap size and is used in mempool-mode.

The intention of this flag is to prepare for soft-forking on just the limit on the number of atoms and pairs.

### DIFF
--- a/chia-tools/graph.gnuplot
+++ b/chia-tools/graph.gnuplot
@@ -59,6 +59,19 @@ plot "chain-resource-usage-cdf.dat" using ($5)*8/1000000:1 with lines title "Ato
 "chain-resource-usage-cdf.dat" using ($6)*8/1000000:1 with lines title "Pair size", \
 "chain-resource-usage-cdf.dat" using ($7)/1000000:1 with lines title "Heap size"
 
+set output "blockchain-object-usage-cdf.png"
+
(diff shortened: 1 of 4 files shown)


chia-tools/graph.gnuplot
```diff
@@ -59,6 +59,19 @@ plot "chain-resource-usage-cdf.dat" using ($5)*8/1000000:1 with lines title "Ato
...
 "chain-resource-usage-cdf.dat" using ($7)/1000000:1 with lines title "Heap size"
 
+set output "blockchain-object-usage-cdf.png"
+
+set title "block object usage"
+set xlabel "count (million)"
+set xrange [0:*]
+
+plot "chain-resource-usage-cdf.dat" using ($5)/1000000:1 with lines title "Allocated atoms", \
+"chain-resource-usage-cdf.dat" using ($6)/1000000:1 with lines title "Allocated pairs"
```
arvidn, line 69: this is somewhat independent, I used this to double check that these limits have plenty of margin from what's currently being used by blocks

richardkiss (approved):

PR #175, opened 2023-06-11
Add a flag to restrict the number of announcements

arvidn:
When this flag is enabled, a single spend is not allowed to emit more than 1000 create-announcement, assert-announcement, concurrent spend or coin announcements.

This is a soft-fork and is enabled immediately in mempool-mode.

The objective of this is to make it simpler to avoid DoS vectors, since there's no cost associated with these conditions. The current implementation is careful to not copy messages and data associated with conditions for this reason. With this limit, the implementation would not have to be as careful.

### DIFF
--- a/src/gen/conditions.rs
+++ b/src/gen/conditions.rs
@@ -13,8 +13,8 @@ use super::opcodes::{
 use super::sanitize_int::{sanitize_uint, SanitizedUint};
 use super::validation_error::{first, next, rest, ErrorCode, ValidationErr};
 use crate::gen::flags::{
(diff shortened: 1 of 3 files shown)


src/gen/conditions.rs
```diff
@@ -1195,7 +1216,7 @@ fn test_coin_id(parent_id: &[u8; 32], puzzle_hash: &[u8; 32], amount: u64) -> By
 fn parse_list_impl(
     a: &mut Allocator,
     input: &str,
-    callback: &Option<fn(&mut Allocator) -> NodePtr>,
+    callback: &Option<Box<dyn Fn(&mut Allocator) -> NodePtr>>,
```
arvidn, line 1219: this changes a plain function pointer to type-erased function object. This allows passing in closures as well as plain functions. The new test requires capturing `count` and `cond`

PR #176, opened 2023-06-12
pre-compile generator tests, and rust port

arvidn:
This is best reviewed one commit at a time.

The generator tests are currently in CLVM form, compiled right before executing them.
It turns out that, in some cases, the compilation time dominated the run-time. By pre-compiling the generator programs, the timings more accurately represent the time contributed to block validation.

The tests moved from `tests/generators` to the root, `generator-tests`.
They first line was changed from CLVM code to serialized hex form.
The expected output was unchanged.

The test program `tests/test-generators.py` was ported to rust in `src/gen/test_generators.rs`.

The original python test had its upper limits on timings adjusted down, to account for no longer compiling the programs


.github/workflows/build-test.yml
```diff
@@ -362,6 +362,10 @@ jobs:
         os: [macos-latest, ubuntu-latest, windows-latest]
     name: Unit tests
     steps:
+      - name: disable git autocrlf
+        run: |
+          git config --global core.autocrlf false
```
arvidn, line 367: the test files are read verbatim, which means it's problematic if `git` is allowed to insert `\r` when checking out files

src/gen/test_generators.rs
arvidn, line 1: this is a port of the test from `tests/test-generators.py`

tests/run_gen.py
```diff
@@ -2,26 +2,17 @@
...
-
 def run_gen(fn: str, flags: int = 0, args: Optional[str] = None):
 
     # constants from the main chia blockchain:
     # https://github.com/Chia-Network/chia-blockchain/blob/main/chia/consensus/default_constants.py
     max_cost = 11000000000
     cost_per_byte = 12000
 
-    generator = binutils.assemble(open(fn, "r").read()).as_bin()
+    generator = bytes.fromhex(open(fn, "r").read().split('\n')[0])
```
arvidn, line 15: no need to compile the CLVM anymore

.github/workflows/build-test.yml
richardkiss: Would you consider putting this comment into the YAML file itself?

arvidn: good point

richardkiss:
Seems reasonable. I was going to suggest just separating out the "compile" from the "benchmark" phase, but then I realized you're deleting the original CLVM disassembly, and I actually don't think it's significantly more readable than the hex (and it's really just a command-line `opd` away). So I'm on board.

richardkiss (approved):

richardkiss (approved):

PR #180, opened 2023-06-19
introduced a new flag, ALLOW_BACKREFS

arvidn:
which allows encoded programs and arguments be serialized with back-references (a more compact serialization format).

This proposal could be enabled in the hard-fork with the following benefits:

* blocks could use a lot fewer bytes for their block generator
* the response from `get_puzzle_and_solution_for_coin` would no longer be at risk of a DoS attack, where the response is unreasonably large. this serialization format would collapse duplicates.
* running block generators would use less memory, since atoms would be reused.
* If we ever cache sha256 tree hashes, we would be less likely to have to re-compute the same hash multiple times, since identical sub-trees would share nodes. The same goes for if we ever want to introduce JETs, we would only have to identify a function once and replace it.


wheel/src/api.rs
```diff
@@ -89,11 +90,22 @@ pub fn get_puzzle_and_solution_for_coin<'py>(
...
     find_amount: u64,
     find_ph: Bytes32,
+    flags: u32,
 ) -> PyResult<(&'py PyBytes, &'py PyBytes)> {
     let mut allocator = make_allocator(LIMIT_HEAP);
-    let program = node_from_bytes(&mut allocator, program)?;
-    let args = node_from_bytes(&mut allocator, args)?;
-    let dialect = &ChiaDialect::new(LIMIT_STACK);
+
+    let (program, args) = if (flags & ALLOW_BACKREFS) != 0 {
```
richardkiss, line 97: A little petty but slightly shorter.

```suggestion
    let nfb = if (flags & ALLOW_BACKREFS) != 0 { node_from_bytes_backrefs } else { node_from_bytes };
    let program = nfb(&mut allocator, program)?;
    let args = nfb(&mut allocator, args)?;

```

chia-tools/Cargo.toml
```diff
@@ -26,3 +27,8 @@ bench = false
 name = "test-block-generators"
 test = false
 bench = false
+
+[[bin]]
+name = "compress-clvm"
```
richardkiss, line 32: Not necessary here since it's done, but it might be easier in future to write command-line tools like this that are not sensitive to start-up time in python.

src/gen/run_puzzle.rs
```diff
@@ -22,8 +23,14 @@ pub fn run_puzzle(
     max_cost: u64,
     flags: u32,
 ) -> Result<SpendBundleConditions, ValidationErr> {
-    let puzzle = node_from_bytes(a, puzzle)?;
-    let solution = node_from_bytes(a, solution)?;
+    let (puzzle, solution) = if (flags & ALLOW_BACKREFS) != 0 {
```
richardkiss, line 26: Maybe same trick here, assign intermediate function `nfb` to either `node_from_bytes` or `node_from_bytes_backrefs` based on flags and use that.

wheel/src/api.rs
```diff
@@ -106,10 +118,22 @@ pub fn get_puzzle_and_solution_for_coin<'py>(
 
     match r {
         Err(eval_err) => eval_err_to_pyresult(py, eval_err, allocator),
-        Ok((puzzle, solution)) => Ok((
-            PyBytes::new(py, &node_to_bytes(&Node::new(&allocator, puzzle))?),
-            PyBytes::new(py, &node_to_bytes(&Node::new(&allocator, solution))?),
-        )),
+        Ok((puzzle, solution)) => {
```
richardkiss, line 121: Same trick here.

richardkiss (approved):

AmineKhaldi (approved):

PR #181, opened 2023-06-20
update costs of unknown conditions, part of the hard fork

arvidn:
This extends the condition codes that have cost to all 2-byte conditions (except ones with leading zero).

The formula changes to use a lookup table based on the second byte.


src/gen/opcodes.rs
```diff
@@ -61,14 +61,39 @@ pub const CREATE_COIN_COST: Cost = 1800000;
...
+    208000000, 221000000, 235000000, 249000000, 265000000, 282000000, 299000000, 318000000,
+    338000000, 359000000, 382000000, 406000000, 431000000, 458000000, 487000000, 517000000,
+];
 
 pub fn compute_unknown_condition_cost(op: ConditionOpcode) -> Cost {
-    match op {
-        256..=511 => 1000000 + ((op & 0x7) as Cost) * 200000,
-        _ => 0,
+    if op < 256 {
+        println!("op: {op}");
```
richardkiss, line 93: We should probably remove this

src/gen/opcodes.rs
```diff
@@ -61,14 +61,39 @@ pub const CREATE_COIN_COST: Cost = 1800000;
 pub const AGG_SIG_COST: Cost = 1200000;
 
 // when ENABLE_SOFTFORK_CONDITION is enabled
-// 2-byte condition opcodes whose first byte is 1 have costs, despite being unknown:
+// 2-byte condition opcodes have costs according to this table:
 
-// cost = 1000000 + (byte[1] & 0x7) * 200000
+const COSTS: [Cost; 256] = [
```
richardkiss, line 66:
```suggestion
// the values `100 * (17 ** idx)/(16 ** idx)` rounded to three significant decimal figures
const COSTS: [Cost; 256] = [
```

src/gen/opcodes.rs
arvidn: oops :)

richardkiss (approved):
I could try doing a PR to calculate this at run-time, in a portable way.

PR #187, opened 2023-06-25
run_block_generator tests

arvidn:
The main change is to extend the `test_generators` tests to support passing in a block reference into the generator being tested. It also adds one such generator from mainnet,

There are also two minor simplifications of `run_block_generator()` and `tree_hash()`.


src/gen/run_block_generator.rs
arvidn, line 1: there's a test covering the edge cases of accounting cost here: https://github.com/Chia-Network/chia_rs/blob/main/tests/test_run_block_generator.py

AmineKhaldi (approved):

PR #190, opened 2023-06-27
bump pyo3 dependency

arvidn:
fix use of deprecated functions and use new frozen and get_all features.

https://docs.rs/pyo3/latest/pyo3/attr.pyclass.html

# frozen

> Declares that your pyclass is immutable. It removes the borrow checker overhead when retrieving a shared reference to the Rust struct, but disables the ability to get a mutable reference.


wheel/Cargo.toml
```diff
@@ -18,7 +18,7 @@ path = "src/lib.rs"
 chia = { path = "..", features = ["py-bindings"] }
 chia-protocol = { path = "../chia-protocol", features = ["py-bindings"]  }
 clvmr = "=0.2.6"
-pyo3 = { version = "=0.15.1", features = ["extension-module", "multiple-pymethods"] }
+pyo3 = { version = "=0.19.0", features = ["extension-module", "multiple-pymethods"] }
```
richardkiss, line 21: This one is pinned with `=` unlike the others which use `>=`. Is this intentional?

richardkiss (commented):
Do we want to use `>=` or `=`? I see both in use

Rigidity:
> Do we want to use `>=` or `=`? I see both in use

Wouldn't it be easier to prevent semver version issues with shared dependencies if we drop the `=` entirely, and just use the version number itself?

wheel/Cargo.toml
arvidn: yes, it's deliberate. the `wheel` is the root crate, and it makes sense for it to pin the versions. The dependencies don't really need to. This lets us update one place instead of 3 when we bump the version

arvidn:
> I don't think libraries are supposed to pin dependencies to specific versions like this. It should be up to the end application.

We need to at least specify a lower bound to ensure the features we use are available

chia_py_streamable_macro/src/lib.rs
```diff
@@ -55,8 +55,8 @@ pub fn py_streamable_macro(input: TokenStream) -> TokenStream {
     };
 
     let mut py_protocol = quote! {
-        #[pyproto]
-        impl pyo3::class::basic::PyObjectProtocol for #ident {
```
arvidn, line 59: this way of defining special members was deprecated and removed in the new version

Rigidity (approved):
lgtm

PR #191, opened 2023-06-28
Fix clippy warnings

Rigidity:
Fixes over a hundred `cargo clippy` warnings.


arvidn:
I think it's important to make CI enforce this as well, perhaps by bumping the version of clippy we use, or something like that. if CI doesn't enforce it, it will deteriorate quickly.

I'm also a little bit worried about this causing a lot of conflicts with patches I'm working on. Did you need any manual patching for this? If it's all automatic, we could make this change after 2.0 is released, and the major patches have landed.

PR #193, opened 2023-06-28
avoid copying buffers from python

arvidn:
by using the buffer API in a few more places


wheel/src/api.rs
```diff
@@ -83,8 +84,8 @@ pub fn tree_hash(py: Python, blob: pyo3::buffer::PyBuffer<u8>) -> PyResult<&PyBy
 #[pyfunction]
 pub fn get_puzzle_and_solution_for_coin<'py>(
     py: Python<'py>,
-    program: &[u8],
-    args: &[u8],
+    program: PyBuffer<u8>,
```
richardkiss, line 87: I'd be surprised if this is necessary. Since `program` is a `PyBytes`, I don't see what `pyo3` would convert it to; it would just get a pointer to the `PyBytes`. Do you have some plan to keep the `program` in something besides a python `bytes`?

richardkiss: I think maybe I get it now... you have a large `bytes` blob that has the entire block in it, and you use `memoryview(blob)[generator_start:]` to get a buffer with just the generator in it without copying?

arvidn: exactly. Currently we effectively do exactly that to pull out generators from block-refs ([here](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/util/full_block_utils.py)). My main consideration when making this patch was the block generators, because they have the potential to be quite large (and many).

Converting `program` and `args` to use the buffer API may not be as strongly motivated. In the future I imagine we'd want to raise the abstraction level slightly of this function though (`get_puzzle_and_solution_for_coin()`) to take a list of block references rather than just `args`. That would also let us avoid copying them into a new buffer.

PR #194, opened 2023-06-28
remove unused function, run_generator()

arvidn:


xdustinface (dismissed):
Shouldn't the below comment also be dropped? 

https://github.com/Chia-Network/chia_rs/blob/a1edc7a0546e88be78225a2825f0c51effcf955f/src/gen/conditions.rs#L506

xdustinface (approved):

PR #199, opened 2023-06-29
Rust rom generator

arvidn:
This should be reviewed one commit at a time.

The objective of this patch is to have a way to validate a block without charging cost for the ROM generator program. This is a hard fork, and therefore exposed as a separate function, run_block_generator2().

This is achieved by porting the CLVM implementation of the ROM generator to rust, which is not subject to the cost limits, or any of the stack and heap allocation limits.

* implementing `run_block_generator2()`, which is pretty straight forward. The tests ensure we get the same output as the original `run_block_generator()` and that the cost is never greater than the original.
* the tool `test-block-generators` is extended to support running all blocks on mainnet through both versions, to ensure they produce the same output.

This is how one can verify the output of `run_block_generator2()` against the original version, for all mainnet blocks:

```
cargo run --release --bin test-block-generators -- --file ~/.chia/mainnet/db/blockchain_v2_mainnet.sqlite --validate --rust-generator
```

# performance

This patch gives a decent speed-up of 2.48
its max memory use is ~40% of `main`

## main
```
        Command being timed: "cargo run --release --bin test-block-generators -- --file /home/arvid/.chia/mainnet/db/blockchain_v2_mainnet.sqlite --max-height 3000000"
        User time (seconds): 41509.19
        System time (seconds): 953.03
        Percent of CPU this job got: 3008%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 23:31.41
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 46339456
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 303119956
        Voluntary context switches: 2140282
        Involuntary context switches: 3991203
        Swaps: 0
        File system inputs: 0
        File system outputs: 8
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

## rust-generator

```
        Command being timed: "cargo run --release --bin test-block-generators -- --file /home/arvid/.chia/mainnet/db/blockchain_v2_mainnet.sqlite --rust-generator --max-height 3000000"
        User time (seconds): 16337.93
        System time (seconds): 312.56
        Percent of CPU this job got: 2932%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 9:27.72
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 18787056
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 98084186
        Voluntary context switches: 1290841
        Involuntary context switches: 1968380
        Swaps: 0
        File system inputs: 30915888
        File system outputs: 8
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0

```


src/gen/run_block_generator.rs
```diff
@@ -75,3 +78,111 @@ pub fn run_block_generator<GenBuf: AsRef<[u8]>>(
...
+    Ok(ret)
+}
+
+// This has the same behavior as run_block_generator() but implements the
+// generator ROM in rust instead of using the CLVM implementation.
+// it is not backwards compatible in the CLVM cost computation (in this version
+// you only pay cost for the generator, the puzzles and the conditions).
+// it also does not apply the stack depth or object allocation limits the same,
+// as each puzzle run in its own environment.
+pub fn run_block_generator2<GenBuf: AsRef<[u8]>>(
```
richardkiss, line 111: Maybe `run_block_generator_native` or `run_block_generator_rust`?

src/gen/run_block_generator.rs
```diff
@@ -75,3 +78,111 @@ pub fn run_block_generator<GenBuf: AsRef<[u8]>>(
...
+    // (parent-coin-id puzzle-reveal amount solution . extra)
+    // where extra may be nil, or additional extension data
+
+    let mut ret = SpendBundleConditions::default();
+    let mut state = ParseState::default();
+
+    while let Some((spend, rest)) = a.next(all_spends) {
+        all_spends = rest;
+        // process the spend
+        let [parent_id, puzzle, amount, solution, _spend_level_extra] =
```
richardkiss, line 160: I didn't know you could use `[` square brackets here

src/gen/run_block_generator.rs
```diff
@@ -75,3 +78,111 @@ pub fn run_block_generator<GenBuf: AsRef<[u8]>>(
...
+            extract_n::<5>(a, spend, ErrorCode::InvalidCondition)?;
+
+        let Reduction(clvm_cost, conditions) =
+            run_program(a, &dialect, puzzle, solution, cost_left)?;
+
+        subtract_cost(a, &mut cost_left, clvm_cost)?;
+
+        let buf = tree_hash(a, puzzle);
```
richardkiss, line 168: We're not caching tree hashes here, but I suppose you're probably aware. We could may add a cache parameter to `tree_hash` (it could be a linear `Vec` using the same trick we use for allocator indices). But I suppose that's intentionally not a part of this patch.

src/gen/run_block_generator.rs
```diff
@@ -75,3 +78,111 @@ pub fn run_block_generator<GenBuf: AsRef<[u8]>>(
...
+        let puzzle_hash = a.new_atom(&buf)?;
+
+        process_single_spend(
```
richardkiss, line 171: Oh interesting, I guess this already exists for wallet API stuff.

richardkiss (approved):
Seems reasonable.

src/gen/run_block_generator.rs
arvidn: I split this functionality up a while ago in anticipation of this

src/gen/run_block_generator.rs
arvidn: as far as I can tell, we only compute these hashes once. I can't envision a way to accelerate these tree hashes. (unless you're talking about a future where the hard fork has activated, but that's for later)

src/gen/run_block_generator.rs
arvidn: in my mind, the main property of this function is that it doesn't charge cost for the ROM. Everything else is an implementation detail. I thought `2` was a reasonable name because it's version 2 and the hard fork is even introduced in `2.0`

src/gen/run_block_generator.rs
arvidn: yeah, you can destructure arrays of fixed size

PR #201, opened 2023-07-01
make message types constructible from rust

arvidn:
and add a coin ID unit test which requires `Coin` to be constructible.


Rigidity (approved):
lgtm

PR #202, opened 2023-07-01
solution_generator

arvidn:
This adds a function that creates a (simple) generator given a list of spends. It's a port of `simple_solution_generator` from `chia-blockchain`.

The purpose is to expose the ability to serialize the resulting generator with backrefs. This serialization will be supported when the hard fork activates. Having this functionality now lets us test it in chia-blockchain.


richardkiss:
Seems reasonable

richardkiss (approved):

PR #203, opened 2023-07-02
extend the test-block-generators tool

arvidn:
to support re-serializing generators with back-references. This allows validating that the result is identical after re-serializing the generator, like this:
```
cargo run --release --bin test-block-generators -- --file ~/.chia/mainnet/db/blockchain_v2_mainnet.sqlite --test-backrefs --validate
```


richardkiss (approved):
Seems good

PR #205, opened 2023-07-07
include fuzz in workspace

arvidn:
there are two justifications for this:
1. by including the fuzz projects in the main workspace, they share the `Cargo.lock` file with all other build targets, and we ensure that we fuzz the exact same dependencies that we build the production wheel against
2. `cargo fmt --all` and `cargo clippy --workspace` will cover the fuzz targets as well. This lets us simplify the workflow file a bit and also adds new coverage for clippy that we previously did not have

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 # the "wheel" crate is excluded from the workspace because pyo3 has problems with
 # "cargo test" and "cargo bench"
 [workspace]
-members = ["wasm", "chia_streamable_macro", "chia-bls", "clvm-utils", "chia-protocol", "chia_py_streamable_macro", "chia-tools"]
+members = ["wasm", "chia_streamable_macro", "chia-bls", "clvm-utils", "chia-protocol", "chia_py_streamable_macro", "chia-tools", "fuzz", "clvm-utils/fuzz"]
 exclude = ["wheel"]
 
 [package]
--- a/chia-tools/src/bin/test-block-generators.rs
+++ b/chia-tools/src/bin/test-block-generators.rs
@@ -142,7 +142,7 @@ fn main() {
 
     let pool = ThreadPool::new(
         args.num_jobs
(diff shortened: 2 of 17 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 # the "wheel" crate is excluded from the workspace because pyo3 has problems with
 # "cargo test" and "cargo bench"
 [workspace]
-members = ["wasm", "chia_streamable_macro", "chia-bls", "clvm-utils", "chia-protocol", "chia_py_streamable_macro", "chia-tools"]
+members = ["wasm", "chia_streamable_macro", "chia-bls", "clvm-utils", "chia-protocol", "chia_py_streamable_macro", "chia-tools", "fuzz", "clvm-utils/fuzz"]
```
arvidn, line 4: clippy and fmt used to not cover `clvm-utils/fuzz` but do now

chia-tools/src/bin/test-block-generators.rs
```diff
@@ -142,7 +142,7 @@ fn main() {
 
     let pool = ThreadPool::new(
         args.num_jobs
-            .unwrap_or(available_parallelism().unwrap().into()),
+            .unwrap_or_else(|| available_parallelism().unwrap().into()),
```
arvidn, line 145: this was previously not reached by `clippy`

fuzz/src/fuzzing_utils.rs
```diff
@@ -27,7 +27,7 @@ impl<'a> BitCursor<'a> {
             Some((self.data[0] & mask(self.bit_offset)) >> (8 - num - self.bit_offset))
         } else if self.data.len() < 2 {
             num = 8 - self.bit_offset;
-            Some(&self.data[0] & mask(self.bit_offset))
+            Some(self.data[0] & mask(self.bit_offset))
```
arvidn, line 30: clippy did not reach this previously

Rigidity (approved):
lgtm

PR #212, opened 2023-07-11
Unify workspace

Rigidity:


chia-bls/fuzz/fuzz_targets/blspy-fidelity.rs
```diff
@@ -50,39 +55,56 @@ fuzz_target!(|data: &[u8]| {
...
+            .unwrap();
         assert_eq!(to_bytes(py_sk1), rust_sk1.to_bytes());
 
         let rust_pk1 = rust_pk.derive_unhardened(idx);
-        let py_pk1 = aug.call_method1("derive_child_pk_unhardened",
-            PyTuple::new(py, &[py_pk, idx.to_object(py).as_ref(py)])).unwrap();
+        let py_pk1 = aug
+            .call_method1(
+                "derive_child_pk_unhardened",
+                PyTuple::new(py, &[py_pk, idx.to_object(py).as_ref(py)]),
```
arvidn, line 70: are these `fmt` changes? it looks like perhaps this is to address clippy, but clippy still seems to be flagging this

Rigidity: I'm going to be making a Clippy PR soon, this is to satisfy fmt since these weren't previously included in the workspace (and thus weren't checked by CI)

arvidn (dismissed):
it looks like we should have a follow-up to address clippy as well

Rigidity:
I'll revisit this later

PR #213, opened 2023-07-12
new agg-sig conditions

arvidn:
This PR adds support for new `AGG_SIG`-conditions in the same family as `AGG_SIG_ME`.
Since these conditions affect the aggregated signature field on the full block, it is a hard-fork, and is only enabled when the `ENABLE_SOFTFORK_CONDITION`-flag is set.

The new conditions are:

| name | code |
| --- | --- |
| `AGG_SIG_PARENT` | 43 |
| `AGG_SIG_PUZZLE` | 44 |
| `AGG_SIG_AMOUNT` | 45 |
| `AGG_SIG_PUZZLE_AMOUNT` | 46 |
| `AGG_SIG_PARENT_AMOUNT` | 47 |
| `AGG_SIG_PARENT_PUZZLE` | 48 |

They all take the same parameters as `AGG_SIG_ME` and `AGG_SIG_UNSAFE`, i.e. `public-key` and `message`.

The cost are the same as `AGG_SIG_ME` and `AGG_SIG_UNSAFE`, i.e. 1200000.

These act the same as `AGG_SIG_ME`, but instead of including the genesis hash and the coin-id in the signed message, they include a different domain separator string and *just* the combination of parent, puzzle and amount.

### DIFF
--- a/src/gen/conditions.rs
+++ b/src/gen/conditions.rs
@@ -1,14 +1,16 @@
 use super::coin_id::compute_coin_id;
 use super::condition_sanitizers::{parse_amount, sanitize_announce_msg, sanitize_hash};
 use super::opcodes::{
-    compute_unknown_condition_cost, parse_opcode, ConditionOpcode, AGG_SIG_COST, AGG_SIG_ME,
-    AGG_SIG_UNSAFE, ASSERT_BEFORE_HEIGHT_ABSOLUTE, ASSERT_BEFORE_HEIGHT_RELATIVE,
-    ASSERT_BEFORE_SECONDS_ABSOLUTE, ASSERT_BEFORE_SECONDS_RELATIVE, ASSERT_COIN_ANNOUNCEMENT,
-    ASSERT_CONCURRENT_PUZZLE, ASSERT_CONCURRENT_SPEND, ASSERT_EPHEMERAL, ASSERT_HEIGHT_ABSOLUTE,
-    ASSERT_HEIGHT_RELATIVE, ASSERT_MY_AMOUNT, ASSERT_MY_BIRTH_HEIGHT, ASSERT_MY_BIRTH_SECONDS,
-    ASSERT_MY_COIN_ID, ASSERT_MY_PARENT_ID, ASSERT_MY_PUZZLEHASH, ASSERT_PUZZLE_ANNOUNCEMENT,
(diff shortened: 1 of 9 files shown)


src/gen/opcodes.rs
```diff
@@ -6,6 +6,8 @@ use clvmr::cost::Cost;
 pub type ConditionOpcode = u16;
 
 // AGG_SIG is ascii "1"
+pub const AGG_SIG_PUZZLE_HASH_AMOUNT: ConditionOpcode = 47;
```
richardkiss, line 9: This seems like a reasonable compromise. In ASCII, 47 & 48 are `/0`

src/gen/conditions.rs
```diff
@@ -2778,6 +2943,50 @@ fn test_agg_sig_unsafe_extra_arg() {
...
 
+#[cfg(test)]
+#[rstest]
+#[case(AGG_SIG_PARENT)]
+#[case(AGG_SIG_PUZZLE)]
+#[case(AGG_SIG_AMOUNT)]
+#[case(AGG_SIG_PUZZLE_AMOUNT)]
+#[case(AGG_SIG_PARENT_PUZZLE)]
+#[case(AGG_SIG_PARENT_AMOUNT)]
+fn test_agg_sig_ph_extra_arg(#[case] condition: ConditionOpcode) {
```
AmineKhaldi, line 2954: We're missing a case for `AGG_SIG_ME` here. Also `ph` in the name would become less informative now that we're covering several cases.

arvidn: I addressed this by removing the `AGG_SIG_ARGS` flag, and renamed the test function

AmineKhaldi (approved):
Looks good.

src/gen/opcodes.rs
```diff
@@ -231,8 +248,14 @@ fn test_parse_opcode(
 #[case(&[AGG_SIG_UNSAFE as u8], Some(AGG_SIG_UNSAFE), Some(AGG_SIG_UNSAFE))]
 #[case(&[AGG_SIG_ME as u8], Some(AGG_SIG_ME), Some(AGG_SIG_ME))]
 #[case(&[CREATE_COIN as u8], Some(CREATE_COIN), Some(CREATE_COIN))]
-// the SOFTOFORK condition is only recognized when the flag is set
+// the SOFTOFORK and new AGG_SIG_* condition is only recognized when the flag is set
```
richardkiss, line 251:
```rust
// the SOFTFORK and new AGG_SIG_* condition is only recognized when the flag is set
``` 
`s/SOFTOFORK/SOFTFORK/`

PR #215, opened 2023-07-18
Allow null conditions

richardkiss:


Rigidity:
This isn't something we plan on doing, right?

richardkiss:
Yeah, I don't remember why I opened this.

PR #216, opened 2023-07-18
Fix ROM deserializer bug.

richardkiss:


Rigidity:
Is this still an issue or can this be closed?

arvidn:
This change is a hard fork and we don't (yet) use the deserializer in our generators. So I think it's fair to say that at best this may want to be considered for a future hard fork. But I don't think there's a champion for it.

richardkiss:
It's more just an embarrassment of legacy, but I doubt this code would ever be used even if it were fixed. I'd actually prefer a hard fork that removes it.

richardkiss:
Closing — ancient draft, no longer relevant.

PR #220, opened 2023-07-29
move traits into new crate

arvidn:
move Streamable, ToJsonDict and FromJsonDict traits into a separate crate. They depends on Error, so move that as well.

This is required for allowing "chia-protocol" compatible types to be defined outside of the `chia-protocol` crate, but still used by message types defined in `chia-protocol`.

Specifically, `G1Element` and `G2Element` want to implement the `Streamable` protocol, as well as the `ToJsonDict` and `FromJsonDict`.


chia-protocol/src/bytes.rs
```diff
@@ -48,6 +54,33 @@ impl Streamable for Bytes {
     }
 }
 
+#[cfg(feature = "py-bindings")]
+impl ToJsonDict for Bytes {
```
arvidn, line 58: these were moved verbatim. They used to be in the same translation unit as the trait definition

Rigidity (approved):

PR #224, opened 2023-08-02
factor out PyJsonDict macro from PyStreamable

arvidn:
This lets types use a custom ToJsonDict and FromJsonDict implementation, but still use the PyStreamable macro.

The PyStreamable macro also used to support 1-tuples and treat them as the underlying type. This special case was also removed in this patch, as it was just there to simplify place-holder types.

### DIFF
--- a/chia_py_streamable_macro/src/lib.rs
+++ b/chia_py_streamable_macro/src/lib.rs
@@ -23,19 +23,6 @@ pub fn py_streamable_macro(input: proc_macro::TokenStream) -> proc_macro::TokenS
         syn::Data::Struct(s) => s.fields,
         syn::Data::Enum(_) => {
             return quote! {
-                impl #crate_name::to_json_dict::ToJsonDict for #ident {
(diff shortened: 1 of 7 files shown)


chia_py_streamable_macro/src/lib.rs
```diff
@@ -109,89 +96,29 @@ pub fn py_streamable_macro(input: proc_macro::TokenStream) -> proc_macro::TokenS
...
-                impl #crate_name::from_json_dict::FromJsonDict for #ident {
-                    fn from_json_dict(o: &pyo3::PyAny) -> pyo3::PyResult<Self> {
-                        Ok(Self{
-                            #(#fnames: <#ftypes as #crate_name::from_json_dict::FromJsonDict>::from_json_dict(o.get_item(stringify!(#fnames))?)?,)*
-                        })
-                    }
                 }
             });
         }
-        syn::Fields::Unnamed(FieldsUnnamed { unnamed, .. }) => {
```
arvidn, line 141: this is the case of 1-tuples being treated as the inner type.

Rigidity (approved):

PR #227, opened 2023-08-08
disable backref encoding for get_puzzle_and_solution_for_coin

arvidn:
for backwards compatibility.

The target branch `release-0.2.10` is cut from the `0.2.9` release tag.


arvidn:
the benchmark runners are down. I think we'll have to disregard those tests

AmineKhaldi (approved):

PR #228, opened 2023-08-10
Add clvm-traits and clvm-derive crates

Rigidity:
This PR adds two crates, one that provides `ToClvm` and `FromClvm` macros and implementations for various common types, and another one that is re-exported in the other that adds derive macros for these traits.

### DIFF
--- a/clvm-derive/src/from_clvm.rs
+++ b/clvm-derive/src/from_clvm.rs
@@ -1,0 +1,51 @@
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::{parse_quote, Data, DataStruct, DeriveInput, Fields};
+
+use crate::helpers::{add_trait_bounds, crate_ident, parse_args, Repr};
+
+pub fn from_clvm(mut ast: DeriveInput) -> TokenStream {
+    let args = parse_args(&ast.attrs);
+    let crate_name = crate_ident();
+
+    let fields = match &ast.data {
+        Data::Struct(DataStruct {
+            fields: Fields::Named(fields),
+            ..
+        }) => &fields.named,
+        _ => panic!("expected a struct with named fields"),
+    };
+
+    let struct_name = &ast.ident;
+    let field_type = fields.iter().map(|field| &field.ty);
+    let field_names = fields.iter().map(|field| &field.ident);
+    let destructure_names = field_names.clone();
+
+    let (match_macro, destructure_macro) = match args.repr {
+        Repr::ProperList => (
+            quote!( #crate_name::match_list ),
+            quote!( #crate_name::destructure_list ),
+        ),
+        Repr::Tuple => (
+            quote!( #crate_name::match_tuple ),
+            quote!( #crate_name::destructure_tuple ),
+        ),
+        Repr::CurriedArgs => (
+            quote!( #crate_name::match_curried_args ),
+            quote!( #crate_name::destructure_curried_args ),
+        ),
+    };
+
+    add_trait_bounds(&mut ast.generics, parse_quote!(#crate_name::FromClvm));
+    let (impl_generics, ty_generics, where_clause) = ast.generics.split_for_impl();
+
+    quote! {
+        #[automatically_derived]
+        impl #impl_generics #crate_name::FromClvm for #struct_name #ty_generics #where_clause {
+            fn from_clvm(a: &clvmr::Allocator, node: clvmr::allocator::NodePtr) -> #crate_name::Result<Self> {
+                let #destructure_macro!( #( #destructure_names, )* ) = <#match_macro!( #( #field_type ),* ) as #crate_name::FromClvm>::from_clvm(a, node)?;
+                Ok(Self { #( #field_names, )* })
+            }
+        }
+    }
+}
--- a/clvm-derive/src/from_clvm.rs
+++ b/clvm-derive/src/from_clvm.rs
@@ -2,11 +2,11 @@ use proc_macro2::TokenStream;
 use quote::quote;
 use syn::{parse_quote, Data, DataStruct, DeriveInput, Fields};
 
-use crate::helpers::{add_trait_bounds, crate_ident, parse_args, Repr};
+use crate::helpers::{add_trait_bounds, parse_args, Repr};
 
 pub fn from_clvm(mut ast: DeriveInput) -> TokenStream {
     let args = parse_args(&ast.attrs);
-    let crate_name = crate_ident();
+    let crate_name = quote!(clvm_traits);
 
     let fields = match &ast.data {
         Data::Struct(DataStruct {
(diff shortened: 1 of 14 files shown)


clvm-derive/src/from_clvm.rs
```diff
@@ -0,0 +1,51 @@
...
+pub fn from_clvm(mut ast: DeriveInput) -> TokenStream {
+    let args = parse_args(&ast.attrs);
+    let crate_name = crate_ident();
+
+    let fields = match &ast.data {
+        Data::Struct(DataStruct {
+            fields: Fields::Named(fields),
+            ..
+        }) => &fields.named,
+        _ => panic!("expected a struct with named fields"),
```
arvidn, line 16: there's no reason this wouldn't work for tuples though, right?

clvm-traits/src/match_byte.rs
```diff
@@ -0,0 +1,45 @@
...
+        a.new_number(BYTE.into()).map_err(Error::Allocator)
+    }
+}
+
+impl<const BYTE: u8> FromClvm for MatchByte<BYTE> {
+    fn from_clvm(a: &Allocator, node: NodePtr) -> Result<Self> {
+        if let SExp::Atom() = a.sexp(node) {
+            let value: u8 = to_unsigned(a.number(node)).try_into().map_err(
+                |error: <u8 as TryFrom<BigInt>>::Error| Error::Custom(error.to_string()),
+            )?;
```
arvidn, line 23: this construct makes me a bit uneasy, especially with the `to_unsigned` function that doesn't seem to do what its name suggests.

Unless I'm misunderstanding what you're doing, I think it would be much more reliable to ask for the `atom()` rather than the `number()` then ensure it's 1 byte long and then pull out that one byte.

However, you would need to take the special case of 0 into account, which is an empty atom.

clvm-traits/src/match_byte.rs
```diff
@@ -0,0 +1,45 @@
...
+use num_bigint::{BigInt, Sign};
+
+use crate::{Error, FromClvm, Result, ToClvm};
+
+#[derive(Debug, Copy, Clone)]
+pub struct MatchByte<const BYTE: u8>;
+
+impl<const BYTE: u8> ToClvm for MatchByte<BYTE> {
+    fn to_clvm(&self, a: &mut Allocator) -> Result<NodePtr> {
+        a.new_number(BYTE.into()).map_err(Error::Allocator)
```
arvidn, line 14: the allocator has special cases for 0 and 1 (`Allocator::null()` and `Allocator::one()`) which would be more efficient to use for those cases.

clvm-traits/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+
+#[cfg(test)]
+#[cfg(feature = "derive")]
+mod tests {
+    use clvmr::Allocator;
+
+    use super::*;
+
+    #[derive(Debug, ToClvm, FromClvm, PartialEq, Eq)]
+    #[clvm(tuple)]
```
arvidn, line 24: I would like these kinds of encodings documented somewhere. Maybe here, along with the tests, is a good place.
i.e. `tuple`, `proper_list` and `curried_args`

arvidn (commented):
I like this approach. I haven't reviewed the macros yet, but I think this warrants some documentation to make it clear how they are supposed to be used.
Perhaps it would make sense to put a `README` in the crate root.

Both how to use the macros, and what the options mean. but also how to implement the traits for new primitive types

Rigidity:
Sounds good, I'll look into it

clvm-derive/src/from_clvm.rs
Rigidity: We can support tuple structs, numeric enums, data enums, and maybe option-like enums. The complexity scales the more we add support for, so I haven't implemented those yet. I want to make sure we like this before adding more functionality if possible

clvm-traits/src/match_byte.rs
Rigidity: Improved the code here

clvm-traits/src/lib.rs
Rigidity: I'll add it to the crate level docs

clvm-traits/src/match_byte.rs
Rigidity: Added the optimization (it should decide the branch at compile time since the value is a const generic)

arvidn (commented):
This looks really good. I especially like how you've split up the macros into smaller pieces.

I just have a few minor comments and questions

clvm-traits/src/macros.rs
```diff
@@ -0,0 +1,122 @@
...
+                ($crate::match_curried_args!( $( $rest ),* ), ()),
+            ),
+        )
+    };
+}
+
+#[macro_export]
+macro_rules! destructure_list {
+    () => {
+        $crate::MatchByte::<0>
```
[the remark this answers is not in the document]
Rigidity: You can destructure a struct into its parts, which in this are are none. I can probably simplify this since you're right in that you don't need the data. Just have to make sure you can bind to _ multiple times

clvm-traits/src/from_clvm.rs
```diff
@@ -0,0 +1,242 @@
...
+    fn test_nil() {
+        let a = &mut Allocator::new();
+        assert_eq!(decode(a, "80"), Ok(()));
+    }
+
+    #[test]
+    fn test_array() {
+        let a = &mut Allocator::new();
+        assert_eq!(decode(a, "ff01ff02ff03ff0480"), Ok([1, 2, 3, 4]));
+        assert_eq!(decode(a, "80"), Ok([0; 0]));
```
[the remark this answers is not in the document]
Rigidity: Correct, the type argument is required and this is the most concise way to represent it

clvm-traits/src/match_byte.rs
```diff
@@ -11,35 +10,27 @@ pub struct MatchByte<const BYTE: u8>;
...
-                |error: <u8 as TryFrom<BigInt>>::Error| Error::Custom(error.to_string()),
-            )?;
-
-            if value == BYTE {
-                Ok(Self)
-            } else {
-                Err(Error::Custom(format!(
+            match a.atom(node) {
+                [] if BYTE == 0 => Ok(Self),
+                [byte] if *byte == BYTE && BYTE > 0 => Ok(Self),
```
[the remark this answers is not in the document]
Rigidity: Not currently, this is mainly an implementation detail for the macros so they can parse quote, apply, and cons operators without having to check the value separately.

clvm-traits/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+        a: u64,
+        b: i32,
+    }
+
+    #[test]
+    fn test_round_trip_tuple() {
+        let mut a = Allocator::new();
+        let value = TupleStruct { a: 52, b: -32 };
+        let node = value.to_clvm(&mut a).unwrap();
+        let round_trip = TupleStruct::from_clvm(&a, node).unwrap();
```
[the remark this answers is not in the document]
Rigidity: Added

arvidn (approved):

PR #243, opened 2023-08-16
Update clvm utils

Rigidity:


clvm-traits/src/curried_args.rs
```diff
@@ -0,0 +1,86 @@
...
+    Allocator,
+};
+
+use crate::{
+    clvm_list, clvm_quote, destructure_list, destructure_quote, match_list, match_quote, Error,
+    FromClvm, MatchByte, Result, ToClvm,
+};
+
+#[derive(Debug, Clone)]
+pub struct CurriedArgs(pub Vec<NodePtr>);
```
arvidn, line 12: don't you need a way to get the inner puzzle as well?
in the future, it would be nice to have a way to get the curried args into an array as well, to avoid the heap allocation, when you know what the puzzle is supposed to be

clvm-traits/src/curried_args.rs
```diff
@@ -0,0 +1,86 @@
...
+            "b".to_clvm(a).unwrap(),
+            "c".to_clvm(a).unwrap(),
+        ]);
+
+        check(
+            a,
+            args,
+            "ff04ffff0161ffff04ffff0162ffff04ffff0163ff01808080",
+        );
+    }
```
arvidn, line 85: I would like to see some edge cases as well. like, 0 curried args and failure to match

arvidn: I think it would be good to have a test covering this failure case as well:

https://coveralls.io/builds/62027638/source?filename=clvm-traits%2Fsrc%2Fcurried_args.rs#L23

clvm-traits/src/curried_args.rs
```diff
@@ -0,0 +1,105 @@
...
+use crate::{
+    clvm_list, clvm_quote, destructure_list, destructure_quote, match_list, match_quote, Error,
+    FromClvm, MatchByte, Result, ToClvm,
+};
+
+#[derive(Debug, Clone)]
+pub struct CurriedArgs(pub Vec<NodePtr>);
+
+impl FromClvm for CurriedArgs {
+    fn from_clvm(a: &Allocator, mut ptr: NodePtr) -> Result<Self> {
```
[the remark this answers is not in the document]
Rigidity: Can't be derived since Vec is encoded differently and is nil terminated rather than 1 terminated

clvm-utils/src/curry.rs
```diff
@@ -0,0 +1,67 @@
+use clvm_traits::{
+    clvm_list, clvm_quote, destructure_list, destructure_quote, match_list, match_quote, FromClvm,
+    MatchByte, Result, ToClvm,
+};
+use clvmr::{allocator::NodePtr, Allocator};
+
+#[derive(Debug, Clone)]
+pub struct Curry<T>(NodePtr, T);
```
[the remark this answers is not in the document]
Rigidity: Changed to `CurriedProgram`

clvm-utils/src/curried_program.rs
```diff
@@ -0,0 +1,67 @@
+use clvm_traits::{
+    clvm_list, clvm_quote, destructure_list, destructure_quote, match_list, match_quote, FromClvm,
+    MatchByte, Result, ToClvm,
+};
+use clvmr::{allocator::NodePtr, Allocator};
+
+#[derive(Debug, Clone)]
+pub struct CurriedProgram<T>(NodePtr, T);
```
arvidn, line 8: the members of this tuple needs to be public, don't they?
When used from another crate, I get an error because these are private fields

Rigidity: Yep

clvm-utils/src/curried_program.rs
```diff
@@ -5,7 +5,7 @@ use clvm_traits::{
 use clvmr::{allocator::NodePtr, Allocator};
 
 #[derive(Debug, Clone)]
-pub struct CurriedProgram<T>(NodePtr, T);
+pub struct CurriedProgram<T>(pub NodePtr, pub T);
```
arvidn, line 8: in fact, I think these fields would benefit from being named. e.g. `puzzle` and `args`. or maybe `program` and `args`

clvm-utils/src/lib.rs
```diff
@@ -1,5 +1,5 @@
+mod curry;
 mod tree_hash;
-mod uncurry;
```
arvidn: since you removed the readme, maybe you could document how to use the new uncurry mechanism in rustdoc here

clvm-traits/src/curried_args.rs
arvidn: since this function will take untrusted input, I think it's especially important to be very confident in it not breaking. We add tests to allow future changes, refactoring and optimizations, while being confident that we're not changing behavior

arvidn (approved):

PR #249, opened 2023-08-18
Initial WebSocket client implementation

Rigidity:
This implements a simple WebSocket peer client intended for connecting to trusted full nodes only.


chia-client/src/peer.rs
```diff
@@ -0,0 +1,207 @@
...
+pub struct Peer {
+    sink: Mutex<SplitSink<WebSocket, tungstenite::Message>>,
+    inbound_task: JoinHandle<()>,
+    event_receiver: broadcast::Receiver<PeerEvent>,
+    requests: Requests,
+    nonce: Mutex<u16>,
+}
+
+impl Peer {
+    pub fn new(ws: WebSocket) -> Self {
```
arvidn, line 35: if `WebSocket` would be a template parameter, we can test this really well, with some helping infrastructure

chia-client/src/peer.rs
```diff
@@ -0,0 +1,207 @@
...
+        if let Err(error) = self.sink.lock().await.send(bytes.into()).await {
+            // Prevent memory leak.
+            self.requests.lock().await.remove(&message_id);
+
+            return Err(error.into());
+        }
+
+        // Increment the nonce and drop its lock.
+        *nonce += 1;
+        drop(nonce);
```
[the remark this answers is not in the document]
Rigidity: I wanted to wait until the request is successfully sent to increment the nonce, so there aren't gaps. Maybe this doesn't matter much and will pointlessly slow things down.

chia-client/src/peer.rs
```diff
@@ -0,0 +1,207 @@
...
+        let mut nonce = self.nonce.lock().await;
+        let message_id = *nonce;
+
+        // Create the message.
+        let message = Message {
+            msg_type: T::msg_type(),
+            id: Some(message_id),
+            data: {
+                let mut body_bytes = Vec::new();
+                body.stream(&mut body_bytes)?;
```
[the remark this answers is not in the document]
Rigidity: Perhaps a helper method on `Streamable` named `stream_to_bytes` for example?

chia-client/src/peer.rs
```diff
@@ -0,0 +1,207 @@
...
+        // Convert message to bytes.
+        let mut bytes = Vec::new();
+        message.stream(&mut bytes)?;
+
+        // Create a saved oneshot channel to receive the response.
+        let (sender, receiver) = oneshot::channel::<Message>();
+        self.requests.lock().await.insert(message_id, sender);
+
+        // Send the message.
+        if let Err(error) = self.sink.lock().await.send(bytes.into()).await {
```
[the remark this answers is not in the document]
Rigidity: I did this just in case the response is somehow processed before the next line of the function after the request has been sent. I don't want to somehow miss the response and have it hang indefinitely since the oneshot channel wasn't in the hashmap at the point it was received. If you think this will never happen, the code can be a lot simpler.

chia-client/src/peer.rs
Rigidity: I'm not sure how to do use a generic Sink + Stream without losing type information that's necessary to receive messages.

Rigidity: Perhaps we can revisit this in the future

chia-client/src/peer.rs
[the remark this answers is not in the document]
arvidn: yeah, on the python side we just have `__bytes__()` for this

chia-client/src/peer.rs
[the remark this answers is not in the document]
arvidn: I'm not entirely sure how this code is supposed to be put together and work with coroutines. the way we do it in python is that every request is a coroutine (or rather, if you want to pipeline requests, you have to use multiple coroutines). There we have a main coroutine per peer that triages incoming messages to the coroutine that's waiting for that message. this is where the transaction IDs come in, to know where to send it.

but in that model, there's a kind of driver that does the actual sending and receiving of messages.

chia-client/src/peer.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        // Increment the nonce and drop its lock.
+        *nonce += 1;
+        drop(nonce);
+
+        // Wait for the response.
+        let response = receiver.await;
+
+        // Remove the one shot channel.
+        self.requests.lock().await.remove(&message_id);
```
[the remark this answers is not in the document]
arvidn: my point is that you remove `message_id` here unconditionally. regardless of the ID of the message you received. presumably, you may end up removing a channel for a message you haven't received yet

chia-client/src/peer.rs
```diff
@@ -32,6 +32,40 @@ pub struct Peer {
...
+        )
+    }
+}
+
+impl Drop for ResponseHandler {
+    fn drop(&mut self) {
+        let requests = Arc::clone(&self.requests);
+        let message_id = self.message_id;
+
+        tokio::spawn(async move {
```
arvidn, line 63: I'm not super familiar with `tokio`, but spawning a coroutine without storing a handle to it looks like a mistake. there won't be a way to synchronize/wait for this coroutine to exit.

chia-client/src/peer.rs
```diff
@@ -0,0 +1,219 @@
...
+                let found_type = message.msg_type;
+
+                if found_type != expected_type {
+                    return Err(Error::InvalidResponse(Some(message)));
+                }
+
+                match R::parse(&mut Cursor::new(message.data.as_ref())) {
+                    Ok(response) => Ok(response),
+                    _ => Err(Error::InvalidResponse(Some(message))),
+                }
```
arvidn, line 200: this looks like it could just be `.map_err()`

chia-client/src/peer.rs
```diff
@@ -0,0 +1,219 @@
...
+                                    event_sender.send(PeerEvent::CoinStateUpdate(body)).ok();
+                                }
+                            }
+                            ProtocolMessageTypes::NewPeakWallet => {
+                                let cursor = &mut Cursor::new(message.data.as_ref());
+                                if let Ok(body) = NewPeakWallet::parse(cursor) {
+                                    event_sender.send(PeerEvent::NewPeakWallet(body)).ok();
+                                }
+                            }
+                            _ => {}
```
arvidn, line 107: this might warrant some logging, but I recognize this is a proof of concept for now

chia-client/src/peer.rs
```diff
@@ -0,0 +1,219 @@
...
+
+        let inbound_task = tokio::spawn(async move {
+            while let Some(message) = stream.next().await {
+                if let Ok(message) = message {
+                    let bytes = message.into_data();
+                    let cursor = &mut Cursor::new(bytes.as_slice());
+
+                    // Parse the message.
+                    let Ok(message) = Message::parse(cursor) else {
+                        continue;
```
arvidn, line 85: this case might also warrant some logging, and maybe even closing the peer connection

arvidn (approved):
I see how this works now. it would be good to get some example built, and maybe even run as part of CI

arvidn (approved):

PR #250, opened 2023-08-18
port `blspy` to chia_rs

arvidn:
# introduction

This is a large change that encompasses a few related changes, all with the goal of making `chia_rs` a drop-in replacement for `blspy` which allows BLS keys and signatures (`G1Element` and `G2Element`) also be used in rust types with python bindings. The long term goal is to be able to perform more of block validation on the rust side, reducing serialization and deserialization costs from python.

# review

The changes are separated into (mostly) independent commits. Reviewing them one commit at a time is highly recommended.

The steps involved are, in order of commits:

1. make the BLS types implement `ToJsonDict` and `FromJsonDict`, this is part of the Streamable protocol which is required for `G1Element` and `G2Element`. Technically, the python streamable code has special cases for these types right now. This will allow us to remove those special cases and make the protocol clean, once switching over to these types.
2. make the BLS types implement `ToClvm` and `FromClvm`, this is to match the existing `G1Element` and `G2element` placeholders we currently have in `chia-protocol`.
3. make the BLS types export the streamable protocol to python, by applyinh the `PyStreamable` derive macro
4. switch the `chia-protocol` crate to use the BLS types from `chia-bls`. This also introduces type aliases for `PublicKey` and `Signature` to match the current naming of `G1Element` and `G2Element` respectively.
5. extend the python bindings for the BLS types (SecretKey, PublicKey, Signature) to match that of `blspy`. We also need to introduce `GTElement`. This also adds the `AugSchemeMPL` python type which exposes various free functions for signing and verifying.
6. update the script generating the type stubs for python, `chia_rs.pyi`. These are generated by `wheel/generate_type_stubs.py`
7. add test that ensures the types exported by `chia_rs` behave identically to `blspy`. This adds a CI dependency on `blspy` to be able to import it to the test

# benchmarks

Both `chia-bls` and `blspy` build on top of BLST, so no performance differences are expected. To validate that I ran the `benchmark.py` from the `bls-signatures` repo (where `blspy` comes from).

## blspy

```
Public key validation
Total: 100000 runs in 5211.8 ms
Avg: 0.052118

Signature validation
Total: 100000 runs in 6989.0 ms
Avg: 0.069890

Aggregation
Total: 100000 runs in 159.7 ms
Avg: 0.001597

Batch verification
Total: 100000 runs in 30790.5 ms
Avg: 0.307905
```

## chia_rs

```
Public key validation
Total: 100000 runs in 5229.4 ms
Avg: 0.052294

Signature validation
Total: 100000 runs in 6917.3 ms
Avg: 0.069173

Aggregation
Total: 100000 runs in 156.7 ms
Avg: 0.001567

Batch verification
Total: 100000 runs in 34694.5 ms
Avg: 0.346945
```


chia-bls/Cargo.toml
```diff
@@ -8,15 +8,22 @@ authors = ["Arvid Norberg <arvid@chia.net>"]
 homepage = "https://github.com/Chia-Network/chia_rs/chia-bls/"
 repository = "https://github.com/Chia-Network/chia_rs/chia-bls/"
 
+[features]
+py-bindings = ["dep:pyo3", "chia_py_streamable_macro", "chia-traits/py-bindings"]
+
 [dependencies]
 chia-traits = { path = "../chia-traits" }
+clvm-traits = { path = "../clvm-traits" }
```
Rigidity, line 16: These dependencies will need to specify a version, if we're to ever publish these to crates.io

chia-bls/src/gtelement.rs
```diff
@@ -0,0 +1,159 @@
...
+#[cfg(feature = "py-bindings")]
+use chia_py_streamable_macro::PyStreamable;
+#[cfg(feature = "py-bindings")]
+use chia_traits::from_json_dict::FromJsonDict;
+#[cfg(feature = "py-bindings")]
+use chia_traits::to_json_dict::ToJsonDict;
+#[cfg(feature = "py-bindings")]
+use pyo3::exceptions::PyValueError;
+#[cfg(feature = "py-bindings")]
+use pyo3::{pyclass, pymethods, IntoPy, PyAny, PyObject, PyResult, Python};
```
Rigidity, line 20: It'd be preferable if the Python bindings weren't interspersed with the implementation code. This is a problem we have in `chia-protocol` with `PyStreamable` as well, for example. If we were to use a newtype, we could separate the bindings out and add them to the `wheel` crate instead. Much easier to maintain and expand to other language bindings in the future, but we'd need thorough tests to prevent breakage.

chia-bls/src/signature.rs
```diff
@@ -11,14 +14,30 @@ use std::io::Cursor;
...
+    feature = "py-bindings",
+    pyclass(name = "G2Element"),
+    derive(PyStreamable)
+)]
 #[derive(Clone)]
 pub struct Signature(pub(crate) blst_p2);
 
 impl Signature {
-    pub fn from_bytes(buf: &[u8; 96]) -> Result<Self> {
+    pub fn from_bytes_unchecked(buf: &[u8; 96]) -> Result<Self> {
```
Rigidity, line 40: Being named `_unchecked` implies that this function is `unsafe` and that there is a safe version that performs such checks. If this function can cause undefined behavior it should be marked `unsafe`. It also looks like it returns a `Result`, which doesn't make sense if it's unchecked.

chia-bls/src/signature.rs
```diff
@@ -30,7 +49,11 @@ impl Signature {
             blst_p2_from_affine(p2.as_mut_ptr(), &p2_affine.assume_init());
             p2.assume_init()
         };
-        let ret = Self(p2);
+        Ok(Self(p2))
+    }
+
+    pub fn from_bytes(buf: &[u8; 96]) -> Result<Self> {
```
Rigidity, line 55: If this checks `is_valid`, what's the difference between the builtin check done by `p2_uncompress`?

chia-bls/src/gtelement.rs
```diff
@@ -0,0 +1,159 @@
+use blst::*;
+use chia_traits::chia_error::Result;
```
Rigidity, line 2: Why are we reusing this error type for other crates? Would be better for `chia-bls` to have its own concrete error enum with common variants (can use `thiserror` or `impl std::error::Error` to make it a proper error type) instead of a generic `Custom` String variant.

chia-bls/src/signature.rs
```diff
@@ -283,12 +403,27 @@ pub fn sign<Msg: AsRef<[u8]>>(sk: &SecretKey, msg: Msg) -> Signature {
     Signature(p2)
 }
 
+pub fn sign<Msg: AsRef<[u8]>>(sk: &SecretKey, msg: Msg) -> Signature {
```
Rigidity, line 406: Wouldn't it make sense for this to be a method of `SecretKey` since you always need one to sign a message?

chia-bls/src/gtelement.rs
arvidn: the main challenge with that approach is that it would spread to all other chia classes and messages. e.g. most of `chia-protocol`. I think we would need to generate the python types by a macro in order for it to be practical.

The other risk is that objects would need to be copied in- and out of the python versions, which may be expensive in certain cases. like `FullBlock` for instance

chia-bls/src/signature.rs
arvidn: this is the current naming convention in `blspy`. It means the G2 point is not validated to be on the G2 curve. This check is quite expensive. When serializing and deserializing between two trusted party we try to avoid it. But when reading an untrusted signature, it's critical to check it, as the signature verification may give false positives otherwise (at least that's my understanding)

chia-bls/src/signature.rs
arvidn: I don't know. I was not expecting `p2_uncompress()` would check if the point is on the curve, but it looks like it is. at least from the error code names.

chia-bls/src/signature.rs
arvidn: perhaps it would. this is modelled after blspy where there are 3 different ways of signing a message (aug, pop and basic). Chia only uses Aug, so that's the only supported mechanism in `chia-bls`.

In general I think there's a tendency in the world to make free functions members, because that's what Java does, or maybe that's what their IDE suggests. You also always need a message to sign with a key, so maybe it should be a member of the message? I'm not seriously suggesting that, but I bet some would.

chia-bls/src/gtelement.rs
arvidn: no good reason, just laziness. I'm not familiar with `thiserror`. Do you think we should stick to that in the chia repos?

chia-bls/src/gtelement.rs
Rigidity: They can't be moved?

chia-bls/src/gtelement.rs
Rigidity: It makes it easy to derive From for wrapped types, set how each variant is Display'ed, and derive the Error trait

arvidn: I believe this has been addressed outside of this PR now (this PR has been rebased on top of it)

chia-bls/src/signature.rs
arvidn: as far as I can tell, BLST doesn't have a version of `blst_p2_uncompress()` that doesn't also check if the point is on the curve. The only remaining check in `is_valid()` is the infinity check, which may also be redundant. I think this should be improved in a separate PR. I don't believe there's a correctness issue right now, just a performance issue

chia-bls/src/public_key.rs
```diff
@@ -175,6 +286,25 @@ impl DerivableKey for PublicKey {
     }
 }
 
+impl FromClvm for PublicKey {
+    fn from_clvm(a: &Allocator, ptr: NodePtr) -> clvm_traits::Result<Self> {
+        Self::from_bytes(
+            &node_to_bytes(a, ptr)
```
Rigidity, line 292: I don't think this is what you want here, `node_to_bytes` is the serialized CLVM program rather than the atom, which means it includes length information. You need to do:

```rust
if let SExp::Atom = a.sexp(ptr) {
    // a.atom(ptr)
} else {
    Err(clvm_traits::Error::ExpectedAtom)
}
```

chia-bls/src/public_key.rs
```diff
@@ -175,6 +286,25 @@ impl DerivableKey for PublicKey {
...
+                .try_into()
+                .map_err(|_error| clvm_traits::Error::Custom("invalid size".to_string()))?,
+        )
+        .map_err(|error| clvm_traits::Error::Custom(error.to_string()))
+    }
+}
+
+impl ToClvm for PublicKey {
+    fn to_clvm(&self, a: &mut Allocator) -> clvm_traits::Result<NodePtr> {
+        node_from_bytes(a, &self.to_bytes())
```
Rigidity, line 303: Same here, needs to encode this with `a.new_atom` instead of `node_from_bytes`

chia-bls/src/public_key.rs
arvidn: I'm not sure what I was thinking there..

Rigidity (approved):

Rigidity (approved):

AmineKhaldi (approved):

PR #256, opened 2023-08-24
remove `ENABLE_ASSERT_BEFORE` flag

arvidn:
the 1.8.0 softfork has been activated and the `ASSERT_BEFORE` conditions are now enabled unconditionally. We no longer need the flag to be able to disable them, so simplify the code and tests.


richardkiss (approved):
Seems reasonable.

PR #260, opened 2023-09-01
improve run_block_generator2() fidelity with CLVM implementation

arvidn:
the CLVM implementation of the generator (`run_block_generator()`) required the list of spends, returned by the generator, to be terminated by NIL. The rust implementation (`run_block_generator2()`) was fine with it being terminated by any atom.

This fixes the discrepancy as well as extends the fuzzer to actually check fidelity with `run_block_generator()`


richardkiss (approved):
Looks good.

PR #263, opened 2023-09-06
Chia wallet primitive puzzles and hashes

Rigidity:


chia-wallet/src/primitives/cat.rs
```diff
@@ -0,0 +1,167 @@
...
+    "
+);
+
+/// This is the puzzle reveal of the old [CAT1 standard](https://chialisp.com/cats) puzzle.
+///
+/// **Warning:**
+/// It is recommended not to use CAT1 for anything other than backwards compatibility (e.g. offer compression),
+/// due to security issues uncovered in an audit. You can read more about the vulnerability that prompted the creation
+/// of CAT2 in the [CATbleed Post Mortem](https://github.com/Chia-Network/post-mortem/blob/main/2022-08/2022-08-19-CATbleed.md).
+pub static CAT_PUZZLE_V1: [u8; 1420] = hex!(
```
arvidn, line 83: do we still need this?
It seems risky to just keep it around.

chia-wallet/src/primitives/standard.rs
```diff
@@ -0,0 +1,56 @@
...
+    use clvmr::{serde::node_from_bytes, Allocator};
+
+    use super::*;
+
+    #[test]
+    fn standard() {
+        let mut a = Allocator::new();
+        let ptr = node_from_bytes(&mut a, &STANDARD_PUZZLE).unwrap();
+        let hash = tree_hash(&mut a, ptr);
+        assert_eq!(STANDARD_PUZZLE_HASH, hash);
```
arvidn, line 46: Maybe a macro to generate these tests would be appropriate. I think it's important that we have tests for every single pre-computed hash constant, so the hurdle to add one should be minimal

chia-wallet/src/primitives/standard.rs
```diff
@@ -0,0 +1,56 @@
...
+    ff02ffff01ff02ffff03ff0bffff01ff02ffff03ffff09ff05ffff1dff0bffff
+    1effff0bff0bffff02ff06ffff04ff02ffff04ff17ff8080808080808080ffff
+    01ff02ff17ff2f80ffff01ff088080ff0180ffff01ff04ffff04ff04ffff04ff
+    05ffff04ffff02ff06ffff04ff02ffff04ff17ff80808080ff80808080ffff02
+    ff17ff2f808080ff0180ffff04ffff01ff32ff02ffff03ffff07ff0580ffff01
+    ff0bffff0102ffff02ff06ffff04ff02ffff04ff09ff80808080ffff02ff06ff
+    ff04ff02ffff04ff0dff8080808080ffff01ff0bffff0101ff058080ff0180ff
+    018080
+    "
+);
```
arvidn, line 15: Can you think of ways to test that this is correct as well?
We could either have a test that compiles these programs from source and ensures this constant is correct. Or we have some CI job that does that and compares against the constants in here.

arvidn (dismissed):
this looks good. But I really think we need to have tests (or at least CI jobs) that ensure the correctness of these constants back to the source code.

chia-wallet/src/primitives/cat.rs
Rigidity: For offer compression

arvidn (commented):
what would you think of using `include_bytes!()` for the compiled puzzles?
that might make it simpler for a CI job to validate them

chia-wallet/src/primitives/standard.rs
Rigidity: Makes sense

arvidn (approved):

PR #265, opened 2023-09-07
Setup riscv workflow

emlowe:


arvidn:
in an ideal world you would also run `cargo test --all --release`

emlowe:
Takes 2 hours under QEMU to run `cargo build --release && cargo test && cargo test --release`
Do we need to do the build or is the single `cargo test --release` ok?

arvidn:
> Do we need to do the build or is the single cargo test --release ok?

That will both build and run the tests. I would think that's sufficient, especially since it's just for this architecture.

There are some pretty expensive tests, that's why it's good to run in release mode. Alternatively, there might be a way to exclude the generator tests (which are essentially stress tests)

emlowe:
Undraft - takes about 30 mins to build and test the crate as well as build the wheel - parallel jobs though so total time is about 30m

.github/workflows/build-riscv64.yml
```diff
@@ -0,0 +1,80 @@
...
+        docker run --rm --platform linux/riscv64 \
+          -v ${{ github.workspace }}:/ws --workdir=/ws \
+          chianetwork/ubuntu-22.04-risc-builder:latest \
+          bash -exc '\
+          pyenv global 3.10
+          python -m venv venv && \
+          source ./venv/bin/activate && \
+          pip install --upgrade pip && \
+          pip install --extra-index-url https://pypi.chia.net/simple/ maturin==1.2.3 && \
+          maturin build -i python --release -m wheel/Cargo.toml \
```
arvidn, line 73: since you don't run the python tests, building the wheel in release mode will only slow it down. The python tests aren't very demanding, so even running them against a debug build is fine

arvidn: oh, right. you publish the wheel, so it should be in release mode

arvidn (approved):

PR #266, opened 2023-09-14
more BLS operations

arvidn:
add negate, scalar_multiply, hash_to_g1, from_integer to point types as well as hash_to_g2.

These operations are sufficient to make `clvm_rs` use this crate instead of `bls12_381`


Rigidity (approved):

arvidn:
it turns out there's some missing test coverage, especially of `aggregate_pairing`, but also some other failure cases. I'll add that before landing

Rigidity (approved):

PR #267, opened 2023-09-19
fast-forward

arvidn:
this PR is best reviewed one commit at a time.

# overview

It introduces a feature I call "fast forward". Perhaps a better name would be "rebase". What it does is to change the coin being spent by a singleton spend. It obviously only work for a small subset of singleton spends:

1. The puzzle hash may not be altered by the spend, i.e. the state of the singleton must stay the same
2. The inner puzzle or solution may not tie the spend to a specific coin or parent coin (via `AGG_SIG_ME`, `ASSERT_MY_COIN_ID`, `ASSERT_MY_PARENT_ID`, etc.)
3. The output coin, that forms the new state for the singleton, must have the same coin amount as the input coin
4. The spend must use a `singleton_top_layer_v1_1.clsp` outer puzzle.

In order the change the coin being spent, the singleton top layer solution is altered to include a new parent's parent coin ID.

# commits

The change is broken down into the following commits:

1. The core function (`fast_forward_singleton()`) that validates that spend as eligible and produces a new solution for the spend on the new coin.
2. ~~Updates to the conditions parsing, identifying traits of spends as (possibly) being eligible for fast forward, setting the flag `ELIGIBLE_FOR_FF`. This flag does not guarantee that a spend supports fast-forward, but together with the `fast_forward_singleton()` function, it's believed to guarantee that the new spend is valid. (However, it's best to validate it by running the new spend). This flag is primarily meant when running a spend bundle to validate whether a spend can be fast-forwarded or not.~~
3. Add a fuzzer for `fast_forward_spend()` and extend the `gen-corpus` tool to pick out singletons to seed the corpus with
4. Add two new tools (under `chia-tools`): `run-spend` and `fast-forward-spend`. These were very helpful in the development of this feature both for manual testing and understanding of the puzzles. My vision for `run-spend` is to extend it to know about more puzzles in the future, to make it a convenient way of inspecting puzzles on the chain.
5. Add python bindings for this function. For convenience, the python wrapper adds two functions, one to validate whether the spend supports fast-forward, and one that also returns the new solution.

# run-spend

Here's an example output from the `run-spend` tool, for a singleton spend:

```
$ cargo run --bin run-spend ../ff-tests/e3c0.spend
```

```
Spending Coin { parent_coin_info: 24dde051988ea30bb269921e1844051fc42078d1fd561086f235ddff1354e734, puzzle_hash: e3c000a395f8f69d5e263a9548f13bffb1c4b701ab8f3faa03f7647c8750d077, amount: 1 }
   coin-id: fd65e4b0f21322f78d1025e8a8ff7a1df77cd40b86885b851f4572e5ce06e4ff

Conditions

  [73] ASSERT_MY_AMOUNT 1
  [71] ASSERT_MY_PARENT_ID 24dde051988ea30bb269921e1844051fc42078d1fd561086f235ddff1354e734
  [51] CRATE_COIN e3c000a395f8f69d5e263a9548f13bffb1c4b701ab8f3faa03f7647c8750d077 1 

Puzzle Info

Puzzle: e3c000a395f8f69d5e263a9548f13bffb1c4b701ab8f3faa03f7647c8750d077
singleton_top_layer_1_1.clsp
  singleton-struct:
    mod-hash: 7faa3253bfddd1e0decb0906b2dc6247bbc4cf608f58345d173adb63e8b47c9f
    launcher-id: a314920efd80f4ea2c50a2b756c4b9b6d567e2c11e686e1c19e706342d88b021
    launcher-puzzle-hash: eff07522495060c066f66f32acc2a77e3a3e737aca8baea4d1a64ea4cdc13da9
  solution
    lineage-proof: LineageProof { parent_parent_coin_id: 8703edff9f09bb7efd3748b9323aaa29c93f4afb041956158927cee33ef3d485, parent_inner_puzzle_hash: 1590dc8c2190782bf3828bcd9257c8b743d7163a37c6ec6eb9819498809393ef, parent_amount: 1 }
    amount: 1

Inner Puzzle:

Puzzle: 1590dc8c2190782bf3828bcd9257c8b743d7163a37c6ec6eb9819498809393ef
  Unknown puzzle 635fe962cefb0ded17a52293b7057057a6bb1e60417e565ded3c8594e8f8c4ca

```


src/fast_forward.rs
```diff
@@ -0,0 +1,505 @@
...
+    use rstest::rstest;
+    use std::fs;
+    use std::io::Cursor;
+
+    // this test loads CoinSpends from file (Coin, puzzle, solution)-triples
+    // and "fast-forwards" the spend onto a few different parent-parent coins
+    // and ensures the spends are still valid
+    #[rstest]
+    #[case("e3c0")]
+    #[case("bb13")]
```
arvidn, line 194: these are filenames of actual singleton spends I pulled out of the mainnet chain

trepca:
Interesting, can you explain some use cases that this feature could enable? If puzzle hash can't change, is it mainly for oracle-like coins?

chia-tools/src/bin/run-spend.rs
```diff
@@ -0,0 +1,285 @@
...
+    fn debug_print(&self, a: &Allocator) -> String;
+}
+
+impl DebugPrint for NodePtr {
+    fn debug_print(&self, a: &Allocator) -> String {
+        hex::encode(a.atom(*self))
+    }
+}
+
+impl DebugPrint for Condition {
```
richardkiss, line 30: Seems like this could be generally useful and you might want to move it to a separate file.

arvidn:
@trepca yes, it's primarily for oracle-like coins. One of the features it enables is for farmers to fast-forward mempool items and combine multiple spends of the same coin in the same block.

chia-tools/src/bin/run-spend.rs
arvidn: yeah, it would be great to be a trait and to have a derive macro. I'm a bit hesitant to expand the scope of this PR though. I think it would make sense to punt on that until the need for a facility like this comes up again

AmineKhaldi (commented):
Looks reasonable. I think we should expose `ELIGIBLE_FOR_FF` in the wheel and update the PR description to no longer mention `CREATE_COIN_ANNOUNCEMENT` as a restriction for instance.

richardkiss:
I remain of the opinion that adding code to the core node that is aware of specific spend types is not a good idea.

arvidn:
@richardkiss 
> I remain of the opinion that adding code to the core node that is aware of specific spend types is not a good idea.

This PR doesn't do that though

src/fast_forward.rs
```diff
@@ -0,0 +1,505 @@
...
+    pub lineage_proof: LineageProof,
+    pub amount: u64,
+    pub inner_solution: NodePtr,
+}
+
+// TODO: replace this with a generic function to compute the hash of curried
+// puzzles
+fn curry_single_arg(arg_hash: [u8; 32], rest: [u8; 32]) -> [u8; 32] {
+    tree_hash_pair(
+        tree_hash_atom(&[4]),
```
richardkiss, line 45: Maybe use a named constant for `4` so it's more clear what operators are involved.

src/fast_forward.rs
```diff
@@ -0,0 +1,505 @@
...
+        return Err(Error::CoinAmountEven);
+    }
+
+    // in the case of fast-forwarding a spend, we require the amount to remain
+    // unchanged
+    if coin.amount != new_coin.amount || coin.amount != new_parent.amount {
+        return Err(Error::CoinAmountMismatch);
+    }
+
+    // we can only fast-forward spends of singleton's whose puzzle hash doesn't
```
richardkiss, line 101:
```
-    // we can only fast-forward spends of singleton's whose puzzle hash doesn't
+    // we can only fast-forward spends of singletons whose puzzle hash doesn't
```

src/fast_forward.rs
```diff
@@ -0,0 +1,505 @@
...
+        return Err(Error::PuzzleHashMismatch);
+    }
+
+    let singleton = CurriedProgram::<SingletonArgs>::from_clvm(a, puzzle)?;
+    let mut new_solution = SingletonSolution::from_clvm(a, solution)?;
+
+    // this is the tree hash of the singleton top layer puzzle
+    // the tree hash of singleton_top_layer_v1_1.clsp
+    if singleton.args.singleton_struct.mod_hash
+        != hex!("7faa3253bfddd1e0decb0906b2dc6247bbc4cf608f58345d173adb63e8b47c9f")
```
richardkiss, line 113: constant maybe?

chia-tools/src/bin/gen-corpus.rs
```diff
@@ -76,10 +79,16 @@ fn main() {
                             _ => puzzle_hash,
                         };
 
-                        if !seen_puzzles.lock().unwrap().insert(mod_hash) {
+                        const SINGLETON_MOD_HASH: [u8; 32] = hex!(
+                            "7faa3253bfddd1e0decb0906b2dc6247bbc4cf608f58345d173adb63e8b47c9f"
```
richardkiss, line 83: Constant?

richardkiss:
It seems strange to me to look for particular condition outputs deep in the "spend condition aggregator". It's a layer violation that gives too much preference to a particular kind of spend. I get that you don't want to have to run the spend again just to examine the output conditions.

What if we pass in an optional callback in that's run on each iterator at the appropriate time, and it can do whatever it wants? Wherever it needs to go to identify the magic spends and notate them.

src/gen/conditions.rs
```diff
@@ -474,6 +474,15 @@ pub const ELIGIBLE_FOR_DEDUP: u32 = 1;
 // If the spend bundle contained *any* relative seconds or height condition, this flag is set
 pub const HAS_RELATIVE_CONDITION: u32 = 2;
 
+// If the CoinSpend is eligible for fast-forward, this flag is set. A spend is
```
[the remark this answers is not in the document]
arvidn: The original idea was that since it involves reviewing each output coin, it's cheaper to do it when we build the list of outputs instead of a separate pass. However, I ended up deferring this check to perform it only if all other conditions are met.
The only remaining reason to keep these checks here is because of the `condition_counter`. The order the conditions are created in is lost in this function, and the only valid position for `ASSERT_MY_PARENT_ID` is as the second condition. This can't really be checked outside of this function. The order could be recorded perhaps, but it's not obvious that's an overall simplification.

arvidn:
> What if we pass in an optional callback in that's run on each iterator at the appropriate time, and it can do whatever it wants? Wherever it needs to go to identify the magic spends and notate them.

The time when we need to identify or extract information for some other purpose or puzzle would be the perfect time to introduce such abstraction. I don't think there's much value in trying to predict what it would look like, and introduce additional complexity for a potential future use.

That said, I think it *would* reduce complexity to find a way to separate this checking into a separate pass. Iterating the output coins ended up being separate pass anyway.

src/fast_forward.rs
```diff
@@ -79,6 +80,33 @@ fn curry_and_treehash(inner_puzzle_hash: &Bytes32, singleton_struct: &SingletonS
...
+
+        // if this still looks like it might be a singleton, check the output coins
+        // to look for something that looks like a singleton output, with the same
+        // puzzle hash as our input coin
+        // if there is at least one output coin that has the same puzzle hash
+        // and an odd coin amount, it looks like a valid singleton
+        spend
+            .create_coin
+            .iter()
+            .any(|c| (c.amount & 1) == 1 && a.atom(spend.puzzle_hash) == c.puzzle_hash)
```
arvidn, line 102: I made a logic change here as well, to not require the amounts to be the same, just for the new amount to also be odd (as that's a requirement for singletons)

src/gen/conditions.rs
```diff
@@ -2681,7 +2664,7 @@ fn test_multiple_create_coin() {
         amount: 43_u64,
         hint: a.null()
     }));
-    assert_eq!(spend.flags, ELIGIBLE_FOR_DEDUP);
+    assert_eq!(spend.flags, ELIGIBLE_FOR_DEDUP | ELIGIBLE_FOR_FF);
```
arvidn, line 2667: this test exercises the case where there's an output with a different amount, but still odd. This expected result changed because of the logic change above

richardkiss:
Much less controversial now.

richardkiss (approved):

PR #268, opened 2023-09-20
validate program when parsing it from JsonDict

arvidn:
when parsing a string from a Json Dict containing a CLVM program, reject strings that have garbage at the end, or otherwise has invalid CLVM serialization.


Rigidity (approved):
lgtm

PR #270, opened 2023-09-22
fix bug in enforcing nil-termination of new AGG_SIG_* conditions

arvidn:
This fixes a bug when parsing the new `AGG_SIG_*` conditions in mempool mode.
The affected conditions are:

1. `AGG_SIG_PARENT`
2. `AGG_SIG_PUZZLE`
3. `AGG_SIG_AMOUNT`
4. `AGG_SIG_PUZZLE_AMOUNT`
5. `AGG_SIG_PARENT_PUZZLE`
6. `AGG_SIG_PARENT_AMOUNT`

In mempool mode, we require the exact correct number of arguments to be passed to conditions. Whereas in consensus mode we're more forgiving, to allow for soft-forks. The test for an NIL-terminator was incorrect and always looked at the last *cons* cell, not the last item in the list. This made this test fail unconditionally.

The bulk of this change is to extend the tests to cover this case.

The original feature landed here: https://github.com/Chia-Network/chia_rs/pull/213

This PR is targeting a new branch called `release-0.2.11` which was cut off of the most recent `0.2.10` release, to avoid pulling in any other changes for this release.

### DIFF
--- a/src/gen/conditions.rs
+++ b/src/gen/conditions.rs
@@ -186,7 +186,7 @@ pub fn parse_args(
             // AGG_SIG_* take two parameters
 
             if (flags & STRICT_ARGS_COUNT) != 0 {
-                check_nil(a, c)?;
+                check_nil(a, rest(a, c)?)?;
             }
             match op {
                 AGG_SIG_PARENT => Ok(Condition::AggSigParent(pubkey, message)),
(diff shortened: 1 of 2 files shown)


src/gen/conditions.rs
```diff
@@ -186,7 +186,7 @@ pub fn parse_args(
             // AGG_SIG_* take two parameters
 
             if (flags & STRICT_ARGS_COUNT) != 0 {
-                check_nil(a, c)?;
+                check_nil(a, rest(a, c)?)?;
```
arvidn, line 189: this is the bugfix. the nil-terminator is expected to be the *next* item in the list, not the last cons-cell.
The rest are extended tests to cover this and possible other differences in mempool-mode and consensus-mode

richardkiss (approved):
Looks good.

PR #278, opened 2023-10-05
tool to bump crate versions

arvidn:
This tool checks each crate for changes since the last release tag (as specified on the command line). Any crate that has had changes has its version updated to the latest version (also specified on the command line).

All internal dependencies of crates are update to match this version.

This is meant to simplify and make releases less error prone.

### DIFF
--- a/bump-version.py
+++ b/bump-version.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+
+# usage:
+# bump-version <new-version-string> <previous-release-tag>
+
+import os
+import re
+import sys
+from typing import Callable, Set
+
+v = sys.argv[1]
+tag = sys.argv[2]
+
+our_crates = [
+    "chia-bls",
+    "clvm-traits",
+    "chia-traits",
+    "chia_py_streamable_macro",
+    "chia_streamable_macro",
+    "chia-protocol",
+    "chia-tools",
+    "clvm-utils",
+    "clvm-derive",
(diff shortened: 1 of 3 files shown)


bump-version.py
```diff
@@ -0,0 +1,68 @@
...
+                subst += line
+                continue
+
+            if split[0] == "version" and name in crates:
+                line = f"version = \"{v}\"\n"
+            elif split[0] in crates:
+                line = re.sub('version = "(>?=?)\d\.\d\.\d"', f'version = "\\g<1>{v}"', line)
+            subst += line
+
+    with open(f"{name}/Cargo.toml", "w+") as f:
```
AmineKhaldi, line 51: Maybe "w" as we don't need to read from the file here, we just write to it?

bump-version.py
```diff
@@ -0,0 +1,68 @@
...
+    subst = ""
+    with open(f"{name}/Cargo.toml") as f:
+        for line in f:
+            split = line.split()
+            if split == []:
+                subst += line
+                continue
+
+            if split[0] == "version" and name in crates:
+                line = f"version = \"{v}\"\n"
```
AmineKhaldi, line 46: Maybe using '' instead of "" to avoid extra escape sequences?

AmineKhaldi (approved):
Looks reasonable, I left a couple comments.

AmineKhaldi (approved):

PR #280, opened 2023-10-15
Make BLS types support bytes as input in from_json_dict()

arvidn:
in addition to hex strings. This is something the tests in chia-blockchain seem to expect.
Also fix typo in `generate_type_stubs.py`.

This patch is best viewed ignoring whitespace, since there's indentation change.


wheel/chia_rs.pyi
```diff
@@ -75,7 +75,7 @@ class AugSchemeMPL:
     @staticmethod
     def derive_child_sk_unhardened(pk: PrivateKey, index: int) -> PrivateKey: ...
     @staticmethod
-    def derive_child_pk_unhardened(pk: PublicKey, index: int) -> PublicKey: ...
```
arvidn, line 78: the `PublicKey` and `Signature` types are exported as `G1Element` and `G2Element` to python, to stay compatible with `blspy`

AmineKhaldi (approved):

PR #284, opened 2023-10-18
fixup member functions for ClassgroupElement

arvidn:
This is to match the corresponding types in `chia-blockchain`.

https://github.com/Chia-Network/chia-blockchain/blob/main/chia/types/blockchain_format/classgroup.py#L21

### DIFF
--- a/chia-protocol/src/classgroup.rs
+++ b/chia-protocol/src/classgroup.rs
@@ -7,18 +7,42 @@ use pyo3::prelude::*;
 
 streamable_struct!(ClassgroupElement { data: Bytes100 });
 
-#[cfg(feature = "py-bindings")]
-#[cfg_attr(feature = "py-bindings", pymethods)]
 impl ClassgroupElement {
-    #[staticmethod]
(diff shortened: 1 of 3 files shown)


chia-protocol/src/classgroup.rs
```diff
@@ -7,18 +7,42 @@ use pyo3::prelude::*;
...
+    #[staticmethod]
+    #[pyo3(name = "get_default_element")]
+    pub fn py_get_default_element() -> ClassgroupElement {
+        Self::get_default_element()
+    }
+
+    #[staticmethod]
+    #[pyo3(name = "get_size")]
+    pub fn py_get_size() -> i32 {
+        Self::get_size()
```
arvidn, line 46: the reason for having the underlying rust functions be separate is that the PyO3-attributes (e.g. `#[staticmethod]`) won't work if the python bindings aren't enabled. They aren't enabled for `cargo clippy` for instance. Separating them like this, and disabling the whole `impl ClassgroupElement` when the python bindings are disabled, makes it work.

AmineKhaldi (approved):

PR #289, opened 2023-10-24
introduce ELIGIBLE_FOR_FF flag on Spends via policy class

arvidn:
when parsing and validating conditions. This indicates whether some of the requirements are met to be able to fast-forward a singleton spend.

This patch also introduces a customization class (`ConditionPolicy`) to hook into the parsing of conditions. the `ELIGIBLE_FOR_DEDUP` is moved out into the mempool-mode policy, and the new `ELIGIBLE_FOR_FF` is added to that same policy class.

The other policy is `NonePolicy`, which is used in consensus mode, which doesn't do any extra analysis of spends.

Given the simplicity of the two spend analyses (implemented in `MempoolPolicy`) it's not obvious to me that this abstraction and customization point carries its own weight. I think it's a net increase in complexity.

All the "business logic" for this patch is concentrated here: https://github.com/Chia-Network/chia_rs/pull/289/files#diff-f80b284fd6e11b143a06f56f735b48c7901525f11ff62bed4eb15fc1aa08a1ccR64-R138

The rest is the patch is plumbing and tests. All the tests are only covering the `MempoolPolicy` as the `NonePolicy` is empty.

### DIFF
--- a/fuzz/fuzz_targets/fast-forward.rs
+++ b/fuzz/fuzz_targets/fast-forward.rs
@@ -1,5 +1,6 @@
 #![no_main]
 use chia::fast_forward::fast_forward_singleton;
+use chia::gen::conditions::MempoolPolicy;
 use chia::gen::run_puzzle::run_puzzle;
 use chia::gen::validation_error::ValidationErr;
 use chia_protocol::Bytes32;
@@ -64,6 +65,7 @@ fuzz_target!(|data: &[u8]| {
         spend.coin.amount,
         11000000000,
         0,
+        &mut MempoolPolicy::default(),
     );
 
     // run new spend
@@ -75,6 +77,7 @@ fuzz_target!(|data: &[u8]| {
         new_coin.amount,
         11000000000,
         0,
+        &mut MempoolPolicy::default(),
     );
 
     match (conditions1, conditions2) {
--- a/fuzz/fuzz_targets/fast-forward.rs
+++ b/fuzz/fuzz_targets/fast-forward.rs
@@ -1,6 +1,6 @@
 #![no_main]
 use chia::fast_forward::fast_forward_singleton;
-use chia::gen::conditions::MempoolPolicy;
+use chia::gen::conditions::MempoolVisitor;
 use chia::gen::run_puzzle::run_puzzle;
 use chia::gen::validation_error::ValidationErr;
 use chia_protocol::Bytes32;
@@ -57,7 +57,7 @@ fuzz_target!(|data: &[u8]| {
     let new_solution = node_to_bytes(&a, new_solution).expect("serialize new solution");
 
     // run original spend
-    let conditions1 = run_puzzle(
+    let conditions1 = run_puzzle::<MempoolVisitor>(
         &mut a,
         spend.puzzle_reveal.as_slice(),
         spend.solution.as_slice(),
@@ -65,11 +65,10 @@ fuzz_target!(|data: &[u8]| {
         spend.coin.amount,
         11000000000,
         0,
-        &mut MempoolPolicy::default(),
     );
 
     // run new spend
-    let conditions2 = run_puzzle(
+    let conditions2 = run_puzzle::<MempoolVisitor>(
         &mut a,
         spend.puzzle_reveal.as_slice(),
         new_solution.as_slice(),
@@ -77,7 +76,6 @@ fuzz_target!(|data: &[u8]| {
         new_coin.amount,
         11000000000,
         0,
-        &mut MempoolPolicy::default(),
     );
 
     match (conditions1, conditions2) {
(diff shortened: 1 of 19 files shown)


wheel/src/run_generator.rs
```diff
@@ -166,7 +167,27 @@ pub fn run_block_generator(
...
 
-    match native_run_block_generator(&mut allocator, program, &refs, max_cost, flags) {
+    let r = if (flags & ANALYZE_SPENDS) == 0 {
+        native_run_block_generator(
+            &mut allocator,
+            program,
+            &refs,
+            max_cost,
+            flags,
+            &mut NonePolicy::default(),
```
richardkiss, line 177: You can switch on just the single different parameter to make this a bit less repetitive.

```suggestion
            if (flags & ANALYZE_SPENDS) == 0 {&mut NonePolicy::default()} else {&mut MempoolPolicy::default()}
```
or whatever the rust ternary operator is.

richardkiss (commented):
I think it can be simplified by only calling the policy thing one time, passing in all parameters including the list of conditions. Yeah, it would have to iterate over the conditions itself, but that doesn't seem like too big a deal.

As for the name `Policy`, I think of it more as an observer. For that reason, I think it's a bit odd to pass in the `spend` mutable. It also gives the observer a chance to mutate and invalidate spends (maybe you want this for the mempool though?). If you pass the spend's index, the observer can just keep a corresponding list of additional properties for each spend that it thinks relevant.

I do agree this would increase complexity further. The benefit would be further insulating the consensus. But maybe that's not necessary.

Do you plan to parameterize compilation on the policy (so compile-time changes to the execution path rather than runtime)?

wheel/src/run_generator.rs
arvidn: the policy type is a template parameter, so I don't think that would work

arvidn:
> I think it can be simplified by only calling the policy thing one time, passing in all parameters including the list of conditions. Yeah, it would have to iterate over the conditions itself, but that doesn't seem like too big a deal.

I don't currently build a list of the conditions, they are coalesced and possibly ignored as they are parsed. Building this list just for the purpose of this customization point would, by far, outweigh the cost of the work being done to detect eligibility for dedup and fast-forward.

I can see an option where the "policy" parses the conditions itself, but it seems quite risky to do the parsing twice. There are a few risks in the parsing.

> As for the name `Policy`, I think of it more as an observer. For that reason, I think it's a bit odd to pass in the `spend` mutable.

I agree that "Policy" is not a great name here. It is closer to an observer, at least if it would keep its state itself rather than updating the `Spend`. Doing that, though, would not only increase complexity further, but also (most likely) dwarf the cost of the work being done to detect these kinds of spends. Passing `Spend` as a mutable reference, to some extent, justifies calling the customization point "Policy", so does the fact that it's a template argument.

> It also gives the observer a chance to mutate and invalidate spends (maybe you want this for the mempool though?). If you pass the spend's index, the observer can just keep a corresponding list of additional properties for each spend that it thinks relevant.

One possibility would be to pass in `Spend` as an immutable reference and have it return flags to set. It would insulate the main logic from the customization point's ability to mutate spends. I think this would be getting into the territory of trying to predict future uses of this interface at the expense of complexity. If we need to do more here in the future, I expect it to look completely different.

> I do agree this would increase complexity further. The benefit would be further insulating the consensus. But maybe that's not necessary.

The insulation of the consensus-only logic from the mempool logic comes in the template argument itself. As long as the consensus-only version of this function uses the `NonePolicy`, it doesn't really matter what the policy is *able* to do, the consensus logic is still insulated. That said, the mempool logic also matters, and it's not OK for it to misbehave. This is partly why I think overall complexity probably trumps the separation of consensus-only vs. mempool logic.

> Do you plan to parameterize compilation on the policy (so compile-time changes to the execution path rather than runtime)?

It is a template parameter already, that's partly why I named it "Policy", as in policy-based design.

At first I was planning to tweak the customization point as I described above, but as I was thinking about it, I'm not sure it's an improvement. I will try to come up with a better name than "Policy" and see if I can tweak it in some way to simplify it.

wheel/src/run_generator.rs
arvidn: Originally I was thinking that passing in the policy as an object would open up the possibility to pass in custom state here. But since we don't have a use case for that yet, it would probably be a simplification to just pass in the type, and make the trait require `Default`. The "Policy" needs to be able to keep state, but (for now) there's no need to pass in state.

fuzz/fuzz_targets/fast-forward.rs
```diff
@@ -57,27 +57,25 @@ fuzz_target!(|data: &[u8]| {
...
-    let conditions2 = run_puzzle(
+    let conditions2 = run_puzzle::<MempoolVisitor>(
         &mut a,
         spend.puzzle_reveal.as_slice(),
         new_solution.as_slice(),
         &new_coin.parent_coin_info,
         new_coin.amount,
         11000000000,
         0,
-        &mut MempoolPolicy::default(),
```
richardkiss: I think you have to leave this line in. Otherwise we can't ever add a callback that has a context, like a mempool sniffer would almost certainly need.

richardkiss (changes requested):
I think we need to pass in an instance of the thing that we're parameterizing on.

richardkiss (approved):

arvidn:
I think it's not obvious what requirements this interface would have for implementing a mempool sniffer. In an ideal world it would also be possible to disable validation as well, and just parse the conditions.

PR #291, opened 2023-10-25
add all member functions to Program

arvidn:
matching `chia.types.blockchain_format.SerializedProgram`.

The main challenge in this PR is to mimic the `Program.to()` of the python version. This converts arbitrary python objects into a CLVM structure, to be passed as arguments to programs. The rust implementation of this is `clvm_convert()`.
The python implementation can be found here:
https://github.com/Chia-Network/clvm/blob/main/clvm/SExp.py#L59-L115
and here:
https://github.com/Chia-Network/clvm/blob/main/clvm/SExp.py#L38-L55

There are tests for all types supported in this conversion on the rust side, as well as some testing to make sure the rust implementation behaves the same as the python one (`test_program_fidelity.py`)


chia-protocol/src/program.rs
```diff
@@ -59,6 +50,257 @@ impl Program {
...
+            // the args of curried puzzles are in the form of:
+            // (c . ((q . <arg1>) . (<rest> . ())))
+            let (_, ((_, arg), (rest, _))) =
+                <(
+                    clvm_traits::MatchByte<4>,
+                    (clvm_traits::match_quote!(NodePtr), (NodePtr, ())),
+                ) as clvm_traits::FromClvm>::from_clvm(&a, args)?;
+            curried_args.push(arg);
+            args = rest;
+        }
```
arvidn, line 292: @Rigidity I couldn't find a better way of uncurrying an arbitrary number of arbitrary arguments. There's no trick to get `CurriedProgram` to do this, right?

Rigidity: Not currently. I want to improve a bunch of the clvm-traits stuff in the future and make the code cleaner, but I don't have time right now.

.github/workflows/benchmark.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
```
arvidn, line 22: we no longer distribute `chia-blockchain` for python 3.7, so testing against it here pulls down an old version (1.8.2) which has a slightly different interface to `Program` (it doesn't have the `_run()` method).

chia-protocol/Cargo.toml
```diff
@@ -12,15 +12,16 @@ repository = "https://github.com/Chia-Network/chia_rs/chia-protocol/"
 py-bindings = ["dep:pyo3", "dep:chia_py_streamable_macro", "chia-traits/py-bindings"]
 
 [dependencies]
-pyo3 = { version = "0.19.0", features = ["multiple-pymethods"], optional = true }
+pyo3 = { version = "0.19.0", features = ["multiple-pymethods", "num-bigint"], optional = true }
```
arvidn, line 15: the `num-bigint` feature is to enable support PyO3 support for converting python `int` to `BigInt` types in rust.

chia-protocol/src/program.rs
```diff
@@ -59,6 +50,270 @@ impl Program {
...
+            .map_err(|e| PyMemoryError::new_err(e.to_string()))
+    } else {
+        Err(PyTypeError::new_err(format!(
+            "unknown parameter to run_with_cost() {o}"
+        )))
+    }
+}
+
+#[cfg(feature = "py-bindings")]
+fn to_program(py: Python<'_>, node: LazyNode) -> PyResult<&PyAny> {
```
arvidn, line 165: This is not great, but in order to be compatible with the current Python version of `SerializedProgram` we need to be able to convert to and from the `Program` type, which is based on `SExp`

PR #302, opened 2023-10-29
Wallet primitive serialization

Rigidity:


clvm-utils/src/curry_tree_hash.rs
```diff
@@ -0,0 +1,57 @@
+use crate::{tree_hash_atom, tree_hash_pair};
+
+pub fn curry_tree_hash(program_hash: [u8; 32], arg_hashes: &[[u8; 32]]) -> [u8; 32] {
+    let nil = tree_hash_atom(&[]);
+    let one = tree_hash_atom(&[1]);
+    let op_q = one;
```
arvidn, line 6:
```suggestion
    let op_q = tree_hash_atom(&[1]);
```
it looks a little bit redundant

chia-wallet/src/puzzles/nft.rs
```diff
@@ -1,5 +1,142 @@
...
+                _ => (),
+            }
+        }
+
+        Ok(metadata)
+    }
+}
+
+impl ToClvm for NftMetadata {
+    fn to_clvm(&self, a: &mut Allocator) -> Result<NodePtr> {
```
arvidn, line 104: I think it would be good to have round-trip tests for the types that implement custom `to_clvm()` and `from_clvm()`, to ensure they match. One think I added recently was a fuzzer for serialization round-trips.

If you implement `Arbitrary` (at least when built with fuzzing enabled) it's quite easy to make a fuzzer that turns an object into CLVM and then back again, to test round-trips.

arvidn: we only run the fuzzers for 20 seconds or so. I investigated this a while back and the CI run-time was dominated by link-time. So I disabled LTO, which sped it up quite a lot. Linking each fuzzer binary still takes quite a while though.

There's also some work to offload the fuzzing to separate machines, probably outside of github actions.

arvidn (dismissed):

Rigidity:
I added a trait and implementations of synthetic key derivation for secret keys and public keys, similar to what I had on the wallet-dev branch.

chia-wallet/src/puzzles/nft.rs
Rigidity: I believe I added fuzzing correctly. I also added code to implement synthetic key derivation, so another review will be needed for these two things

chia-wallet/src/derive_synthetic.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+impl DeriveSynthetic for SecretKey {
+    fn derive_synthetic(&self, hidden_puzzle_hash: &[u8; 32]) -> Self {
+        self + &synthetic_offset(&self.public_key(), hidden_puzzle_hash)
+    }
+}
+
+pub fn hash_to_secret_key(bytes: [u8; 32]) -> SecretKey {
```
arvidn, line 21: is this different from `SecretKey::from_seed()`?
Perhaps the name of this function could highlight the difference, or the specific kind of key generation it uses

chia-wallet/src/derive_synthetic.rs
```diff
@@ -0,0 +1,125 @@
+use chia_bls::{PublicKey, SecretKey};
+use num_bigint::BigInt;
+use sha2::{digest::FixedOutput, Digest, Sha256};
+
+pub trait DeriveSynthetic {
```
arvidn, line 5: is synthetic derivation definitely a wallet feature?
I'm not suggesting it belongs in `chia-bls`, I just want to point out that both hardened- and unhardened derivation is implemented there.

chia-wallet/src/proof.rs
```diff
@@ -0,0 +1,40 @@
+use clvm_traits::{FromClvm, Result, ToClvm};
+use clvmr::{allocator::NodePtr, Allocator};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Proof {
```
arvidn, line 5: I was hoping this would implement `Arbitrary` too, to get some round-trip coverage of `from_clvm` and `to_clvm`

chia-wallet/src/puzzles/did.rs
```diff
@@ -1,5 +1,45 @@
+use clvm_traits::{clvm_list, match_list, match_tuple, Error, FromClvm, Result, ToClvm};
+use clvmr::{allocator::NodePtr, Allocator};
 use hex_literal::hex;
 
+use crate::singleton::SingletonStruct;
+
+#[derive(Debug, Clone, PartialEq, Eq, ToClvm, FromClvm)]
+#[clvm(curried_args)]
+pub struct DidArgs {
```
arvidn, line 9: I suppose `NodePtr` is a bit difficult to implement `Arbitrary` for. maybe it should just be initialized to `nul()` or `one()`.
`DidArgs` also has custom `to_clvm` and `from_clvm`

arvidn (commented):
looks good. excuse my ignorance of the synthetic key derivation. I'm mostly want to make sure it belongs in the wallet or the chia BLS library

chia-wallet/src/derive_synthetic.rs
```diff
@@ -0,0 +1,125 @@
...
+impl DeriveSynthetic for SecretKey {
+    fn derive_synthetic(&self, hidden_puzzle_hash: &[u8; 32]) -> Self {
+        self + &synthetic_offset(&self.public_key(), hidden_puzzle_hash)
+    }
+}
+
+pub fn hash_to_secret_key(bytes: [u8; 32]) -> SecretKey {
+    let value = BigInt::from_signed_bytes_be(bytes.as_slice());
+    let group_order = BigInt::parse_bytes(
+        b"73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001",
```
arvidn, line 24: this looks like ideal for the `hex!` macro, from hex-literal

chia-wallet/src/derive_synthetic.rs
```diff
@@ -0,0 +1,125 @@
...
+        16,
+    )
+    .unwrap();
+    let modulo = ((&value % &group_order) + &group_order) % &group_order;
```
arvidn, line 28: I would expect you could move `value` into this expression, but perhaps the operator doesn't implement that variant

chia-wallet/Cargo.toml
```diff
@@ -10,8 +10,17 @@ repository = "https://github.com/Chia-Network/chia_rs/chia-wallet/"
 
 [dependencies]
 clvmr = "0.3.0"
+sha2 = "0.9.9"
```
arvidn, line 13: I just bumped the other crates to use "0.10.8" of the sha2 dependency

chia-wallet/src/puzzles/did.rs
Rigidity: Assuming you meant DidSolution, since DidArgs is derived. Added a unit test, since nil is only a single value and there's no need to implement Arbitrary for it.

chia-wallet/src/derive_synthetic.rs
Rigidity: The only use of it is in the standard transaction. Though given that it derives keys, I'm fine moving into chia-bls, don't really have an opinion either way.

arvidn (approved):

PR #303, opened 2023-10-29
bump blst dependency, and build it in portable mode

arvidn:
`blspy` builds it in portable mode, and not being portable is likely what's causing CI in chia-blockchain to fail with `SIGILL`.

https://github.com/Chia-Network/bls-signatures/blob/main/src/CMakeLists.txt#L47


hoffmang9 (approved):
lgtm

PR #306, opened 2023-10-29
make `from_bytes()` accept python's buffer API

arvidn:
otherwise it's not possible to directly pass in a slice of a `bytes` object. i.e. `.from_bytes(b[1:])`. It would have to be copied into a new `bytes` object first, i.e. `.from_bytes(bytes(b[1:]))`

### DIFF
--- a/chia_py_streamable_macro/src/lib.rs
+++ b/chia_py_streamable_macro/src/lib.rs
@@ -120,8 +120,14 @@ pub fn py_streamable_macro(input: proc_macro::TokenStream) -> proc_macro::TokenS
         impl #ident {
             #[staticmethod]
             #[pyo3(name = "from_bytes")]
-            pub fn py_from_bytes(blob: &[u8]) -> pyo3::PyResult<Self> {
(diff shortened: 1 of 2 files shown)


chia_py_streamable_macro/src/lib.rs
```diff
@@ -120,8 +120,14 @@ pub fn py_streamable_macro(input: proc_macro::TokenStream) -> proc_macro::TokenS
         impl #ident {
             #[staticmethod]
             #[pyo3(name = "from_bytes")]
-            pub fn py_from_bytes(blob: &[u8]) -> pyo3::PyResult<Self> {
-                let mut input = std::io::Cursor::<&[u8]>::new(blob);
+            pub fn py_from_bytes(blob: pyo3::buffer::PyBuffer<u8>) -> pyo3::PyResult<Self> {
```
arvidn, line 123: taking a `PyBuffer` here is the key difference

AmineKhaldi (approved):

PR #313, opened 2023-11-04
optimize Vec<T>::parse(), for Streamable

arvidn:
by pre-allocating its size

### DIFF
--- a/chia-traits/src/streamable.rs
+++ b/chia-traits/src/streamable.rs
@@ -115,9 +115,12 @@ impl<T: Streamable> Streamable for Vec<T> {
     fn parse(input: &mut Cursor<&[u8]>) -> Result<Self> {
         let len = u32::parse(input)?;
 
-        // TODO: pre-allocate capacity, but we'd need safe-guards for overflow
-        // attacks
-        let mut ret = Vec::<T>::new();
+        let mut ret = if std::mem::size_of::<T>() == 0 {
+            Vec::<T>::new()
+        } else {
+            let limit = 2 * 1024 * 1024 / std::mem::size_of::<T>();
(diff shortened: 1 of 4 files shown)


chia-traits/src/streamable.rs
```diff
@@ -115,9 +115,8 @@ impl<T: Streamable> Streamable for Vec<T> {
     fn parse(input: &mut Cursor<&[u8]>) -> Result<Self> {
         let len = u32::parse(input)?;
 
-        // TODO: pre-allocate capacity, but we'd need safe-guards for overflow
-        // attacks
-        let mut ret = Vec::<T>::new();
+        let limit = 2 * 1024 * 1024 / std::mem::size_of::<T>();
```
Rigidity, line 118: Not sure if it matters in this context, but if the type is zero-sized, this will panic. In this edge case, the capacity should default to `usize::MAX`, since no allocation is required.

arvidn: right. I was considering this case. I assume it would fail to compile, since this is a constant expression. (which I think is fine)

Rigidity: I think it would need to be marked as `const` rather than `let` to be evaluated at compile-time

arvidn: it appears `const` can't depend on template parmeters in rust. I get:
```
error[E0401]: can't use generic parameters from outer item
   --> chia-traits/src/streamable.rs:118:68
    |
95  | impl<T: Streamable> Streamable for Vec<T> {
    |      - type parameter from outer item
...
118 |         const LIMIT: usize = 2 * 1024 * 1024 / std::mem::size_of::<T>();
    |                                                                    ^ use of generic parameter from outer item
```

AmineKhaldi (approved):

PR #314, opened 2023-11-12
Redesign clvm-traits to be more generic

Rigidity:
This PR overhauls the `clvm-traits` and `clvm-derive` crates, making them more generic.

This is to simplify implementing the following functionality when needed in the future:

- Ability to serialize and deserialize CLVM values directly to bytes.
- Ability to calculate the tree hash of a value without going through the CLVM `Allocator`.
- Walk trees produced by CLVM values for any other reason, such as debugging.

Task list:

- [X] Rewrite `ToClvm` and `FromClvm` traits to be more generic, and update `clvm-derive` accordingly.
- [X] Simplify the process of implementing the traits with macros, to get around lack of support for `impl Trait` in type aliases.
- [x] Improve the code quality of `clvm-derive`, and rename `curried_args` and `proper_list` to `curry` and `list`, respectively.
- [x] Add support for deriving enums with the enum variant encoded as the first argument in a tuple or list.
- [x] Add support for deriving enums with no discriminant (a simple `or` operation).
- [x] Settle on final naming for `a.value_to_ptr` and `a.value_from_ptr` from `AllocatorExt`. Perhaps this should be a top-level function.
- [x] Update documentation, and clean up doc tests.
- [ ] Thoroughly review every change and make sure everything is tested. It's critical that no major bugs surface in production wallet code.

Out of scope for this PR:
- Implement the use of the derive macros for various existing structs and enums.
- Add support for currying and uncurrying known quantities of arguments without allocating on the heap.
- Get rid of the various redundant implementations of `SingletonArgs` and `SingletonSolution` split amongst the code base.


Rigidity:
@arvidn this should be good to review, tests are passing on my end and I finished the last few things.

clvm-traits/src/conversions.rs
```diff
@@ -0,0 +1,104 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_tree_hash() {
+        let value = (42, ([1, 2, 3], "Hello, world!"));
+        let ptr = value.tree_hash().unwrap().0;
+        assert_eq!(
+            hex::encode(ptr),
+            "7c3670f319e07cff6d433e4c22e0895f1f0a10bad5bbcd23c32e3bc5589c23cb"
```
arvidn, line 101: it's not obvious that this hash is correct by inspecting this test. I think it would be good to test this implementation of tree_hash against the existing implementations. Maybe even in a fuzzer.

clvm-traits/src/error.rs
```diff
@@ -1,34 +1,32 @@
...
-    Allocator(EvalErr),
+use thiserror::Error;
 
-    #[error("expected atom")]
-    ExpectedAtom(NodePtr),
+/// Any errors that may occur while converting a Rust value to a CLVM value.
+#[derive(Error, Debug, Clone, PartialEq, Eq)]
+pub enum ToClvmError {
+    #[error("limit reached")]
+    LimitReached,
```
arvidn, line 9: it would be less generic and more concrete (I think) to call this something like out of memory. As far as I can tell, it's only used when the Allocator hits its memory limit

clvm-traits/src/conversions.rs
```diff
@@ -0,0 +1,104 @@
...
+{
+    fn to_ptr(&self, a: &mut Allocator) -> Result<NodePtr, ToClvmError> {
+        self.to_clvm(&mut |value| match value {
+            ClvmValue::Atom(bytes) => match bytes {
+                [] => Ok(a.null()),
+                [1] => Ok(a.one()),
+                _ => Ok(a.new_atom(bytes).or(Err(ToClvmError::LimitReached))?),
+            },
+            ClvmValue::Pair(first, rest) => {
+                Ok(a.new_pair(first, rest).or(Err(ToClvmError::LimitReached))?)
```
arvidn, line 27: Maybe I misunderstand something, how come you don't need to call `to_ptr()` on `first` and `rest`? Have they already been converted to allocations in `a`?

clvm-traits/src/lib.rs
```diff
@@ -90,9 +96,9 @@ mod tests {
     }
 
     #[test]
-    fn test_proper_list() {
+    fn test_list() {
```
arvidn, line 99: renaming `proper_list` -> `list` and `curried_args` -> `args` are not convtroversial. I would land that change right away.

clvm-traits/src/error.rs
```diff
@@ -1,34 +1,30 @@
...
+}
 
-    #[error("expected cons")]
-    ExpectedCons(NodePtr),
+#[derive(Error, Debug, Clone, PartialEq, Eq)]
+pub enum FromClvmError {
```
arvidn, line 14: it seems reasonable to split up the error codes for these

arvidn (commented):
I think your first approach (where `Allocator` was a template and could be swapped out) was simpler than this approach. As I mention in the comments, I'm not excited about the basic structure of this abstraction being so complex you need a macro to implement the trait.

I'm thinking that *maybe* it would even make sense to not reuse the `ToClvm` and `FromClvm` traits. What would you think of, instead add a new trait, something like `ClvmSerialize` and `ClvmDeserialize` which converts to- and from serialized binary representation. There will be a little bit of overlap with `ToClvm` and `FromClvm`, but not very much, and I think the resulting code will be a lot more legible.

I'm not convinced that it's important to be able to compute the tree hash directly via this trait. The most efficient tree hash implementations are not recursive anyway, and it seems risky to introduce a sub-optimal one. Perhaps efficient tree-hash of rust types should even be yet another trait. I think that would keep things really simple too

src/fast_forward.rs
```diff
@@ -110,8 +112,8 @@ pub fn fast_forward_singleton(
         return Err(Error::PuzzleHashMismatch);
     }
 
-    let singleton = CurriedProgram::<SingletonArgs>::from_clvm(a, puzzle)?;
-    let mut new_solution = SingletonSolution::from_clvm(a, solution)?;
+    let singleton: CurriedProgram<NodePtr, SingletonArgs<NodePtr>> = FromPtr::from_ptr(a, puzzle)?;
```
arvidn, line 115: why do you need the `NodePtr` template argument here? isn't that implied by `FromPtr`?

src/fast_forward.rs
```diff
@@ -213,8 +215,8 @@ mod tests {
         let new_parents_parent = hex::decode(new_parents_parent).unwrap();
 
         let mut a = Allocator::new_limited(500000000, 62500000, 62500000);
-        let puzzle = spend.puzzle_reveal.to_clvm(&mut a).expect("to_clvm");
-        let solution = spend.solution.to_clvm(&mut a).expect("to_clvm");
+        let puzzle = node_from_bytes(&mut a, spend.puzzle_reveal.as_slice()).expect("to_clvm");
+        let solution = node_from_bytes(&mut a, spend.solution.as_slice()).expect("to_clvm");
```
arvidn, line 219: I think we should preserve the ability for `Program` to convert to and from `NodePtr`

src/fast_forward.rs
Rigidity: You would think. I think the Rust compiler has limitations on what it can infer

src/fast_forward.rs
Rigidity: Yeah, I agree - I'll make sure this is included

PR #315, opened 2023-11-13
`SpendBundle` members

arvidn:

The core of this patch is adding the following member functions to the `SpendBundle` type:
```
    @staticmethod
    def aggregate(sbs: List[SpendBundle]) -> SpendBundle: ...
    def name(self) -> bytes32: ...
    def removals(self) -> List[Coin]: ...
    def additions(self) -> List[Coin]: ...
    def debug(self) -> None: ...
```

By far the most complex one is `additions()`, which actually runs the puzzles in the spend bundle, and pulls out the `CREATE_COIN` conditions to return the coins being created by the spend bundle.

Most of the additional code are tests, a fuzzer and an extension to the corpus generation tool to pull out spend bundles from the block chain database.


Rigidity (approved):
looks good

PR #316, opened 2023-11-16
Rename clvm-derive representations

Rigidity:


clvm-derive/src/helpers.rs
```diff
@@ -33,37 +42,23 @@ pub fn parse_args(attrs: &[Attribute]) -> ClvmDeriveArgs {
...
-                        "curried_args" => {
-                            if let Some(existing) = repr {
-                                panic!("`curried_args` conflicts with `{}`", existing.to_string());
-                            }
-                            repr = Some(Repr::CurriedArgs);
-                        }
-                        ident => panic!("unknown argument `{}`", ident),
+                        "tuple" => repr = Some(Repr::Tuple),
+                        "list" => repr = Some(Repr::List),
+                        "curry" => repr = Some(Repr::Curry),
```
arvidn, line 50: it looks like you could factor out the `repr = Some(...)` from the match arms to make the whole match expression return the `Repr`

arvidn (approved):
just one nit if you thin it's worth addressing

arvidn (approved):

PR #317, opened 2023-11-17
Clvm-traits rewrite with Encoder

Rigidity:


clvm-traits/src/from_clvm.rs
```diff
@@ -60,113 +83,116 @@ clvm_primitive!(i128);
...
-impl<T, const N: usize> FromClvm for [T; N]
+impl<N, T, const LEN: usize> FromClvm<N> for [T; LEN]
 where
-    T: FromClvm,
+    T: FromClvm<N>,
 {
-    fn from_clvm(a: &Allocator, mut ptr: NodePtr) -> Result<Self> {
-        let mut items = Vec::with_capacity(N);
+    fn from_clvm(decoder: &impl ClvmDecoder<Node = N>, mut node: N) -> Result<Self, FromClvmError> {
+        let mut items = Vec::with_capacity(LEN);
```
arvidn, line 118: Not important.
I suspect that this would be more efficient and maybe more straight-forward code if this was allocated as an array on the stack, then filled in one item at a time with a `for`-loop.

I suppose leaving items uninitialized might add some complexity.

clvm-traits/src/match_byte.rs
```diff
@@ -1,36 +1,30 @@
...
+use crate::{ClvmDecoder, ClvmEncoder, FromClvm, FromClvmError, ToClvm, ToClvmError};
 
 #[derive(Debug, Copy, Clone)]
 pub struct MatchByte<const BYTE: u8>;
 
-impl<const BYTE: u8> ToClvm for MatchByte<BYTE> {
-    fn to_clvm(&self, a: &mut Allocator) -> Result<NodePtr> {
-        match BYTE {
-            0 => Ok(a.null()),
-            1 => Ok(a.one()),
```
arvidn, line 15: losing this is a little bit unfortunate as well. but we can always extend `ClvmEncoder` if it's a problem

chia-wallet/src/proof.rs
```diff
@@ -1,26 +1,25 @@
...
     Lineage(LineageProof),
     Eve(EveProof),
 }
 
-impl FromClvm for Proof {
-    fn from_clvm(a: &Allocator, node: NodePtr) -> Result<Self> {
-        LineageProof::from_clvm(a, node)
+impl<N> FromClvm<N> for Proof {
+    fn from_clvm(decoder: &impl ClvmDecoder<Node = N>, node: N) -> Result<Self, FromClvmError> {
+        LineageProof::from_clvm(decoder, decoder.clone_node(&node))
```
arvidn, line 12: is the point of the `clone_node()` call to avoid having to specify the trait bounds for `N` as `Clone`?

chia-wallet/src/puzzles/cat.rs
```diff
@@ -1,17 +1,16 @@
...
 
 use crate::LineageProof;
 
 #[derive(Debug, Clone, PartialEq, Eq, ToClvm, FromClvm)]
 #[clvm(curry)]
-pub struct CatArgs {
-    pub mod_hash: [u8; 32],
-    pub tail_program_hash: [u8; 32],
-    pub inner_puzzle: NodePtr,
+pub struct CatArgs<I> {
```
arvidn, line 10: nice

clvm-traits/src/clvm_encoder.rs
```diff
@@ -0,0 +1,34 @@
...
+    type Node: Clone;
+
+    fn encode_atom(&mut self, bytes: &[u8]) -> Result<Self::Node, ToClvmError>;
+    fn encode_pair(
+        &mut self,
+        first: Self::Node,
+        rest: Self::Node,
+    ) -> Result<Self::Node, ToClvmError>;
+
+    fn clone_node(&self, node: &Self::Node) -> Self::Node {
```
arvidn, line 15: this does look a bit unexpected. I think it warrants a comment

clvm-traits/src/to_clvm.rs
```diff
@@ -1,22 +1,50 @@
...
+}
+
 macro_rules! clvm_primitive {
     ($primitive:ty) => {
-        impl ToClvm for $primitive {
-            fn to_clvm(&self, a: &mut Allocator) -> Result<NodePtr> {
-                a.new_number((*self).into()).map_err(Error::Allocator)
+        impl<N> ToClvm<N> for $primitive {
+            fn to_clvm(&self, encoder: &mut impl ClvmEncoder<Node = N>) -> Result<N, ToClvmError> {
+                let number = BigInt::from(*self);
```
arvidn, line 46: since all primitives have fixed size, this could be made a lot more efficient by not involving `BigInt`

clvm-traits/src/wrappers.rs
```diff
@@ -0,0 +1,19 @@
+use crate::{ClvmDecoder, ClvmEncoder, FromClvm, FromClvmError, ToClvm, ToClvmError};
+
+/// A wrapper for an intermediate CLVM value. This is required to
+/// implement `ToClvm` and `FromClvm` for `N`, since the compiler
+/// cannot guarantee that the generic `N` type doesn't already
+/// implement these traits.
+pub struct Raw<N>(pub N);
```
arvidn, line 7: so, in order to convert to- and from `NodePtr`, we have to wrap in in a `Raw<NodePtr>()` now?

clvm-traits/src/from_clvm.rs
```diff
@@ -1,47 +1,70 @@
...
-use clvmr::{
-    allocator::{NodePtr, SExp},
-    op_utils::nullp,
-    Allocator,
-};
-use num_bigint::Sign;
+use crate::{ClvmDecoder, FromClvmError};
 
-use crate::{Error, Result};
+pub trait FromClvm<N>: Sized {
```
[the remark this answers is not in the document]
Rigidity: This prevents you from implementing it twice for different node types

clvm-traits/src/from_clvm.rs
```diff
@@ -1,47 +1,70 @@
...
+        Ok(node)
     }
 }
 
 macro_rules! clvm_primitive {
     ($primitive:ty) => {
-        impl FromClvm for $primitive {
-            fn from_clvm(a: &Allocator, ptr: NodePtr) -> Result<Self> {
-                if let SExp::Atom = a.sexp(ptr) {
-                    let (sign, mut vec) = a.number(ptr).to_bytes_be();
```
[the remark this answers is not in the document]
Rigidity: We can always add type-specific functions to the Encoder trait in the future

clvm-traits/src/from_clvm.rs
```diff
@@ -1,47 +1,70 @@
...
-                    Err(Error::ExpectedAtom(ptr))
+        impl<N> FromClvm<N> for $primitive {
+            fn from_clvm(
+                decoder: &impl ClvmDecoder<Node = N>,
+                node: N,
+            ) -> Result<Self, FromClvmError> {
+                const LEN: usize = std::mem::size_of::<$primitive>();
+
+                let bytes = decoder.decode_atom(&node)?;
+                let number = BigInt::from_signed_bytes_be(bytes);
```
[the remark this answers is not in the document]
Rigidity: Yeah I'll do an optimization pass later, I just don't want to rush it since it's a pretty critical bit of code - last time I touched it I ended up with numbers being encoded wrong in the wallet I'm building

clvm-traits/src/from_clvm.rs
Rigidity: Would this require unsafe code? As long as we're fine with introducing some unsafe into the repo, this does seem like a relevant optimization

clvm-traits/src/match_byte.rs
Rigidity: Yeah or the Allocator Encoder implementation can check these values and perform this optimization there

chia-wallet/src/proof.rs
Rigidity: Yep, unfortunate that the compiler can't infer this, since it's required by `type Node: Clone` in Encoder and Decoder anyway.

clvm-traits/src/to_clvm.rs
Rigidity: The issue is that unsigned integers may need an additional byte to represent that they're positive in two's complement

clvm-traits/src/wrappers.rs
Rigidity: Only when lazy-evaluating within generic ToClvm and FromClvm implementations. NodePtr explicitly implements ToClvm/FromClvm so you can rely on that for code using this library, but there can't be a blanket implementation for any type N.

Rigidity: So unless you're working with generic N node types, you can keep using NodePtr or whatever specific type you were before

clvm-traits/src/clvm_decoder.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use crate::FromClvmError;
+
+pub trait ClvmDecoder {
+    type Node: Clone;
+
+    fn decode_atom(&self, node: &Self::Node) -> Result<&[u8], FromClvmError>;
+    fn decode_pair(&self, node: &Self::Node) -> Result<(Self::Node, Self::Node), FromClvmError>;
+
+    fn clone_node(&self, node: &Self::Node) -> Self::Node {
```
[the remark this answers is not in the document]
Rigidity: I can't think of a situation where that would be required, but since we have `clone_node`, we can always do this in the future if needed.

clvm-traits/src/from_clvm.rs
arvidn: yeah, it would probably be something like this: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initializing-an-array-element-by-element

clvm-traits/src/to_clvm.rs
arvidn: the max size is still known at compile time though

arvidn (dismissed):

clvm-traits/src/match_byte.rs
arvidn: good point

arvidn (approved):

arvidn (approved):

arvidn:
perhaps `clvm` and `blspy` aren't installed by the workflow. I wonder why it used to work though

Rigidity:
Looks like the other platforms passed, maybe it'll work if rerun

PR #321, opened 2023-11-25
from_bytes_unchecked()

arvidn:
This extends the `Streamable` trait to support *unchecked* parsing. This is especially useful for parsing *trusted* BLS public keys and signatures. In unchecked mode the points are not verified to be on the G1 and G2 curves (because they are trusted to be).

## change

The main change is adding a new parameter to the `Streamable` trait's `parse()` function, indicating whether we want checked or unchecked parsing.

Additionally, a new convenience function is added `from_bytes_unchecked()`, as short-hand for just converting a buffer. This follows the naming convention we already use in `chia-blockchain`, and supersedes some existing python bindings for `from_bytes_unchecked()` on the BLS types.

## profile

The first commit adds a benchmark, highlighting how much time is saved when not checking the validity of points:

```
parse Signature         time:   [70.305 µs 70.518 µs 70.770 µs]
parse PublicKey         time:   [52.534 µs 52.647 µs 52.770 µs]
parse Signature (unchecked)
                        time:   [25.412 µs 25.458 µs 25.503 µs]
parse PublicKey (unchecked)
                        time:   [12.471 µs 12.491 µs 12.514 µs]
```

| type | checked | unchecked | unchecked / checked |
| --- | --- | --- | --- |
| Signature | 70.518 µs | 25.458 µs | 36.1% |
| PublicKey | 52.647 µs | 12.491 µs | 23.7% |

This can make a significant different when parsing *trusted* keys. We do this quite a lot, when reading from our database, or just when serializing/deserializing between processes in a python `ProcessPool`


chia-bls/src/gtelement.rs
```diff
@@ -57,12 +57,6 @@ impl GTElement {
     #[pyo3(name = "SIZE")]
     const PY_SIZE: usize = Self::SIZE;
 
-    #[staticmethod]
-    #[pyo3(name = "from_bytes_unchecked")]
-    fn py_from_bytes_unchecked(bytes: [u8; Self::SIZE]) -> Result<GTElement> {
```
arvidn, line 62: this function is now defined by the `PyStreamable` macro

chia-bls/src/public_key.rs
```diff
@@ -217,10 +211,13 @@ impl Streamable for PublicKey {
         Ok(())
     }
 
-    fn parse(input: &mut Cursor<&[u8]>) -> chia_traits::Result<Self> {
-        Ok(Self::from_bytes(
-            read_bytes(input, 48)?.try_into().unwrap(),
-        )?)
+    fn parse<const TRUSTED: bool>(input: &mut Cursor<&[u8]>) -> chia_traits::Result<Self> {
+        let input = read_bytes(input, 48)?.try_into().unwrap();
```
richardkiss, line 215: Why `unwrap` instead of `?`?

chia_py_streamable_macro/src/lib.rs
```diff
@@ -154,17 +154,34 @@ pub fn py_streamable_macro(input: proc_macro::TokenStream) -> proc_macro::TokenS
                 <Self as #crate_name::Streamable>::from_bytes(slice).map_err(|e| <#crate_name::chia_error::Error as Into<pyo3::PyErr>>::into(e))
             }
 
+            #[staticmethod]
+            #[pyo3(name = "from_bytes_unchecked")]
+            pub fn py_from_bytes_unchecked(blob: pyo3::buffer::PyBuffer<u8>) -> pyo3::PyResult<Self> {
+                if !blob.is_c_contiguous() {
+                    panic!("from_bytes_unchecked() must be called with a contiguous buffer");
```
richardkiss, line 161: Maybe `ValueError` is better here?

chia-bls/src/public_key.rs
arvidn: `read_bytes()` guarantees to read 48 bytes (or fail). The `try_into()` is guaranteed to succeed in turning the slice into `[u8; 48]`.

An alternative could be to make `read_bytes()` take a const generic as the number of bytes to read, and return an array directly.

chia_py_streamable_macro/src/lib.rs
arvidn: it would make me a bit uneasy to not have a hard failure. I think it's a problem, in `chia-blockchain`, that we swallow serious errors (sometimes assertion failures, and sometimes without logging) and keep going. I suppose a panic here also results in a python exception, but at least one that stands out a bit more and less likely to be swallowed.

arvidn: I'm actually unclear on which circumstances we would get a non-contiguous buffer from python. I imagine both `bytes` and `bytearray` are guaranteed to be contiguous, no?

chia-bls/src/public_key.rs
richardkiss: Maybe a new function `read_bytes_const`? Not necessary for now though.

chia_py_streamable_macro/src/lib.rs
richardkiss: Seems so.

Easiest way I've found to create non-contiguous is as follows:

```
>>> memoryview(b"foobar")[::2].contiguous
False
```

richardkiss (approved):

PR #325, opened 2023-11-29
CLVM derive support for enums

Rigidity:


clvm-derive/src/helpers.rs
```diff
@@ -20,20 +22,21 @@ impl fmt::Display for Repr {
     }
 }
 
-pub struct ClvmDeriveArgs {
-    pub repr: Repr,
+#[derive(Default)]
+pub struct ClvmAttr {
+    pub repr: Option<Repr>,
+    pub untagged: bool,
```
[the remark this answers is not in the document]
Rigidity: Meaning that the discriminator of the enum (the numeric value) is not used as a prefix when encoding. In other words, the enum is used as a wrapper that checks each of the variants in order instead of the first value first.

Maybe there's a better name?

clvm-traits/src/lib.rs
```diff
@@ -4,12 +4,16 @@
...
-//! for implementing the `ToClvm` and `FromClvm` traits on structs. They be marked
-//! with one of the following encodings:
+//! for implementing the `ToClvm` and `FromClvm` traits on structs and enums. They be
+//! marked with the following representations:
 //!
 //! * `#[clvm(tuple)]` for unterminated lists such as `(A . (B . C))`.
 //! * `#[clvm(list)]` for proper lists such as `(A B C)`, or in other words `(A . (B . (C . ())))`.
 //! * `#[clvm(curry)]` for curried arguments such as `(c (q . A) (c (q . B) (c (q . C) 1)))`.
+//!
+//! Additionally, you can use `#[clvm(untagged)]` on enums which don't have a numeric discriminant value.
```
arvidn, line 14: and `untagged` isn't allowed on non-enums, right?
I think you should describe what it means for an enum to not encode or match a discriminant.
You pick one matching enum value, is it the first matching from the top? The best matching? is it an error to have a value that matches more than one enum?

I think almost everything you put in the PR description belongs in the documentation on how to use the `clvm` derive macro.

clvm-traits/src/lib.rs
```diff
@@ -4,12 +4,16 @@
...
+//!
+//! The `#[repr(int_type)]` attribute can be used to specify the discriminant type for enums.
```
arvidn, line 16: it's not clear what this means. CLVM only has a single type, and presumably the discriminant would be encoded the same regardless of the representation you pick here, no?

clvm-traits/src/lib.rs
```diff
@@ -139,4 +143,64 @@ mod tests {
...
+        check(Enum::A(32), "ff8020");
+        check(Enum::B { x: -72 }, "ff0181b8");
+        check(Enum::C, "ff0280");
+    }
+
+    #[test]
+    fn test_explicit_enum() {
+        #[derive(Debug, ToClvm, FromClvm, PartialEq, Eq)]
+        #[clvm(tuple)]
+        #[repr(u8)]
```
arvidn, line 166: does this affect the result below?

clvm-derive/src/to_clvm.rs
```diff
@@ -59,18 +236,17 @@ pub fn to_clvm(mut ast: DeriveInput) -> TokenStream {
...
     ast.generics
         .params
-        .push(GenericParam::Type(TypeParam::from(node_name.clone())));
+        .push(GenericParam::Type(node_name.clone().into()));
     let (impl_generics, _, _) = ast.generics.split_for_impl();
 
     quote! {
         #[automatically_derived]
-        impl #impl_generics #crate_name::ToClvm<#node_name> for #struct_name #ty_generics #where_clause {
+        impl #impl_generics #crate_name::ToClvm<#node_name> for #item_name #ty_generics #where_clause {
```
arvidn, line 244: "item" is a pretty generic name. Is there a more concrete name that can be used here? maybe "type" or "identifier". Either way, it could probably use a comment to describe what an "item" really is.

clvm-derive/src/from_clvm.rs
```diff
@@ -1,66 +1,313 @@
...
-                let fields = &fields.named;
-                field_types = fields.iter().map(|field| field.ty.clone()).collect();
-                field_names = fields
-                    .iter()
-                    .map(|field| field.ident.clone().unwrap())
-                    .collect();
-                initializer = quote!(Self { #( #field_names, )* });
+        Data::Struct(data_struct) => {
+            if clvm_attr.untagged {
+                panic!("cannot use `untagged` on a struct");
```
arvidn, line 34: I think it would be great to have tests for these compilation failures. I believe it could be done using [compiletest-rs](https://github.com/Manishearth/compiletest-rs)

clvm-derive/src/to_clvm.rs
```diff
@@ -1,53 +1,230 @@
...
+    helpers::{add_trait_bounds, parse_clvm_attr, Repr},
+    macros::{repr_macros, Macros},
+};
 
-pub fn to_clvm(mut ast: DeriveInput) -> TokenStream {
-    let args = parse_args(&ast.attrs);
-    let crate_name = quote!(clvm_traits);
+#[derive(Default)]
+struct FieldInfo {
+    field_names: Vec<Ident>,
```
arvidn, line 15: as far as I can tell, the only think you use these names for are in destructure expressions. The only important property is that they are unique, right? In theory you could just name them all `field_1`, `field_2`, etc.

Does it improve error messages to use the actual field names (when you have them)?

If not, it might simplify the code somewhat, if you don't have to extract the names, but you can just generate them on the fly where you need them

clvm-derive/src/to_clvm.rs
```diff
@@ -1,53 +1,230 @@
...
+
+                if variant_clvm_attr.untagged {
+                    panic!("cannot use `untagged` on an enum variant");
+                }
+
+                let repr = variant_clvm_attr
+                    .repr
+                    .unwrap_or_else(|| clvm_attr.expect_repr());
+                if !clvm_attr.untagged && repr == Repr::Curry {
+                    panic!("cannot use `curry` on a tagged enum variant");
```
arvidn, line 60: why is this combination of attributes disallowed? (maybe the error message should say)

clvm-derive/src/to_clvm.rs
```diff
@@ -1,53 +1,230 @@
...
+
+                let macros = repr_macros(&crate_name, repr);
+                let variant_info = VariantInfo {
+                    name: variant.ident.clone(),
+                    value: variant
+                        .discriminant
+                        .as_ref()
+                        .map(|(_, value)| {
+                            if clvm_attr.untagged {
+                                panic!("cannot use `untagged` on an enum with discriminants");
```
arvidn, line 71: this error message is not clear to me. Is this the case where the enum has at least one "plain" or C-style value?

clvm-derive/src/to_clvm.rs
```diff
@@ -1,53 +1,230 @@
...
+            let macros = repr_macros(&crate_name, clvm_attr.expect_repr());
+            let field_info = fields(&data_struct.fields);
+            impl_for_struct(&crate_name, &ast, &macros, &field_info)
+        }
+        Data::Enum(data_enum) => {
+            if !clvm_attr.untagged && clvm_attr.repr == Some(Repr::Curry) {
+                panic!("cannot use `curry` on a tagged enum");
+            }
+
+            let mut next_value: Expr = parse_quote!(0);
```
arvidn, line 45: is this the "tag" you will use to convert the enum into CLVM?
I associate the term "tag" with "tagged union", later "variant" in the C++ world. This is of course essentially the same as a rust enum. Is the term "tag" used in rust as well? I get the impression the term "discriminant" is used as the underlying value representing which enum variant is active.

If so, it seems like a better name for this variable would be `next_discriminant` or maybe `next_tag`. But perhaps best would be to have a comment describing what this is used for.

clvm-derive/src/to_clvm.rs
```diff
@@ -1,53 +1,230 @@
...
+                                panic!("cannot use `untagged` on an enum with discriminants");
+                            }
+
+                            next_value = parse_quote!(#value + 1);
+                            value.clone()
+                        })
+                        .unwrap_or_else(|| {
+                            let value = next_value.clone();
+                            next_value = parse_quote!(#next_value + 1);
+                            value
```
arvidn, line 80: what's going on here? (a comment in the code might be warranted). Presumably if this call fails, it does so before incrementing `next_value`, so it's never done twice. However, it's not obvious in what case we *would* fail here, and why we need to recover from that failure.

clvm-derive/src/to_clvm.rs
```diff
@@ -1,53 +1,230 @@
...
+                    name: variant.ident.clone(),
+                    value: variant
+                        .discriminant
+                        .as_ref()
+                        .map(|(_, value)| {
+                            if clvm_attr.untagged {
+                                panic!("cannot use `untagged` on an enum with discriminants");
+                            }
+
+                            next_value = parse_quote!(#value + 1);
```
arvidn, line 74: if I understand this correctly, you compute the discrimintnat/tag yourself here, by just incrementing a counter from 0. Is this better or more reliable than using `mem::discriminant()`?

arvidn (commented):
I'm not all the way through this review yet. But here's what I've got so far.
Generally I think it would be good with more comments explaining what's going on. Macro code is especially unintuitive sometimes.

I appreciate the refactoring of `fields()` and `repr_macros()` to handle the different kinds of types in a uniform way (although I think they could both use comments describing what they do).

As for the named "untagged", maybe it would make sense to give it a name that describes the behavior instead. i.e. that it will perform a match instead. I don't know, untagged isn't so bad either (as along as it's well documented).

The exact behavior would ideally also be demonstrated in a test. regarding match-first, match-best or ambiguity being an error.

clvm-derive/src/to_clvm.rs
Rigidity: `mem::discriminant` returns an opaque type instead of the raw integer value, which I need for serialization.

clvm-derive/src/to_clvm.rs
Rigidity: Explained it in the error - Unlike other representations, each argument in a curried parameter list is wrapped so it doesn't work with the two step parsing method I'm using.

In the future, I will write utilities to simplify parsing list/tuple/curry/similar structures, and remove this limitation.

clvm-derive/src/to_clvm.rs
Rigidity: Improved the name

clvm-derive/src/to_clvm.rs
Rigidity: `variant.discriminant` is an Option since you're free to specify an explicit discriminant on any of the fields. If specified, it's used as the current value. Otherwise, the next incremented value is used as the default, per the Rust enum docs.

clvm-derive/src/to_clvm.rs
Rigidity: It's used by the initializer for destructuring like you mentioned:
```rs
let initializer = quote!({ #( #field_names, )* });
```

You need to know the names of the parameters to destructure them. It's possible to alias it to `field_1` but there's no need that I can think of. And it should improve error messages not to, yeah

clvm-traits/src/lib.rs
Rigidity: The docs do need to be improved though.

clvm-derive/src/to_clvm.rs
Rigidity: Yes, but this is for when there are named fields. For example:

```rust
struct X {
  a: i32,
  b: i32,
}

enum X {
  Y {
    a: i32,
    b: i32,
  }
}
```

You cannot construct `Self` or `Self::Y` in these cases without using the proper names of the fields.

Rigidity: When the fields are unnamed, an indexed name is used instead

clvm-derive/src/from_clvm.rs
```diff
@@ -1,66 +1,309 @@
...
+                    discriminant: variant
+                        .discriminant
+                        .as_ref()
+                        .map(|(_, discriminant)| {
+                            next_discriminant = parse_quote!(#discriminant + 1);
+                            discriminant.clone()
+                        })
+                        .unwrap_or_else(|| {
+                            let discriminant = next_discriminant.clone();
+                            next_discriminant = parse_quote!(#next_discriminant + 1);
```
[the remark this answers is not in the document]
Rigidity: You can use constants instead of integer literals for the discriminant value, and I don't think constant values are known at the time this macro is running. This all gets optimized away at compile time though.

clvm-derive/src/from_clvm.rs
```diff
@@ -1,66 +1,309 @@
...
+            if clvm_attr.untagged {
+                impl_for_untagged_enum(&crate_name, &ast, &variants)
+            } else {
+                let int_repr = parse_int_repr(&ast.attrs);
+                impl_for_enum(&crate_name, &ast, &int_repr, &variants)
             }
-            Fields::Unit => panic!("unit structs are not supported"),
-        },
+        }
         _ => panic!("expected struct with named or unnamed fields"),
```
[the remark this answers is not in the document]
Rigidity: Forgot to update this

clvm-derive/src/from_clvm.rs
```diff
@@ -1,66 +1,309 @@
...
+
+    generate_from_clvm(crate_name, ast, &node_name, &body)
+}
+
+fn impl_for_untagged_enum(
+    crate_name: &TokenStream,
+    ast: &DeriveInput,
+    variants: &[VariantInfo],
+) -> TokenStream {
     let node_name = Ident::new("Node", Span::mixed_site());
```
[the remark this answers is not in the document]
Rigidity: The generic parameter is both defined and used within the scope of the implementation. Though there is a chance of it clashing with an existing generic parameter with the same name. Perhaps prefixing the user-provided ones with an underscore or something could solve that?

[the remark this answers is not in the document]
Rigidity: Proc macro hygiene is a bit hard to get right

clvm-traits/src/lib.rs
```diff
@@ -139,4 +118,98 @@ mod tests {
...
+
+        check(Enum::A(32), "ff8020");
+        check(Enum::B { x: -72 }, "ff0181b8");
+        check(Enum::C, "ff0280");
+    }
+
+    #[test]
+    fn test_explicit_enum() {
+        #[derive(Debug, ToClvm, FromClvm, PartialEq, Eq)]
+        #[clvm(tuple)]
```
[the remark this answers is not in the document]
Rigidity: You need a representation for the enum variants' fields.

For example for the variant, A:
- The tuple representation is `(42 . <first argument>)`
- The list representation is `(42 . (<first argument> . ()))`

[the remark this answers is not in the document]
Rigidity: It doesn't particularly matter which one is chosen for this test, but it does change the output.

clvm-traits/src/lib.rs
```diff
@@ -139,4 +118,98 @@ mod tests {
 
         check(NewTypeStruct("XYZ".to_string()), "8358595a");
     }
+
+    #[test]
+    fn test_enum() {
+        #[derive(Debug, ToClvm, FromClvm, PartialEq, Eq)]
+        #[clvm(tuple)]
```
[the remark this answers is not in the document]
Rigidity: It does for the same reason as above, the variants are encoded as if the discriminant was just a regular field. So if it's set to `list` instead, the encoding would be null terminated. Doesn't matter for this test, but it does affect the result.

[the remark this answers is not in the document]
Rigidity: Applying `tuple` to the enum itself is the same as applying it to each variant individually.

clvm-traits/docs/derive_macros.md
```diff
@@ -0,0 +1,218 @@
...
+It's possible to override both the type of the discriminator, and the value.
+The `#[repr(...)]` attribute is used by the Rust compiler to allow overriding the discriminator type.
+As such, this attribute is also used to change the underlying type used to serialize and deserialize discriminator values.
+
+```rust
+use clvmr::Allocator;
+use clvm_traits::{ToClvm, FromClvm};
+
+#[derive(Debug, PartialEq, Eq, ToClvm, FromClvm)]
+#[clvm(tuple)]
```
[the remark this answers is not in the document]
Rigidity: Yes, you must specify one of the 3 representations when deriving the macros. There isn't a default, since it wouldn't necessarily be clear which one would be the default.

clvm-traits/docs/derive_macros.md
```diff
@@ -0,0 +1,218 @@
...
+};
+
+let a = &mut Allocator::new();
+let ptr = mode.to_clvm(a).unwrap();
+assert_eq!(SpendMode::from_clvm(a, ptr).unwrap(), mode);
+```
+
+### Untagged Enums
+
+Often, the discriminator isn't necessary to encode, and you'd prefer to match try variant in order until one matches.
```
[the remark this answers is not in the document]
Rigidity: Yes

clvm-traits/src/lib.rs
```diff
@@ -139,4 +118,98 @@ mod tests {
...
+        // This round trips to `A` instead of `C`, since the bytes used to represent
+        // this string are also a valid `i32` value.
+        assert_eq!(
+            coerce_into::<Enum, Enum>(Enum::C("Hi".into())),
+            Enum::A(18537)
+        );
+
+        // This round trips to `C` instead of `A`, since the bytes used to represent
+        // this string exceed the size of `i32`.
+        assert_eq!(
```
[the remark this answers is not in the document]
Rigidity: Might be confusing for this example, but more generally if A fails to parse the input, it will try B instead, and then C.

clvm-derive/src/from_clvm.rs
```diff
@@ -1,66 +1,309 @@
...
+                        Ok(Self::#name #initializer)
+                    }
+                }
+            }
+        })
+        .collect::<Vec<_>>();
+
+    let parse_value = if has_initializers {
+        quote! {
+            let (value, args) = <(#int_repr, #crate_name::Raw<#node_name>)>::from_clvm(decoder, node)?;
```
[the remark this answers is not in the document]
Rigidity: Both of these variables aren't used after the enum fields are brought into scope. I just tried it and it works fine currently, it just shadows the old variable and since the old one isn't used anywhere in that scope nothing breaks.

arvidn (approved):

PR #327, opened 2023-11-30
Cmj.wasm test

ChiaMineJP:
Experimental PR to make `chia-protocol` available as a wasm module.
Current status:

---

### Building wasm module only with `struct` / `enum` succeeded
Steps to reproduce:
1. Make sure you have `wasm-pack` installed on your system. (Just invoke `cargo add wasm-pack` to install)
2. `cd chia-protocol`
3. `wasm-pack build`

---

### To convert methods with fixed sized byte array arguments/return value like `&[u8; 4]`, we need to wait `wasm-bindgen`' update
https://github.com/rustwasm/wasm-bindgen/pull/3458

---

### `u128` not supported yet by `wasm-bindgen`
`wasm-bindgen` does not yet support `u128` so only when compiling for `wasm32` the type will be converted to `u64` temporarily by `#[wasm_patch::conv_u128_to_u64_for_wasm]` proc macro attritute.

---

### BytesImpl<T> should not be used outside `chia-protocol/src/bytes.rs`
`BytesImpl<T>` is basically aliased to types like `Bytes32`, `Bytes48`, and so on.
Since `wasm-bindgen` cannot treat generic parameters well when it is converting rust code to wasm, please do not use `BytesImpl<T>` notation outside `bytes.rs`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -489,7 +489,6 @@ name = "chia_wasm"
 version = "0.2.7"
 dependencies = [
  "chia",
- "chia-protocol",
  "js-sys",
  "wasm-bindgen",
  "wasm-bindgen-test",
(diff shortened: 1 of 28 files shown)


Rigidity:
Is it fine to close this? This is quite out of date

ChiaMineJP:
@Rigidity Yes, this's too old. Closing.

PR #332, opened 2023-12-04
drop clvm-trait's dependency on clvmr

arvidn:
remove the `ToNodePtr` and `FromNodePtr` traits as well as `FromClvm` and `ToClvm` for `Allocator` and `NodePtr` (those will be moved into `clvmr`).

This is one step in untangling the dependency cycle caused by making `clvmr` use `chia-bls`.

In order to keep everything building, all other crates that depend on `clvm-traits` now use the crates.io version (i.e. stay on the current version).

once this lands, `clvm_rs` can be updated to implement the `clvm-traits`. Then this repository can be restored to depend on the new `clvm_rs` and `clvm-traits`

### DIFF
--- a/chia-tools/Cargo.toml
+++ b/chia-tools/Cargo.toml
@@ -12,7 +12,7 @@ repository = "https://github.com/Chia-Network/chia_rs/chia-tools"
 chia-protocol = { version = "0.2.14", path = "../chia-protocol" }
 chia-traits = { path = "../chia-traits" }
 clvm-utils = { path = "../clvm-utils" }
-clvm-traits = { path = "../clvm-traits" }
+clvm-traits = { version = "=0.2.14" }
 chia-wallet = { version = "=0.2.14", path = "../chia-wallet" }
 clvmr = { version = "=0.3.0", features = ["counters"] }
 chia = { version = "0.2.14", path = ".." }
--- a/clvm-traits/docs/derive_macros.md
+++ b/clvm-traits/docs/derive_macros.md
@@ -24,7 +24,7 @@ For example:
 - `(A, B, C)` is encoded as `(A . (B . C))`, since every cons-pair must contain two values.
 - `(A, B, C, D)` is encoded as `(A . (B . (C . D)))` for the same reason as above.
 
-```rust
+```rust compile_fail
 use clvmr::Allocator;
 use clvm_traits::{ToClvm, FromClvm};
 
@@ -59,7 +59,7 @@ For example:
 Note that the following code is for example purposes only and is not indicative of how to create a secure program.
 Using a password like shown in this example is an insecure method of locking coins, but it's effective for learning.
 
-```rust
+```rust compile_fail
 use clvmr::Allocator;
 use clvm_traits::{ToClvm, FromClvm};
 
@@ -91,7 +91,7 @@ You can read more about currying on the [Chia blockchain documentation](https://
 Note that the following code is for example purposes only and is not indicative of how to create a secure program.
 Using a password like shown in this example is an insecure method of locking coins, but it's effective for learning.
 
-```rust
+```rust compile_fail
 use clvmr::Allocator;
 use clvm_traits::{ToClvm, FromClvm};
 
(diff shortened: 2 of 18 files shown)


chia-tools/Cargo.toml
```diff
@@ -12,7 +12,7 @@ repository = "https://github.com/Chia-Network/chia_rs/chia-tools"
 chia-protocol = { version = "0.2.14", path = "../chia-protocol" }
 chia-traits = { path = "../chia-traits" }
 clvm-utils = { path = "../clvm-utils" }
-clvm-traits = { path = "../clvm-traits" }
+clvm-traits = { version = "=0.2.14" }
```
Rigidity, line 15: nit: this can just be `clvm-traits = "=0.2.14"`. same for chia-wallet, chia-wallet/fuzz, clvm-utils, clvm-utils/fuzz, fuzz, and anywhere else

clvm-traits/docs/derive_macros.md
```diff
@@ -24,7 +24,7 @@ For example:
 - `(A, B, C)` is encoded as `(A . (B . C))`, since every cons-pair must contain two values.
 - `(A, B, C, D)` is encoded as `(A . (B . (C . D)))` for the same reason as above.
 
-```rust
+```rust compile_fail
```
Rigidity, line 27: Why does this need to be compile_fail?

clvm-traits/src/lib.rs
```diff
@@ -27,37 +27,157 @@ pub use to_clvm::*;
...
+        match input {
+            TestNode::Atom(v) => {
+                let atom = a.atom(*v);
+                if atom.len() == 0 {
+                    "NIL".to_owned()
+                } else {
+                    format!("{}", &hex::encode(&atom))
+                }
+            }
+            TestNode::Pair(l, r) => format!("( {} {}", node_to_str(a, l), node_to_str(a, r)),
```
Rigidity, line 81: This format might be easier to read with a closing brace for cons pairs, something like ({} {})

clvm-traits/src/lib.rs
```diff
@@ -27,37 +27,157 @@ pub use to_clvm::*;
...
-mod tests {
+pub mod tests {
     extern crate self as clvm_traits;
 
-    use std::fmt;
+    use super::*;
+
+    #[derive(Clone)]
+    pub enum TestNode {
+        Atom(usize),
```
Rigidity, line 37: This won't work for tests that have atoms that exceed this, such as `i128::MAX`. Maybe either `BigInt` or `Vec<u8>` is more appropriate here?

Rigidity (commented):
Can't approve yet, but here's some comments

clvm-traits/docs/derive_macros.md
arvidn: because it's assuming that `Allocator` and `NodePtr` implements `FromClvm` and `ToClvm`. Once all of this is sorted, I expect to restore this.

doctest tries to build it, and it fails because it's using this local version of clvm-traits

clvm-traits/src/lib.rs
arvidn: yeah, but it's more complicated to write a parser for :)

chia-tools/Cargo.toml
Rigidity: true

clvm-traits/src/lib.rs
arvidn: my first attempt (trying to keep this as simple as possible) was to store the atom data in the `Node` itself. However, this is not possible with our current interfaces/traits. When we ask an allocator for the bytes, it has to return a slice whose lifetime is tied to the allocator. You can't make it return a slice whose lifetime is tied to the node itself (because that violates the trait).

So, this `usize` is an index into the `atoms` member of the `TestAllocator`

Rigidity (approved):

clvm-traits/src/lib.rs
Rigidity: Ok, we can probably improve the test API later once this is all sorted out

PR #333, opened 2023-12-04
remove LIMIT_OBJECTS flag and disable it unconditionally

arvidn:
Now that soft-fork3 has activated, this feature has been enabled unconditionally in `chia-blockchain` ([here](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/full_node/mempool_check_conditions.py#L58)). We no longer need it, we just enable it unconditionally.


src/gen/test_generators.rs
```diff
@@ -223,6 +223,7 @@ fn run_generator(#[case] name: &str) {
             Err(code) => (0, format!("FAILED: {}\n", u32::from(code.1))),
         };
 
+        let mut a = make_allocator(*flags);
```
arvidn, line 226: this was an oversight earlier. not creating a new allocator when running the test again meant we just doubled the number of allocations. With this new limit, we exceed the limit the second time we run it, causing a discrepancy. We expect the output to be the same. This fixes it by starting both runs with fresh allocators.

tests/run-programs.py
```diff
@@ -54,7 +54,7 @@
     output += completed_process.stdout.decode('UTF-8')
     end = time.perf_counter()
 
-    if 'FAIL: ' not in output or 'cost exceeded' not in output:
+    if 'FAIL: ' not in output or ('cost exceeded' not in output and 'too many pairs' not in output):
```
arvidn, line 57: some of these stress-tests exceed the allocation limit now, and they can fail in two waus

PR #337, opened 2023-12-07
add tool to print chia-blockchain event log

arvidn:
this parses the log produced by `chia-blockchain`, with this PR: https://github.com/Chia-Network/chia-blockchain/pull/16982


Rigidity:
Is this something we should still add?

PR #343, opened 2023-12-11
Untangle dependencies

Rigidity:


Cargo.toml
```diff
@@ -60,3 +60,6 @@ crate-type = ["rlib"]
 
 [profile.release]
 lto = "thin"
+
+[patch.crates-io]
+clvm-traits = { path = "clvm-traits" }
```
arvidn, line 65: I think a comment here is warranted. at least something brief about how this is meant to be temporary. It is, right?

arvidn (commented):
looks good

arvidn (approved):

PR #344, opened 2023-12-12
relax coin amount in fast-forward logic

arvidn:
Update the fast-forward feature to allow the coin amount to change when "rebasing" a spend onto a later version of the same singleton.


AmineKhaldi:
I see two remaining tasks with this patch:

1. We need to account for (and cover with tests) the fact that not only the latest version's amount can change, but its parent's amount can change too.
2. We need to relax (and cover with tests) the check for output coins as it still requires the same amount.

src/gen/conditions.rs
```diff
@@ -4311,6 +4312,7 @@ fn test_eligible_for_ff_even_amount() {
 #[cfg(test)]
 #[rstest]
 #[case(123, "{h2}", true)]
+#[case(121, "{h2}", true)]
```
arvidn, line 4315: the same puzzle-hash a different amount (but still odd) makes for an eligible FF spend

Rigidity (approved):
lgtm

PR #351, opened 2023-12-19
build with clang-cl on windows to get proper portable SHA256

emlowe:
Using latest BLST building with clang to workaround AMD K10 issues with portable


wjblanke (dismissed):
aok

arvidn:
this should still land in `main` right?

arvidn (approved):
I think this should be squashed when merged

PR #352, opened 2023-12-19
Switch to dtolnay/rust-toolchain

Rigidity:
[https://github.com/actions-rs/toolchain](actions-rs/toolchain) is archived and unmaintained, so we should use [https://github.com/dtolnay/rust-toolchain](dtolnay/rust-toolchain) instead.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -140,9 +140,7 @@ jobs:
           python -m pip install --upgrade pip
 
     - name: Set up rust
-      uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
+      uses: dtolnay/rust-toolchain@stable
 
     - name: Install dependencies
       run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test.yml
```diff
@@ -298,11 +297,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@nightly
         with:
-          toolchain: nightly
           components: clippy
-          override: true
+        run: rustup override set nightly
```
arvidn, line 303: is this necessary given you use `@nightly`?

.github/workflows/build-test.yml
```diff
@@ -282,12 +280,13 @@ jobs:
...
           fetch-depth: 1
+
       - name: Install rust
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@nightly
         with:
-            toolchain: nightly
             components: rustfmt, clippy
-            override: true
+        run: rustup override set nightly
```
arvidn, line 288: and here

.github/workflows/build-test.yml
Rigidity: This is what the behavior of the old workflow was (`override: true`), didn't check if it's necessary. I believe it essentially just overrides the `toolchain.toml` file, though I'm not sure we have one of those. I can try removing it?

arvidn: I see

arvidn (approved):

PR #359, opened 2023-12-28
add BlockRecord

arvidn:

This adds the `BlockRecord` type, defined in python [here](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/consensus/block_record.py#L46).

The main challenge in this type is implementing `sp_sub_slot_total_iters()`, `ip_sub_slot_total_iters()`, `sp_iters()`, `ip_iters()`, `sp_total_iters()`. The underlying functions used by these are implemented in python.

This patch calls back into python to run these functions. It's not elegant or efficient, but it's a stop-gap until we have ported these functions to rust.

In order to ensure this works, a fuzzer constructs an arbitrary `BlockRecord` object and calls all these functions on both the rust and python version, ensuring the results are identical, including failures.


Rigidity (approved):
Looks good for now, would definitely like to remove any calls back into Python at some point

PR #364, opened 2024-01-09
Fix lineage proof byte representation

Rigidity:


chia-wallet/src/proof.rs
```diff
@@ -29,13 +30,13 @@ impl<'a> Arbitrary<'a> for Proof {
...
-                parent_coin_info: u.arbitrary()?,
+                parent_coin_info: (u.arbitrary::<[u8; 32]>()?).into(),
                 amount: u.arbitrary()?,
             }))
         } else {
             Ok(Self::Lineage(LineageProof {
-                parent_coin_info: u.arbitrary()?,
-                inner_puzzle_hash: u.arbitrary()?,
+                parent_coin_info: (u.arbitrary::<[u8; 32]>()?).into(),
+                inner_puzzle_hash: (u.arbitrary::<[u8; 32]>()?).into(),
```
arvidn, line 39: does `Bytes32` not implement `Arbitrary`? (I think it should)

arvidn (commented):
looks good, but I think it would look even better if `Bytes32` implemented `Arbitrary`

chia-wallet/src/proof.rs
arvidn: `Bytes32` *does* implement `Arbitrary`, but only when the `fuzzing` feature is enabled.
I imagine you really only need the `Arbitrary` implementation when fuzzing as well

arvidn (approved):
🎉

PR #365, opened 2024-01-10
Add methods to peer and make message ids atomic

Rigidity:
This adds various helper methods to `Peer` to simplify the process of sending requests, provide better type safety, and handle rejections as part of the error variant of the response.

Message ids beyond `u16::MAX` would also previously panic in debug mode, but now a thread-safe `AtomicU16` is used instead of `Mutex<u16>`, which also means it will wrap on overflow instead of panicking.

### DIFF
--- a/chia-client/src/peer.rs
+++ b/chia-client/src/peer.rs
@@ -86,7 +86,7 @@ impl Peer {
             coin_name: coin_id,
             height,
         };
-        let response: RespondPuzzleSolution = self.request(body).await?;
+        let response: RespondPuzzleSolution = self.request_or_reject(body).await?;
         Ok(response.response)
     }
 
@@ -97,7 +97,7 @@ impl Peer {
         let body = SendTransaction {
             transaction: spend_bundle,
         };
-        self.request_infallible(body).await
+        self.request(body).await
     }
 
     pub async fn request_block_header(
@@ -105,7 +105,7 @@ impl Peer {
         height: u32,
     ) -> Result<HeaderBlock, Error<RejectHeaderRequest>> {
         let body = RequestBlockHeader { height };
-        let response: RespondBlockHeader = self.request(body).await?;
+        let response: RespondBlockHeader = self.request_or_reject(body).await?;
         Ok(response.header_block)
     }
 
@@ -121,7 +121,7 @@ impl Peer {
             return_filter,
         };
         let response: RespondBlockHeaders =
-            self.request(body)
+            self.request_or_reject(body)
                 .await
                 .map_err(|error: Error<RejectBlockHeaders>| match error {
                     Error::Rejection(_rejection) => Error::Rejection(()),
@@ -144,7 +144,7 @@ impl Peer {
             header_hash,
             coin_names: coin_ids,
         };
-        self.request(body).await
+        self.request_or_reject(body).await
     }
 
     pub async fn request_additions(
@@ -158,7 +158,7 @@ impl Peer {
             header_hash,
             puzzle_hashes,
         };
-        self.request(body).await
+        self.request_or_reject(body).await
     }
 
     pub async fn register_for_ph_updates(
@@ -170,7 +170,7 @@ impl Peer {
             puzzle_hashes,
             min_height,
         };
-        let response: RespondToPhUpdates = self.request_infallible(body).await?;
+        let response: RespondToPhUpdates = self.request(body).await?;
         Ok(response.coin_states)
     }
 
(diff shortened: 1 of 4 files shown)


chia-client/src/peer.rs
```diff
@@ -127,18 +225,46 @@ impl Peer {
...
+            Response::from_bytes(data).or(Err(Error::InvalidResponse(message)))
+        } else if message.msg_type == Rejection::msg_type() {
+            let rejection = Rejection::from_bytes(data).or(Err(Error::InvalidResponse(message)))?;
+            Err(Error::Rejection(rejection))
+        } else {
+            Err(Error::InvalidResponse(message))
+        }
+    }
+
+    pub async fn request_infallible<Response, T>(&self, body: T) -> Result<Response, ()>
```
arvidn, line 247: what does "infallible" refer to here? the function can still fail, right?
or does it refer to the fact that being `async` makes it really just return a coroutine, and it never fails to return a coroutine?

chia-client/src/peer.rs
```diff
@@ -127,18 +225,46 @@ impl Peer {
...
+    where
+        Response: Streamable + ChiaProtocolMessage,
+        T: Streamable + ChiaProtocolMessage,
+    {
+        let message = self.request_raw(body).await?;
```
arvidn, line 252: in this case, the error returned by `?`, is that part of the co-routine as well?

chia-client/src/peer.rs
```diff
@@ -106,10 +70,144 @@ impl Peer {
...
+        let response: RespondBlockHeaders =
+            self.request(body)
+                .await
+                .map_err(|error: Error<RejectBlockHeaders>| match error {
+                    Error::Rejection(_rejection) => Error::Rejection(()),
+                    Error::Chia(error) => Error::Chia(error),
+                    Error::WebSocket(error) => Error::WebSocket(error),
+                    Error::InvalidResponse(error) => Error::InvalidResponse(error),
+                    Error::MissingResponse => Error::MissingResponse,
+                })?;
```
arvidn, line 125: I'm a bit confused by this. what happens to these error codes? the function is declared to return `()` as the error.

chia-client/src/peer.rs
```diff
@@ -194,6 +309,40 @@ impl Peer {
...
+                            .send(PeerEvent::$event($event::from_bytes(message.data.as_ref())?))
+                            .ok();
+                    } )+
+                    _ => {}
+                }
+            };
+        }
+
+        events!(CoinStateUpdate, NewPeakWallet);
+
```
arvidn, line 343: we don't seem to react to unknown messages being sent. I suppose we'd want to do that at some point, before production use

chia-client/src/peer.rs
```diff
@@ -194,6 +309,40 @@ impl Peer {
     pub fn receiver_mut(&mut self) -> &mut broadcast::Receiver<PeerEvent> {
         &mut self.event_receiver
     }
+
+    async fn handle_inbound(
```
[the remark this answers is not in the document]
Rigidity: No

chia-client/src/peer.rs
```diff
@@ -106,10 +70,144 @@ impl Peer {
...
+    pub async fn request_puzzle_and_solution(
+        &self,
+        coin_id: Bytes32,
+        height: u32,
+    ) -> Result<PuzzleSolutionResponse, RejectPuzzleSolution> {
+        let body = RequestPuzzleSolution {
+            coin_name: coin_id,
+            height,
+        };
+        let response: RespondPuzzleSolution = self.request(body).await?;
```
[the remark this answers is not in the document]
Rigidity: You would get an `Err(Error::InvalidResponse)`

chia-client/src/peer.rs
Rigidity: Perhaps there's a better name. What I mean is there is no rejection response message for this request. It can still fail in other ways.

chia-client/src/peer.rs
Rigidity: Not sure what you mean, if there is no response message because something fails, it will return the error early.

chia-client/src/peer.rs
Rigidity: I think it may be confusing to read since `Result<ResponseMessage, RejectionMessage>` is actually a custom type, shadowing the normal `std::result::Result<T, ErrorType>`. The `()` just means than the error variant is `Error<()>`, meaning no rejection message but still other types of errors can be returned.

Would it be better if I rename `Result` to `Response` or something like that?

arvidn: oh, the `Result<>` type here is not `std::result::Result`? I guess that's where my confusion came from.

Rigidity: Fixed

chia-client/src/peer.rs
Rigidity: Added a comment

chia-client/src/peer.rs
```diff
@@ -127,18 +225,46 @@ impl Peer {
...
-        R: Streamable + ChiaProtocolMessage,
     {
         // Get the current nonce and increment.
-        let message_id = {
-            let mut nonce = self.nonce.lock().await;
-            let id = *nonce;
-            *nonce += 1;
-            id
-        };
+        let message_id = self.nonce.fetch_add(1, Ordering::SeqCst);
```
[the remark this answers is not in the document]
Rigidity: Added a comment

arvidn (approved):

PR #366, opened 2024-01-10
Add Default implementation to empty streamables

Rigidity:


chia-protocol/src/message_struct.rs
```diff
@@ -1,45 +1,56 @@
...
+        $crate::streamable_struct!( impl $name );
+
+        impl Default for $name {
+            fn default() -> Self {
+                Self
+            }
+        }
+    };
+
+    ( impl $name:ident $( { $( $field:ident: $t:ty ),* $(,)? } )? ) => {
```
arvidn, line 40: it there a convention to use the word "impl" to essentially split a macro into two modes?
my instinct would have been to have a separate macro for, maybe called `streamable_struct_impl` or something

arvidn (commented):
looks reasonable. just one question

arvidn (approved):

PR #371, opened 2024-01-17
Re-export crates from the chia_rs meta-crate

Rigidity:
This turns the main `chia` crate into a meta-crate, allowing it to be installed instead of each of the libraries in this mono-repo for convenience.


arvidn (commented):
it's not clear to me that this is a solid step forward. it seems like it kind of makes this crate a bit messy. here are a few observations:

* chia-bls is not an optional dependency and can't be disabled by a feature
* rather than enabling the "meta crate" features by features, it seems you have to disabled them with `no-default-features`, which alters the existing crates that depend on this

If the end goal is to make the root project a pure meta crate (without features required to enable dependencies), it's not obvious that this is a step in that direction

Rigidity:
This can be revisited at a later date.

PR #372, opened 2024-01-22
Update trait impls and conversions for byte types

Rigidity:
This PR adds missing conversions for `Bytes`, `BytesImpl`, and `Program`, and removes `PartialEq` implementations for other types to reduce complexity.

It also changes `From` to `TryFrom` in some cases where the conversion is fallible:
> **Note: This trait must not fail.** The `From` trait is intended for perfect conversions. If the conversion can fail or is not perfect, use `TryFrom`.

https://doc.rust-lang.org/std/convert/trait.From.html

### DIFF
--- a/chia-protocol/src/bytes.rs
+++ b/chia-protocol/src/bytes.rs
@@ -163,7 +163,7 @@ impl<const N: usize> BytesImpl<N> {
         &self.0
     }
 
-    pub fn into_inner(self) -> [u8; N] {
+    pub fn to_bytes(self) -> [u8; N] {
         self.0
     }
 
--- a/chia-protocol/src/bytes.rs
+++ b/chia-protocol/src/bytes.rs
@@ -1,12 +1,9 @@
-use chia_traits::chia_error;
-use chia_traits::{read_bytes, Streamable};
+use chia_traits::{chia_error, read_bytes, Streamable};
 use clvm_traits::{ClvmDecoder, ClvmEncoder, FromClvm, FromClvmError, ToClvm, ToClvmError};
-use core::fmt::Formatter;
 use sha2::{Digest, Sha256};
-use std::convert::AsRef;
-use std::convert::TryInto;
+use std::array::TryFromSliceError;
+use std::convert::{AsRef, TryInto};
 use std::fmt;
-use std::fmt::Debug;
 use std::io::Cursor;
 use std::ops::Deref;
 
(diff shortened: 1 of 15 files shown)


chia-protocol/src/bytes.rs
```diff
@@ -211,153 +253,87 @@ impl<N, const LEN: usize> FromClvm<N> for BytesImpl<LEN> {
...
+        &value.0
     }
 }
 
-impl<const N: usize> PartialEq<[u8; N]> for BytesImpl<N> {
-    fn eq(&self, lhs: &[u8; N]) -> bool {
-        &self.0 == lhs
+impl<const N: usize> From<&BytesImpl<N>> for [u8; N] {
+    fn from(value: &BytesImpl<N>) -> [u8; N] {
+        value.0
```
arvidn, line 316: how does this work? This is moving the array out of `value`, how come it's allowed to be taken as an immutable reference?
If it's making a copy, how come you don't need to say `.clone()`?

src/fast_forward.rs
```diff
@@ -109,7 +109,7 @@ pub fn fast_forward_singleton(
 
     // this is the tree hash of the singleton top layer puzzle
     // the tree hash of singleton_top_layer_v1_1.clsp
-    if singleton.args.singleton_struct.mod_hash != SINGLETON_TOP_LAYER_PUZZLE_HASH {
+    if singleton.args.singleton_struct.mod_hash.into_inner() != SINGLETON_TOP_LAYER_PUZZLE_HASH {
```
arvidn, line 112: this is moving the inner `Vec<u8>` out of `singleton.args.singleton_struct.mod_hash`, right?
maybe it doesn't make a difference, but I would have expected something like `.as_ref()`, to have a non-destructive comparison

chia-protocol/src/program.rs
```diff
@@ -81,9 +75,71 @@ impl Program {
     }
 }
 
-impl Default for Program {
-    fn default() -> Self {
-        Self(vec![0x80_u8].into())
+impl From<Bytes> for Program {
```
arvidn, line 78: it's not obvious to me that a `Program` should look so much like a bytes buffer. Do you have specific use cases in mind for this?
My gut says conversions to and from buffers should be explicit

arvidn (commented):
looks good, just a few comments and questions

chia-protocol/src/bytes.rs
Rigidity: Both `BytesImpl<N>` and `[u8; N]` implement the `Copy` trait, so they can be freely copied without worrying about ownership or references. It doesn't actually modify the original byte array, just copies the bits directly.

https://doc.rust-lang.org/std/marker/trait.Copy.html

src/fast_forward.rs
Rigidity: Same as above, since it implements `Copy`, `into_inner` isn't actually dropping the original value. This might be a confusing name for the function though, since typically `into_inner` is reserved for `Owned => Owned` conversions.

https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

Maybe `to_array` would be more appropriate?

Rigidity: In this case I don't know that the copying of 32 bytes will matter all that much, but I can change it to `as_ref()` just in case.

chia-protocol/src/program.rs
Rigidity: No, I suppose `into_inner` is sufficient, I just want to have a way to go from owned `Program` to owned `Bytes`. I can remove the `From` impl if you prefer.

arvidn: I think that would make sense

arvidn (approved):

PR #374, opened 2024-01-23
test building from sdist

altendky:


Rigidity:
@altendky I believe this was implemented again by Earle a while back? Maybe we can close this

altendky:
My recollection was that Earle implemented a fix/workaround for building from sdist and I made this as a check of it.  But, I hadn't gotten back to handling the todo list in the op so that this could be merged.

PR #375, opened 2024-01-23
Add python versions and upload to chia pypi for riscv

emlowe:
Make wheels for supported python versions and added upload step to chia pypi for riscb


arvidn (approved):
There's quite some unix-fu in there :)

generally I think it's good to separate formatting changes from functional changes. It makes it a bit easier to review.

PR #377, opened 2024-01-25
Simplify py-bindings by moving them to dedicated modules

Rigidity:


arvidn (approved):
looks good

Rigidity:
I left alone files that only add like 1 use of `#[cfg(feature = "py-bindings")]`.

chia-bls/src/gtelement.rs
```diff
@@ -153,3 +91,65 @@ impl Streamable for GTElement {
...
+
+#[cfg(feature = "py-bindings")]
+mod py_bindings {
+    use super::*;
+
+    use chia_traits::{from_json_dict::FromJsonDict, to_json_dict::ToJsonDict};
+    use pyo3::{exceptions::PyValueError, prelude::*};
+
+    #[pymethods]
+    impl GTElement {
```
arvidn, line 103: I didn't know this worked. by I suppose since you `use super::*;`, this `impl` block refers to the `GTElement` of the outer scope?

arvidn (approved):

chia-bls/src/gtelement.rs
Rigidity: Yes, you can add an impl block in a separate module. It also affects visibility, making it relative to the module you're implementing it in, which can be useful in some cases.

PR #378, opened 2024-01-29
expose integers as our fixed-width integer types from chia-blockchain

arvidn:
this makes integration of rust types (more) seamless in `chia-blockchain`, and greatly simplifies the patches to transition types. No longer requiring explicit casts to the fixed width integer types


Rigidity (dismissed):
I would love to not require importing `chia-blockchain` anywhere in this repo, but we're already doing so. In the future I think it might be a good idea to implement the sized int types directly in Rust. But this is good for now

Rigidity (approved):

PR #383, opened 2024-02-02
Fix release workflow for riscv

emlowe:
The release workflow wasn't setup correctly, so the release publishing step to publish the riscv wheels to `pypi.chia.net` wasn't working correctly

Hopefully this fixes that problem

Also fixed up copy paste error


hoffmang9 (approved):
lgtm

cmmarslender (approved):

PR #384, opened 2024-02-03
Bump missing versions, update clvmr, add traits and impls back

Rigidity:
This completes the change in https://github.com/Chia-Network/clvm_rs/pull/368, and fixes the unfinished version bump to 0.5.0, hopefully unstalling the publish.


chia-tools/Cargo.toml
```diff
@@ -9,14 +9,14 @@ homepage = "https://github.com/Chia-Network/chia_rs/chia-tools"
 repository = "https://github.com/Chia-Network/chia_rs/chia-tools"
 
 [dependencies]
-chia-protocol = { version = "0.5.0", path = "../chia-protocol" }
-chia-traits = { path = "../chia-traits" }
-clvm-utils = { path = "../clvm-utils" }
-clvm-traits = { version = "=0.3.3", path = "../clvm-traits" }
+chia-protocol = { version = "=0.5.0", path = "../chia-protocol" }
+chia-traits = { version = "=0.5.0", path = "../chia-traits" }
+clvm-utils = { version = "=0.5.0", path = "../clvm-utils" }
```
arvidn, line 14: I thought we were moving away from pinning versions, other than top level projects (say, the wheel). but maybe if we're moving towards `cargo workspace` it won't matter much

arvidn (approved):

chia-tools/Cargo.toml
Rigidity: Technically it doesn't really matter here since you only run it locally and this never gets published. I just made them consistent with the other dependencies in chia-tools. I think it'd be fine to unpin these as well though especially since the version number doesn't affect much.

PR #385, opened 2024-02-03
simplify analyze-chain.rs by using the let-else construct

arvidn:
this patch is best viewed with whitespace changes ignored, since there's a large section with its indentation changed.


Rigidity (approved):
looks good

PR #386, opened 2024-02-04
improve tests and benchmarks of generators

arvidn:
This patch is best reviewed one commit at a time.

1. It adds a tool to print the block generator for a specific block (given its header hash), in hex form.
2. It adds a few large blocks to the `run_generator` test
3. It extends (and fixes) the benchmark to include the new (and some old) block generators

setting `bench=false` for the `[lib]` target means that `cargo bench` no longer will run (and skip) the unit tests. This, in turn, enables `cargo bench` specific command line options, such as:

```
cargo bench -- --save-baseline <name>
cargo bench -- --baseline <name>
```


benches/run-generator.rs
```diff
@@ -46,7 +57,7 @@ fn run(c: &mut Criterion) {
                         11000000000,
                         ALLOW_BACKREFS,
                     );
-                    assert!(conds.is_ok());
+                    let _ = black_box(conds);
```
Rigidity, line 60: Doesn't this ignore the error if there is one?

benches/run-generator.rs
```diff
@@ -63,7 +74,7 @@ fn run(c: &mut Criterion) {
...
```
Rigidity, line 77: This also ignores the errors.

Rigidity (approved):

benches/run-generator.rs
arvidn: yes. some of the generators I included fail. i.e. they exceed the cost limit. I think it's reasonable to have benchmarks that max out the cost. In principle, there could be a valid generator that has a cost just short of this

Rigidity (approved):

PR #387, opened 2024-02-04
Bump to 0.5.1

Rigidity:


arvidn (approved):
I would expect `cargo workspace version` to help with this, right?

Rigidity:
> I would expect `cargo workspace version` to help with this, right?

Well that would republishing everything that changed since the last release, and in this case the last release failed so I don't think that would be accurate.

PR #394, opened 2024-02-06
Add wheel to workspace

Rigidity:


.github/workflows/benchmark.yml
```diff
@@ -128,4 +128,4 @@ jobs:
 
     - name: cargo bench
       run: |
-        cargo bench --all
+        cargo bench --workspace --exclude chia_rs
```
arvidn, line 131: why do you need to exclude `chia_rs`?

arvidn: I think you may be able to say `bench=false` under the `[lib]` target instead

Rigidity: I tried, it didn't work. I needed to exclude it to avoid it complaining about the Python dev libraries not being installed. We don't benchmark the chia_rs wheel so it seemed fine to exclude to me

arvidn (approved):

PR #397, opened 2024-02-08
Add Prettier formatting

Rigidity:


Rigidity:
We can revisit this later

PR #408, opened 2024-02-21
[DRAFT] Add BLSCache struct / pyobject

matt-o-how:


chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,275 @@
+extern crate lru;
```
richardkiss, line 1: I don't think this line is necessary. I don't recall ever having to use `extern crate`. ChatGPT suggests that since the rust 2018 edition you can import with `use` instead.

chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,275 @@
...
+    pyclass(name = "BLSCache"),
+)]
+pub struct BLSCache {
+    cache: LruCache<Bytes32, GTElement>,
+}
+
+impl BLSCache {
+    
+    pub fn generator(cache_size: Option<usize>) -> Self {
+        let cache: LruCache<Bytes32, GTElement> = LruCache::new(NonZeroUsize::new(cache_size.unwrap_or(50000)).unwrap());
```
richardkiss, line 32: I wasn't familiar with `NonZeroUsize`. Seems a bit strange they'd use that in their API. Maybe too clever by half.

Note that using `.unwrap` is a bad smell because it panics on failure. There's probably a way to do this without `.unwrap`. Maybe just do `cache_size = cache_size | 50000`.

```
    pub fn generator(cache_size: usize) -> Self {
        let cache: LruCache<Bytes32, GTElement> = LruCache::new(NonZeroUsize::new(cache_size | 50000).unwrap());
        Self{cache}
    }
```

I haven't test this.

chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,275 @@
...
+        force_cache: bool,
+    ) -> Vec<GTElement> {
+        let mut pairings: Vec<Option<GTElement>> = vec![];
+        let mut missing_count: usize = 0;
+        
+        for (pk, msg) in pks.borrow().iter().zip(msgs.borrow().iter()) {
+            let mut aug_msg = pk.to_vec();
+            aug_msg.extend_from_slice(msg.borrow()); // pk + msg
+            let mut hasher = Sha256::new();
+            hasher.update(aug_msg);
```
richardkiss, line 49: You can call `hasher.update` multiple times, which means you don't actually have to create `aug_msg`.

```diff
-            let mut aug_msg = pk.to_vec();
-            aug_msg.extend_from_slice(msg.borrow()); // pk + msg
             let mut hasher = Sha256::new();
+            hasher.update(pk);
+            hasher.update(msg);
-            hasher.update(aug_msg);
```

richardkiss: (also not tested, so some types may be off by a `&` or something)

chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,275 @@
...
+                    pairings.push(Some(pairing.clone()));
+                },
+                _ => {
+                    pairings.push(None);
+                },
+            }
+            
+        }
+
+        // G1Element.from_bytes can be expensive due to subgroup check, so we avoid recomputing it with this cache
```
richardkiss, line 72: Not sure why there are two loops instead of one. Seems like a natural way to do this would be do dig for the answer in the cache, and if it ain't there, calculate it (and possible cache the result). Maybe it has something to do with the heuristic above. I realize this is just a port so you may not know either.

chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,275 @@
...
+                let mut aug_msg = pks.borrow()[i].to_vec();
+                aug_msg.extend_from_slice(&msgs.borrow()[i]);  // pk + msg
+                let aug_hash: Signature = hash_to_g2(&aug_msg);
+
+                let pk_parsed: &mut PublicKey = pk_bytes_to_g1.entry(pks.borrow()[i]).or_insert_with(|| {
+                    PublicKey::from_bytes(&pks.borrow()[i]).unwrap()
+                });
+
+                let pairing: GTElement = aug_hash.pair(pk_parsed);
+                let mut hasher = Sha256::new();
```
richardkiss, line 89: We do this same hash again, not sure why.

chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,275 @@
...
+        if pairings.is_empty() {
+            let mut data = Vec::<(PublicKey, Vec<u8>)>::new();
+            for (pk, msg) in pks.iter().zip(msgs.iter()) {
+                let pk = PublicKey::from_bytes_unchecked(pk).unwrap();
+                data.push((pk.clone(), msg.clone()));
+            }
+            let res: bool = agg_ver(sig, data);
+            return res
+        }
+        let pairings_prod = pairings.pop(); // start with the first pairing
```
richardkiss, line 117: Note that the `GT` group has an "identity" element which acts like `1` for multiplication of "regular" (ie. real) numbers. If you set the partial product `prod` to this identity element, you don't have to do this weird `.pop` which treats the first argument as special. In fact, you can probably use `.fold` https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold

There are still two cases: an empty list and a non-empty list. It's a bit weird to me that you don't check the signature for the empty list case, but it might be a consensus change to do so depending on how empty blocks are treated (ie. blocks that don't have spends with any `AGG_*` conditions).

chia-bls/src/gtelement.rs
```diff
@@ -19,7 +19,8 @@ use pyo3::exceptions::PyValueError;
 #[cfg(feature = "py-bindings")]
 use pyo3::{pyclass, pymethods, IntoPy, PyAny, PyObject, PyResult, Python};
 
-#[cfg_attr(feature = "py-bindings", pyclass, derive(PyStreamable, Clone))]
+#[cfg_attr(feature = "py-bindings", pyclass, derive(PyStreamable))]
+#[derive(Clone)]
```
richardkiss, line 23: I'm puzzled as to why this might be necessary.

matt-o-how: We want to Clone the GTElement in Rust so we move it out of the py-binding exclusive derives

Rigidity:
Was this superceded by other PRs?

Rigidity:
Going to assume so and close this, feel free to reopen if not

PR #416, opened 2024-03-04
Move crates to dedicated `crates` folder

Rigidity:
There is no convention that I can find per se that says that workspaces should be organized like this, with a `crates` folder rather than putting all of the crates in the root. The closest thing I could find is this [blog post by the creator of Rust Analyzer](https://matklad.github.io/2021/08/22/large-rust-workspaces.html).

However, I think that this is the beginning of a list of changes I want to make:

1. Move all of the crates into a dedicated `crates` folder to reduce clutter in the repository's root.
2. Move the `src` folder and its actual dependencies into a new crate named `chia-consensus`.
3. Turn `chia` into a meta-crate that re-exports each of the workspace's crates as togglable features that are enabled by default.
4. Ideally, find a cleaner way to store test data than the `ff-tests`, `cl`, `test-bundles`, and `generator-tests` folders.
5. Move the `docs` into actual documentation in the root of the appropriate crate.
6. Add an examples folder, with examples for each thing you may want to do with `chia_rs` in Rust (maybe `chia-tools` should live there?).
7. Improve test coverage.


crates/chia-protocol/Cargo.toml
```diff
@@ -5,8 +5,8 @@ edition = "2021"
 license = "Apache-2.0"
 description = "Chia network protocol message types"
 authors = ["Arvid Norberg <arvid@chia.net>"]
-homepage = "https://github.com/Chia-Network/chia_rs/chia-protocol/"
-repository = "https://github.com/Chia-Network/chia_rs/chia-protocol/"
+homepage = "https://github.com/Chia-Network/chia_rs"
+repository = "https://github.com/Chia-Network/chia_rs"
```
arvidn, line 9: An alternative fix would be to point it to: `https://github.com/Chia-Network/chia_rs/tree/main/chia-protocol`. I don't have a strong opinion though.

arvidn (commented):
looks good in general. I'm not sure the globbing is a good idea though

crates/chia-protocol/Cargo.toml
Rigidity: I think it's typical for it to point to the repository's root, like in the following:

- https://github.com/bevyengine/bevy/blob/main/crates/bevy_audio/Cargo.toml#L7
- https://github.com/tokio-rs/tokio/blob/master/tokio-macros/Cargo.toml#L12
- https://github.com/rwf2/Rocket/blob/master/core/codegen/Cargo.toml#L8
- https://github.com/diesel-rs/diesel/blob/master/diesel_cli/Cargo.toml#L9

arvidn (approved):

PR #418, opened 2024-03-04
add sccache to `build-test.yml`

altendky:


.github/workflows/build-test.yml
```diff
@@ -360,6 +360,14 @@ jobs:
         run: |
           git config --global core.autocrlf false
 
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.3
+
+      - name: Set sccache env vars
+        run: |
+          echo "SCCACHE_GHA_ENABLED=true" >> "$GITHUB_ENV"
```
arvidn, line 368:
```suggestion
          echo "SCCACHE_GHA_ENABLED=on" >> "$GITHUB_ENV"
```
according to: https://github.com/mozilla/sccache/blob/main/docs/GHA.md

arvidn (commented):
is this using some sccache service?
it would seem simpler (and perhaps more reliable) to use the github action `cache` action and point it to the cargo `target` directory (which essentially is a ccache already)

altendky:
sccache has builtin GitHub Actions cache support.  In my mind it can be way smarter than caching a directory this way.  I hadn't dug into it but it might be able to be a global cache instead of per-repository with lots of duplication.

PR #420, opened 2024-03-04
Create new `chia-consensus` crate and make `chia` a meta-crate

Rigidity:


crates/chia-consensus/fuzz/Cargo.toml
```diff
@@ -11,11 +11,11 @@ cargo-fuzz = true
...
-clvm-utils = { path = "../crates/clvm-utils" }
-clvm-traits = { path = "../crates/clvm-traits" }
-chia-protocol = { path = "../crates/chia-protocol" }
-chia-traits = { path = "../crates/chia-traits" }
-chia-consensus = { path = "../crates/chia-consensus" }
+clvm-utils = { path = "../../clvm-utils" }
+clvm-traits = { path = "../../clvm-traits" }
+chia-protocol = { path = "../../chia-protocol" }
+chia-traits = { path = "../../chia-traits" }
+chia-consensus = { path = "../../chia-consensus" }
```
arvidn, line 18: technically, this could just be `..`

.github/workflows/build-riscv64.yml
```diff
@@ -40,7 +40,7 @@ jobs:
             -v ${{ github.workspace }}:/ws --workdir=/ws \
             chianetwork/ubuntu-22.04-risc-builder:latest \
             bash -exc '\
-            cargo test --release
+            cargo test --release -p chia-consensus
```
arvidn, line 43: I wonder if it was deliberate to omit the `--all`, and really only run these tests

arvidn (commented):
there's a tool under `chia-tools` that generate an initial corpus for the `chia-consensus` fuzzers. I don't think it's placing the files in the right place now. I think it also needs to be updated.

.github/workflows/build-riscv64.yml
Rigidity: chia-ssl doesn't compile on riscv

arvidn (approved):

PR #422, opened 2024-03-06
Turn consensus docs into a formal readme and crate docs

Rigidity:


arvidn (approved):
looks good, even though I'm not a fan of `_` instead of `*` in markdown :)

PR #424, opened 2024-03-06
Resolve unnecessary reference warnings in benchmarks

Rigidity:


Rigidity:
I presume this wasn't noticed by CI because I'm on nightly and/or the target is not checked by `cargo clippy --workspace` unless you add `-all-targets`?

PR #428, opened 2024-03-09
fewer maturin references

altendky:


llp.sh
```diff
@@ -0,0 +1,3 @@
+# TODO: avoid needing this
+
+export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$(python -c 'import sysconfig; print(sysconfig.get_config_var("LIBDIR"))')
```
arvidn, line 3: I don't understand what this is for

arvidn:
looks like a good approach to me

llp.sh
altendky: i don't remember either...  and it was in my op todo list to look at it.  let's give it a try without.

```suggestion
#export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$(python -c 'import sysconfig; print(sysconfig.get_config_var("LIBDIR"))')
```

PR #429, opened 2024-03-11
Streamable procedural macro

Rigidity:
This is an initial implementation of the streamable procedural macro. It allows you to add additional derives as needed to streamable and message structs.


crates/chia_streamable_macro/src/lib.rs
```diff
@@ -1,15 +1,127 @@
...
+        FoundCrate::Name(name) => {
+            let ident = Ident::new(&name, Span::call_site());
+            quote!(#ident)
+        }
+    };
+
+    let is_message = &attr.to_string() == "message";
+
+    let my_name = std::env::var("CARGO_PKG_NAME").unwrap();
+    let is_protocol = &my_name == "chia-protocol";
```
arvidn, line 30: this looks like a slightly different way to do the same thing as `match found_crate` above. it looks redundant.

crates/chia_streamable_macro/src/lib.rs
```diff
@@ -1,15 +1,127 @@
...
+        }
+    } else {
+        panic!("only structs are supported");
+    }
+
+    let main_derives = quote! {
+        #[derive(Streamable, Hash, Debug, Clone, Eq, PartialEq)]
+    };
+
+    let attrs = if is_protocol {
```
arvidn, line 99: I think it would make sense to add some comments on why some of this depends on `is_protocol`

arvidn (commented):
looks good. I think it would be good with some more comments in the macro implementation justifying the logic in there. For example, answering: "Why not just apply all the same derive macros unconditionally?"

crates/chia_streamable_macro/src/lib.rs
Rigidity: Good point, I'll consolidate it

crates/chia_streamable_macro/src/lib.rs
Rigidity: Yeah I can add that - it's essentially just because `py-bindings` and fuzzing are specific to this crate so if a third party dependent of `chia-traits` tries to define their own `#[streamable]` it shouldn't include those unless they manually annotate them as such

arvidn (approved):

PR #430, opened 2024-03-11
Message conditions

arvidn:
# Overview

This PR implements the message condition codes as described by [CHIP-25](https://github.com/Chia-Network/chips/pull/98).

## Enable/disable

Support for these new conditions is *disabled* by default, and only enabled by passing in the `ENABLE_MESSAGE_CONDITIONS` flag. (the enable/disable logic is implemented in `crates/chia-consensus/src/gen/opcodes.rs`).

## Error codes

There are 3 new error codes introduced by this PR. These will need to be added to `chia-blockchain` as well.

```
ErrorCode::InvalidMessageMode => 145,
ErrorCode::InvalidCoinId => 146,
ErrorCode::MessageNotSentOrReceived => 147,
```

## Deferred memory allocation

Since there's no CLVM cost associated with these new conditions, we need to be careful to mitigate abuse in the form of DoS attacks. When a message is sent or received, we record as little information as possible up front, to resolve later, after all spends has been executed.

Each message has a mask indicating which attributes of itself to take into account. We don't need to store those attributes in the message, just the mask.

Each message also specifies the attributes of the other end of the message (if you're the sender, you specify the receiver's attributes and vice versa). These attributes *do* need to be stored in the message. To avoid memory allocation, only the `NodePtr` are stored. This is encapsulated in the `SpendId` enum` (in `crates/chia-consensus/src/gen/messages.rs`).

Each message has a message `NodePtr`, a mask identifying its own properties to take into account and a `SpendId` indentifying the other end's properties.

When it's time to validate all messages, we turn each message into a key and increment (send message) and decrement (receive message) a counter for each key. At the end we ensure that all counters are 0. This is implemented at the end of the `validate_conditions()` function.

The key generation is where we actually allocate. It's generated in such a way that when the self-attributes matches the parsed-attributes, they are identical. This is how senders and receivers match up. See `SpendId::make_key()` in `crates/chia-consensus/src/gen/messages.rs`.

The justification of avoiding and deferring memory allocation is that one of the most expensive stress tests we have is `duplicate-coin-announce.txt` which takes almost 3 seconds to run. Avoiding the copying of the coin ID (and message) was a material speed-up in earlier tests.

## fast-forward

Spends that send a message that's identified by their parent ID are not eligible for fast forward. Likewise, spends receiving a message identified by their own parent ID are also not eligible for fast forward. This includes messages identified by their coin ID, since the parent ID is part of that.

## deduplication

Spends that send or receive messages are not eligible for de-duplication. De-duplicating such spend would leave the other end of the message dangling. This has some false negatives. Specifically for coins sending a message to itself. It would be safe to de-duplicate such spend, but it's more complicated to carve out that special case. Maybe we can do that in the future if we encounter a use case.

## Testing

There are a large number of test cases in `crates/chia-consensus/src/gen/conditions.rs`.

`test_message_strict_args_count()` makes sure we allow unknown additional parameters, unless `STRICT_ARGS_COUNT` is set (i.e. mempool mode)

`test_message_conditions_single_spend()` has test cases where a single spend sends a message to itself. Some valid in mempool mode, some not.

`test_limit_messages()` ensures that the 1024 limit applies to these new conditions, just like it does to the existing create announcement and assert announcement.

`test_message_conditions_failures()` ensures parse failures are correctly caught and yields the expected error code.

`test_message_conditions_two_spends()` has test cases where one spend sends a message to another spend.

`test_all_message_conditions()` programatically generates all combinations of messages that can be sent from one spend to another. These are only positive tests.

`test_message_eligible_for_ff()` tests all combinations of messages between coins where one coin is eligible for fast-forward. The test ensures that it remains eligible as long as its parent ID is not committed to by the message (both by sending or receiving a message).

## Fuzzing

The `SpendId::parse()` has a fuzzing driver and the `parse-conditions` was extended to enable the message conditions.

# TODO

* [x] Figure out under what circumstances should spends be eligible for fast-forward and dedup. There should probably be more thorough tests and more deliberate logic for this
* [ ] ~~implement stress tests and attempts to flood these messages to ensure the implementation doesn't fall over~~ (I will update the stress tests in a separate PR)

### DIFF
--- a/crates/chia-consensus/src/gen/conditions.rs
+++ b/crates/chia-consensus/src/gen/conditions.rs
@@ -1,5 +1,7 @@
 use super::coin_id::compute_coin_id;
-use super::condition_sanitizers::{parse_amount, sanitize_announce_msg, sanitize_hash};
+use super::condition_sanitizers::{
+    parse_amount, sanitize_announce_msg, sanitize_hash, sanitize_message_mode,
+};
 use super::opcodes::{
     compute_unknown_condition_cost, parse_opcode, ConditionOpcode, AGG_SIG_AMOUNT, AGG_SIG_COST,
     AGG_SIG_ME, AGG_SIG_PARENT, AGG_SIG_PARENT_AMOUNT, AGG_SIG_PARENT_PUZZLE, AGG_SIG_PUZZLE,
@@ -9,8 +11,8 @@ use super::opcodes::{
     ASSERT_HEIGHT_ABSOLUTE, ASSERT_HEIGHT_RELATIVE, ASSERT_MY_AMOUNT, ASSERT_MY_BIRTH_HEIGHT,
     ASSERT_MY_BIRTH_SECONDS, ASSERT_MY_COIN_ID, ASSERT_MY_PARENT_ID, ASSERT_MY_PUZZLEHASH,
     ASSERT_PUZZLE_ANNOUNCEMENT, ASSERT_SECONDS_ABSOLUTE, ASSERT_SECONDS_RELATIVE, CREATE_COIN,
-    CREATE_COIN_ANNOUNCEMENT, CREATE_COIN_COST, CREATE_PUZZLE_ANNOUNCEMENT, REMARK, RESERVE_FEE,
-    SOFTFORK,
+    CREATE_COIN_ANNOUNCEMENT, CREATE_COIN_COST, CREATE_PUZZLE_ANNOUNCEMENT, RECEIVE_MESSAGE,
+    REMARK, RESERVE_FEE, SEND_MESSAGE, SOFTFORK,
 };
 use super::sanitize_int::{sanitize_uint, SanitizedUint};
 use super::validation_error::{first, next, rest, ErrorCode, ValidationErr};
@@ -18,6 +20,7 @@ use crate::gen::flags::{
     AGG_SIG_ARGS, COND_ARGS_NIL, NO_RELATIVE_CONDITIONS_ON_EPHEMERAL, NO_UNKNOWN_CONDS,
     STRICT_ARGS_COUNT,
 };
+use crate::gen::messages::{Message, SpendId};
 use crate::gen::spend_visitor::SpendVisitor;
 use crate::gen::validation_error::check_nil;
 use chia_protocol::bytes::Bytes32;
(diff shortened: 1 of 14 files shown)


crates/chia-consensus/src/gen/messages.rs
```diff
@@ -0,0 +1,271 @@
...
+            _ => Err(ValidationErr(*args, ErrorCode::InvalidMessageMode)),
+        }
+    }
+
+    pub fn from_self(
+        mode: u8,
+        parent: NodePtr,
+        puzzle: NodePtr,
+        amount: u64,
+        coin_id: &Arc<Bytes32>,
```
Rigidity, line 94: Again here, I think `coin_id` should just be `Bytes32`. A reference to an `Arc` is unnecessary for this imo.

Rigidity (commented):
A few small things

crates/chia-consensus/src/gen/messages.rs
```diff
@@ -0,0 +1,271 @@
...
+pub const PUZZLE: u8 = 0b010;
+pub const AMOUNT: u8 = 0b001;
+pub const PUZZLEAMOUNT: u8 = 0b011;
+pub const PARENTAMOUNT: u8 = 0b101;
+pub const PARENTPUZZLE: u8 = 0b110;
+pub const COINID: u8 = 0b111;
+
+#[derive(Debug)]
+pub enum SpendId {
+    OwnedCoinId(Arc<Bytes32>),
```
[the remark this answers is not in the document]
arvidn: it would double the size of `SpendId`. Technically. this *could* be an `Rc`, but pyo3 requires `Send` to pass objects to be run without holding the GIL.
In the tests for create/assert announcement, avoiding the copy was an important step to mitigate DoS attacks. Now that we soft-forked in a limit of 1024, it's still faster to avoid the copy (presumably because of the lower memory footprint.

crates/chia-consensus/src/gen/messages.rs
arvidn: in the stress tests I have, avoiding copying this gives a material speed-up. Granted, I don't have stress tests for these conditions specifically, but there are other stress tests in `sr/gen/test_generators.rs` for `CREATE_COIN` and `CREATE_COIN_ANNOUNCEMENT`

crates/chia-consensus/src/gen/messages.rs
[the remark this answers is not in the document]
Rigidity: Oh I didn't consider the size of the other variants changing, good point.

crates/chia-consensus/src/gen/messages.rs
Rigidity: I'm surprised this is faster, I would think copying 32 bytes to be quite fast. I've been writing `[u8; 32]` instead of `&[u8; 32]` for this reason - but I'll keep it in mind.

arvidn: as long as everything stays in L1 (or maybe even L2) I would share your assumption. However, the stress tests deliberately causes memory pressure to make the whole thing be evicted into L3 and even main RAM. Then you're looking at 2 orders of magnitude longer delays

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -1320,6 +1413,27 @@ pub fn validate_conditions(
...
 
+    if !state.messages.is_empty() {
+        // the integers count the number of times the message has been sent
+        // minus the number of times it's been received. At the end we ensure
+        // all counters are 0, otherwise some message wasn't received or sent
+        // the right number of times.
+        let mut messages = HashMap::<Vec<u8>, i32>::new();
+
+        for msg in &state.messages {
+            *messages.entry(msg.make_key(a)).or_insert(0) += msg.counter as i32;
```
AmineKhaldi, line 1424: According to docs, this is recommended to be something like `messages.entry(msg.make_key(a)).and_modify(|c| *c += msg.counter as i32).or_insert(msg.counter as i32);`

arvidn: I'm not convinced that's an improvement in readability

Rigidity (approved):
lgtm

PR #437, opened 2024-03-15
Add readme badges

emlowe:


hoffmang9 (approved):
We like stinking badges...

Rigidity:
Would like to add one for the Rust meta-crate as well, but that doesn't need to block this

PR #445, opened 2024-03-26
relax mempool mode for message conditions

arvidn:
Any combination of message mode is now allowed in mempool mode as well


Rigidity (approved):
lgtm

Rigidity (approved):

PR #447, opened 2024-03-26
Add UnfinishedBlock2 and released CHIP-0026 protocol messages

Rigidity:


crates/chia-protocol/src/wallet_protocol.rs
```diff
@@ -214,3 +214,79 @@ pub struct RequestFeeEstimates {
...
+    coin_states: Vec<CoinState>,
+}
+
+#[streamable(message)]
+pub struct RejectCoinState {
+    reason: u8, // RejectStateReason
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum RejectStateReason {
```
arvidn, line 289: don't we support `enum` in the `Streamable` macro? I thought as long as they were simple enums it would work

Rigidity: How would the Python bindings work for that?

arvidn: I thought it just turned it into an `int`.

arvidn: I would think you'd want to also add `#[repr(u8)]` here.

wheel/python/chia_rs/chia_rs.pyi
```diff
@@ -3745,10 +3745,10 @@ class RespondPuzzleState:
         coin_states: Union[ List[CoinState], _Unspec] = _Unspec()) -> RespondPuzzleState: ...
 
 class RejectPuzzleState:
-    reason: uint8
+    reason: int
```
arvidn, line 3748: this doesn't look right now. I suppose it doesn't matter much as long as we rely on serializing and de-serializing this on the rust side

arvidn (approved):

PR #450, opened 2024-04-02
Update pyo3 and port consensus constants to Rust

Rigidity:


crates/chia-bls/fuzz/fuzz_targets/blspy-fidelity.rs
```diff
@@ -15,14 +15,6 @@ use chia_bls::secret_key::SecretKey;
...
 use pyo3::types::{PyBytes, PyList, PyTuple};
 
-fn to_bytes(obj: &PyAny) -> &[u8] {
-    obj.call_method0("__bytes__")
-        .unwrap()
-        .downcast::<PyBytes>()
-        .unwrap()
-        .as_bytes()
-}
-
```
arvidn, line 25: it's not obvious to me why you removed this function and inlined its code instead

wheel/generate_type_stubs.py
```diff
@@ -409,3 +412,9 @@ def derive_child_pk_unhardened(pk: G1Element, index: int) -> G1Element: ...
 
     for c in classes:
         print_class(f, c[0], c[1], extra_members.get(c[0]))
+
+    for c in raw_classes:
+        f.write(f"\nclass {c[0]}:\n")
+        for m in c[1]:
+            f.write(f"    {m}\n")
```
arvidn, line 419: by treating this class differently, you don't expose the `replace()` function. I believe we use this function on `ConsensusConstants` in `chia-blockchain`. You also don't expose the streamable interface to python.

Is there a good reason to not parse `ConsensusConstants` into `classes` instead of `raw_classes`?

arvidn:
good catch on `String` -> `str` in the type stubs. I think it would make sense to separate out the fixes and improvement to the stub generation into a separate PR

crates/chia-bls/fuzz/fuzz_targets/blspy-fidelity.rs
Rigidity: Fixed with PyBackedBytes

wheel/generate_type_stubs.py
```diff
@@ -324,51 +340,73 @@ def derive_child_pk_unhardened(pk: G1Element, index: int) -> G1Element: ...
...
-    print_class(f, "PrivateKey", [], [
-        "PRIVATE_KEY_SIZE: ClassVar[int] = ...",
-        "def sign_g2(self, msg: bytes, dst: bytes) -> G2Element: ...",
-        "def get_g1(self) -> G1Element: ...",
-        "def __str__(self) -> str: ...",
-        "def __repr__(self) -> str: ...",
-        ])
-
-    print_class(f, "Spend",
+    print_class(
```
arvidn, line 343: was this caused by the new version of black?

arvidn (commented):
looks good, but let's separate the pyo3 version bump to a separate PR

wheel/generate_type_stubs.py
Rigidity: I don't think the formatter was run on this at all before, but maybe

Rigidity:
Leaving this behind in favor of this for now https://github.com/Chia-Network/chia_rs/pull/455
Will reopen and rebase this after that is merged

PR #451, opened 2024-04-02
Add in a new representation of a MerkleSet which can generate proofs

matt-o-how:
Also adds proof checking and Python bindings.


wheel/src/api.rs
```diff
@@ -77,6 +80,15 @@ pub fn compute_merkle_set_root<'p>(
     Ok(PyBytes::new(py, &compute_merkle_root_impl(&mut buffer)))
 }
 
+#[pyfunction]
+pub fn deserialize_proof(proof: Py<PyBytes>) -> PyResult<MerkleSet> {
+    let bytes = Python::with_gil(|py| proof.as_bytes(py));
+    match deserialise_proof(bytes) {
```
Rigidity, line 86: You can use `.map_err(|_| exceptions::PyValueError::new_err("Error deserialising proof."))` instead of this `match`

src/merkle_tree.rs
```diff
@@ -0,0 +1,1383 @@
...
+            NodeType::Mid
+        };
+        let node_hash = hash(left_type, right_type, &left_hash, &right_hash);
+        merkle_tree.hash_cache.push(node_hash);
+        (node_hash, node_type)
+    }
+}
+
+
+#[cfg(test)]
```
Rigidity, line 571: I typically put everything used in tests inside their own module, like this:
```rs
#[cfg(test)]
mod tests {
    use super::*;
    
    ...
}
```

That way you don't need to manually mark everything as `#[cfg(test)]` and you can import things naturally inside of this module.

src/merkle_tree.rs
```diff
@@ -0,0 +1,1383 @@
...
+    let mut merkle_tree: MerkleSet = MerkleSet::default();
+    let pos = _deserialize(&mut merkle_tree, proof, 0)?;
+    if pos != proof.len() {
+        Err(SetError)
+    } else {
+        Ok(merkle_tree)
+    }
+}
+
+fn _deserialize(
```
Rigidity, line 71: Since things in Rust are private by default, this leading underscore is not necessary. It typically indicates that something is unused, rather than internal.

src/merkle_tree.rs
```diff
@@ -0,0 +1,1383 @@
+// This file contains the code used to create a full MerkleSet and is heavily reliant on the code in merkle_set.rs.
+
+use clvmr::sha2::{Digest, Sha256};
+use crate::merkle_set::{
+    hash,
+    BLANK,
+    get_bit,
+    NodeType,
+};
+#[cfg(test)]
```
Rigidity, line 10: If you move the tests into a module, you wouldn't need these repeated `#[cfg(test)]` attributes. Given, this is something we do in other parts of the code as well - I'd like to clean 
them up eventually. Pretty minor but they get unruly fast if there's a lot of imports or helper functions.

.gitignore
```diff
@@ -1,3 +1,5 @@
 target/
 .idea/
 __pycache__/
+venv
+Cargo.lock
```
Rigidity, line 5: Cargo.lock should not be added to the `.gitignore`, it pins dependency versions and hashes, and ensures consistent builds.

Rigidity: https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control

src/merkle_set.rs
```diff
@@ -64,15 +97,15 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
...
         let right_bit = get_bit(&range[right as usize], depth);
 
-        if left_bit == 1 && right_bit == 0 {
+        if left_bit && !right_bit {
             range.swap(left as usize, right as usize);
             left += 1;
             right -= 1;
         } else {
-            if left_bit == 0 {
+            if !left_bit {
```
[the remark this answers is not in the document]
matt-o-how: This was one of the review suggestions of Richard Kiss

src/merkle_tree.rs
```diff
@@ -0,0 +1,1383 @@
...
+    #[pyo3(name = "get_root")]
+    pub fn py_get_root(&self) -> PyResult<PyObject> {  // compiler doesn't like PyBytes as return type
+        if self.hash_cache.is_empty() {
+            return Err(exceptions::PyValueError::new_err("Tree is empty"))
+        }
+        return Python::with_gil(|py| Ok(PyBytes::new(py, &self.hash_cache[self.hash_cache.len() - 1]).into()));
+    }
+
+    #[pyo3(name = "is_included_already_hashed")]
+    pub fn py_generate_proof(&self, included_leaf: [u8; 32]) -> PyResult<(bool, PyObject)> {
```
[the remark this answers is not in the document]
matt-o-how: the Python name is kept from its current form in chia-blockchain (though I agree it is not a good name)

PR #453, opened 2024-04-02
Create a cache for BLS

matt-o-how:
Creates a BLSCache struct and Py bindings for it.
Also adds a requirements.txt


arvidn:
@matt-o-how I imagine the security warnings comes from just adding the `requirements.txt`. I don't see the dependency graph for those wheels though. Do we need `pluggy` on there? I imagine most of them come via `pytest` though

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,294 @@
...
+                ret.push(pairing);
+            }
+        }
+
+        ret
+    }
+
+    pub fn aggregate_verify(
+        &mut self,
+        pks: &Vec<Bytes48>,
```
Rigidity, line 107: Could do the same thing here, `&[PublicKey]` and `&[M]` where `M: AsRef<[u8]>`

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,294 @@
...
+impl BLSCache {
+    #[new]
+    pub fn init() -> Self {
+        Self::generator(None)
+    }
+
+    #[staticmethod]
+    #[pyo3(name = "generator")]
+    pub fn py_generator(size: Option<&PyInt>) -> Self {
+        match size {
```
Rigidity, line 148: If you were to implement `Default` for `BLSCache`, you could simplify this to:

```rs
size.map(|s| Self::new(s.extract().unwrap())).unwrap_or_default()
```

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,294 @@
...
+            .collect::<PyResult<_>>()?;
+        let msgs_r: Vec<Vec<u8>> = msgs
+            .iter()
+            .map(|item| item.extract::<Vec<u8>>())
+            .collect::<PyResult<_>>()?;
+        let force_cache_bool = force_cache.extract::<bool>()?;
+        Ok(self.aggregate_verify(&pks_r, &msgs_r, sig, force_cache_bool))
+    }
+
+    #[pyo3(name = "len")]
```
Rigidity, line 177: I think you can implement magic methods like `__len__`, but I suppose it doesn't matter much.

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    #[test]
+    pub fn test_aggregate_verify() {
+        let mut bls_cache: BLSCache = BLSCache::generator(None);
+        assert_eq!(bls_cache.cache.len(), 0);
+        let byte_array: [u8; 32] = [0; 32];
+        let sk: SecretKey = SecretKey::from_seed(&byte_array);
+        let pk: PublicKey = sk.public_key();
+        let msg: Vec<u8> = [106; 32].to_vec();
```
Rigidity, line 215: FYI you can use `vec![106; 32]` to shortcut this. Although I'm not convinced you need all these `Vec` instances anyways. They can be potentially expensive to allocate on the heap. In tests it doesn't matter as much though.

wheel/README.md
```diff
@@ -1 +1,12 @@
 The `chia_rs` wheel contains python bindings for code from the `chia` crate.
+
+To run the tests:
+```
+cd wheel
+pytho -m venv venv
```
Rigidity, line 6: typo

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,294 @@
...
+        let mut pairings: Vec<Option<GTElement>> = vec![];
+        let mut missing_count: usize = 0;
+
+        for (pk, msg) in pks.borrow().iter().zip(msgs.borrow().iter()) {
+            let mut aug_msg = pk.to_vec();
+            aug_msg.extend_from_slice(msg.borrow()); // pk + msg
+            let mut hasher = Sha256::new();
+            hasher.update(aug_msg);
+            let h: Bytes32 = hasher.finalize().into();
+            let pairing: Option<&GTElement> = self.cache.get(&h);
```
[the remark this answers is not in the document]
matt-o-how: That first line returns a RangeFull - I'm not really sure what you're saying to replace exactly here.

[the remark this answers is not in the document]
Rigidity: `...` as in the code below in the match. Not necessary but `Option::map` does make the code more concise here.

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,294 @@
...
+}
+
+impl BLSCache {
+    pub fn generator(cache_size: Option<usize>) -> Self {
+        let cache: LruCache<Bytes32, GTElement> =
+            LruCache::new(NonZeroUsize::new(cache_size.unwrap_or(50000)).unwrap());
+        Self { cache }
+    }
+
+    pub fn get_pairings<P: Borrow<[Bytes48]>, M: Borrow<[Vec<u8>]>>(
```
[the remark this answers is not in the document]
matt-o-how: PublicKey isn't accurate though because we're passing in bytes rather than the PublicKey struct

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,294 @@
...
+        pks: &PyList,
+        msgs: &PyList,
+        sig: &Signature,
+        force_cache: &PyBool,
+    ) -> PyResult<bool> {
+        let pks_r: Vec<Bytes48> = pks
+            .iter()
+            .map(|item| item.extract::<Bytes48>())
+            .collect::<PyResult<_>>()?;
+        let msgs_r: Vec<Vec<u8>> = msgs
```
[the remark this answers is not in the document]
matt-o-how: I may be missing something but if we have to convert it to a Vec anyway, how does extracting bytes first save anything over just extracting as Vec?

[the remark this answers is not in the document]
Rigidity: That's the point though, you shouldn't need to convert them to a `Vec` as long as they implement `AsRef<[u8]>`, right?

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,295 @@
...
+        Self { cache }
+    }
+
+    pub fn get_pairings<P: Borrow<[[u8; 48]]>, M: AsRef<[u8]>>(
+        &mut self,
+        pks: &P,
+        msgs: &[M],
+        force_cache: bool,
+    ) -> Vec<GTElement> {
+        let mut pairings: Vec<Option<GTElement>> = vec![];
```
arvidn, line 55: this temporary vector, as well as the `ret` returned vector are probably pretty expensive since they allocate memory and increase the memory footprint of this function. Since `get_pairings()` is only used internally, we don't necessarily need to maintain this interface. For example, we could instead pass in a mutable `blst_pairing` object (or perhaps some abstraction of it) and directly apply the cached `GTElements` with `blst_pairing_raw_aggregate()`. That would very likely be a lot more efficient.

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,295 @@
...
+        let mut pairings: Vec<GTElement> = self.get_pairings(pks, msgs, force_cache);
+        if pairings.is_empty() {
+            let mut data = Vec::<(PublicKey, &[u8])>::new();
+            for (pk, msg) in pks.iter().zip(msgs.iter()) {
+                let pk = PublicKey::from_bytes_unchecked(pk).unwrap();
+                data.push((pk.clone(), msg));
+            }
+            return agg_ver(sig, data);
+        }
+        let pairings_prod = pairings.pop(); // start with the first pairing
```
arvidn, line 128: at this point you're guaranteed that `pairings` is not empty, right? So it would seem you could just do:
```
let pairing_prod = pairing.pop().expect("internal error");
```

But I also think that this loop will be quite slow compared to the built-in BLST functions to verify a signature.

See this function: https://github.com/Chia-Network/chia_rs/blob/main/crates/chia-bls/src/signature.rs#L403

crates/chia-bls/src/gtelement.rs
```diff
@@ -19,7 +19,8 @@ use pyo3::exceptions::PyValueError;
 #[cfg(feature = "py-bindings")]
 use pyo3::{pyclass, pymethods, IntoPy, PyAny, PyObject, PyResult, Python};
 
-#[cfg_attr(feature = "py-bindings", pyclass, derive(PyStreamable, Clone))]
+#[cfg_attr(feature = "py-bindings", pyclass, derive(PyStreamable))]
+#[derive(Clone)]
```
arvidn, line 23: If you remove the temporary vectors, I don't think you need this to be clonable anymore

arvidn (changes requested):

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,295 @@
...
+                ret.push(pairing);
+            }
+        }
+
+        ret
+    }
+
+    pub fn aggregate_verify(
+        &mut self,
+        pks: &Vec<[u8; 48]>,
```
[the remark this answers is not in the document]
matt-o-how: The pubkeys that get passed in are taken from ChiaLisp Conditions so are just bytes

arvidn:
this is a sensitive part of the consensus. we need really thorough tests and fuzzers. I think we should have some benchmarks as well

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,291 @@
...
+        sig: &Signature,
+        force_cache: bool,
+    ) -> bool {
+        let mut pairings: Vec<GTElement> = self.get_pairings(pks, msgs, force_cache);
+        if pairings.is_empty() {
+            let mut data = Vec::<(PublicKey, &[u8])>::new();
+            for (pk, msg) in pks.iter().zip(msgs.iter()) {
+                let pk = PublicKey::from_bytes_unchecked(pk).unwrap();
+                data.push((pk.clone(), msg.as_ref()));
+            }
```
arvidn, line 120: creating a new vector to push the key and message to is not necessary. You can pass in `pks` and `msgs` directly to `agg_ver()`

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,291 @@
...
+            let h: [u8; 32] = hasher.finalize().into();
+
+            let pairing = self.cache.get(&h).cloned();
+
+            if !force_cache && pairing.is_some() {
+                // Heuristic to avoid more expensive sig validation with pairing
+                // cache when it's empty and cached pairings won't be useful later
+                // (e.g. while syncing)
+                missing_count += 1;
+                if missing_count > pks.len() / 2 {
```
arvidn, line 69: I think we need benchmarks to find this cut-off, if it even makes sense to have logic like this.
This is determined by how much faster the normal `aggregate_verify()` is compared to the multiplication loop of `GTElement`s.

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,234 @@
...
+            hasher.update(&aug_msg);
+            let h: [u8; 32] = hasher.finalize().into();
+            self.cache.put(h, pairing.clone());
+            pairing
+        });
+        agg_ver_gt(sig, iter)
+    }
+}
+
+// Python Functions
```
arvidn, line 80: I don't think we should keep the python code as a comment

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,234 @@
...
+impl Default for BLSCache {
+    fn default() -> Self {
+        Self::new(50000)
+    }
+}
+
+impl BLSCache {
+    pub fn new(cache_size: usize) -> BLSCache {
+        let cache: LruCache<[u8; 32], GTElement> =
+            LruCache::new(NonZeroUsize::new(cache_size).unwrap());
```
arvidn, line 36: is it really reasonable to `panic!()` if you pass 0 to `new()`?

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,234 @@
...
+        Self { cache }
+    }
+
+    pub fn generator(cache_size: Option<usize>) -> Self {
```
arvidn, line 40: it's not obvious that it still makes sense to keep this function

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,234 @@
...
+
+#[cfg(test)]
+pub mod tests {
+    use super::*;
+    use crate::aggregate;
+    use crate::sign;
+    use crate::SecretKey;
+
+    #[test]
+    pub fn test_instantiation() {
```
arvidn, line 134: I don't think these tests cover all interesting cases. It would probably easier to read these if the setup-code was factored out, to make each test clearly show the actual function that's being tested.

I would like to see tests where we have 1000 public keys + messages validated with 0% cache hit rate, 100% cache hit rate, and something in between, ensuring we get the same result.

Also some negative tests, ensuring we *catch* invalid signatures (see the tests in `chia-bls`).

When you removed the `force_cache` option, didn't any tests become redundant? Or did we not have tests for that feature before?

tests/test_blscache.py
```diff
@@ -0,0 +1,58 @@
+from chia_rs import G1Element, PrivateKey, AugSchemeMPL, G2Element, BLSCache
+from chia.util.ints import uint64
+from chia.types.blockchain_format.sized_bytes import bytes48
```
arvidn, line 3: you should import these from `chia_rs`.
```
from chia_rs.sized_bytes import bytes48
```

tests/test_blscache.py
```diff
@@ -0,0 +1,58 @@
...
+import pytest
+from typing import List
+
+
+def test_instantiation() -> None:
```
arvidn, line 8: would it make sense to have tests, in python, that validate that the rust implementation behaves exactly like the python one?

arvidn (changes requested):
please also add a benchmark where we have 100% cache hit rate, 0% cache hit rate, 50% cache hit rate, and compare it against the same validation using `aggregate_verify()`. I would expect `aggregate_verify()` to be faster for 0% cache hit rate, but it would be interesting to see by how much.

crates/chia-bls/src/cached_bls.rs
Rigidity: Should imo just make the input parameter NonZeroUsize too

crates/chia-bls/src/cached_bls.rs
arvidn: I think you should still make this benchmark to know if there's still a use for a heuristic like this

crates/chia-bls/src/cached_bls.rs
```diff
@@ -45,40 +43,23 @@ impl BLSCache {
...
-
             if let Some(pairing) = pairing {
                 // equivalent to `if pairing is not None`
-                if let Some(ref current) = ret {
-                    ret = Some(current * &pairing);
-                } else {
-                    ret = Some(pairing);
-                }
-                continue;
+                return pairing
```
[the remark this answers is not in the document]
matt-o-how: No because otherwise it's a move and we can't return it. Annoying.

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,234 @@
...
+
+    pub fn generator(cache_size: Option<usize>) -> Self {
+        let cache: LruCache<[u8; 32], GTElement> =
+            LruCache::new(NonZeroUsize::new(cache_size.unwrap_or(50000)).unwrap());
+        Self { cache }
+    }
+
+    pub fn aggregate_verify<M: AsRef<[u8]>>(
+        &mut self,
+        pks: &[PublicKey],
```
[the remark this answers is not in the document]
Rigidity: Are you thinking something like `impl IntoIterator<Item = impl Borrow<PublicKey>>`?

crates/chia-bls/src/cached_bls.rs
```diff
@@ -43,15 +44,16 @@ impl BLSCache {
         Self { cache }
     }
 
-    pub fn aggregate_verify<M: AsRef<[u8]>>(
+    pub fn aggregate_verify<M: Iterator<Item = T>, T: AsRef<[u8]>, P: Iterator<Item = U>, U: Borrow<PublicKey>>(
```
Rigidity, line 47: Wouldn't IntoIterator be more versatile? You could pass in an actual vector, slice, or array, without manually creating an iterator over it?

Existing iterators also implement IntoIterator for themselves I believe

matt-o-how: Ah, fantastic. This is very useful. I did feel like my solution was not optimal and this is the exact thing I was looking for thank you

crates/chia-bls/src/cached_bls.rs
```diff
@@ -0,0 +1,235 @@
...
+    {
+        let iter = pks.into_iter().zip(msgs).map(|(pk, msg)| -> GTElement {
+            let mut hasher = Sha256::new();
+            hasher.update(pk.borrow().to_bytes());
+            hasher.update(msg); // pk + msg
+            let h: [u8; 32] = hasher.finalize().into();
+
+            let pairing: Option<GTElement> = self.cache.get(&h).cloned();
+
+            if let Some(pairing) = pairing {
```
arvidn, line 69:
```suggestion
            if let Some(pairing) = self.cache.get(&h).cloned() {
```

wheel/generate_type_stubs.py
```diff
@@ -338,6 +338,12 @@ def serialized_length(program: ReadableBuffer) -> int: ...
 def tree_hash(program: ReadableBuffer) -> bytes32: ...
 def get_puzzle_and_solution_for_coin(program: ReadableBuffer, args: ReadableBuffer, max_cost: int, find_parent: bytes32, find_amount: int, find_ph: bytes32, flags: int) -> Tuple[bytes, bytes]: ...
 
+class BLSCache:
+    def len(self) -> uint128: ...
```
arvidn, line 342: this returns an `int`, not `uint128`

tests/test_blscache.py
matt-o-how: After moving to PublicKey we no longer need bytes48 at all - it's removed now

arvidn (commented):
do you feel confident the unit tests cover all interesting cases?
A few cases are:
0% cache hit rate
100% cache hit rate
validating a signature of 0 messages and keys
validating a signature of 1 message and key

Can you think if any aspect that might benefit from fuzzing?

crates/chia-bls/src/cached_bls.rs
```diff
@@ -91,15 +88,16 @@ impl BLSCache {
...
-
-    #[staticmethod]
-    #[pyo3(name = "generator")]
-    pub fn py_generator(size: Option<&PyInt>) -> Self {
-        size.map(|s| Self::new(s.extract().unwrap()))
-            .unwrap_or_default()
+    pub fn init(size: Option<&PyInt>) -> PyResult<Self> {
+        match size {
+            Some(p_size) => {
+                let r_size = p_size.extract::<isize>().unwrap();
```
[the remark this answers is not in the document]
Rigidity: Since this returns a `PyResult` I'd imagine you could just use `?`

crates/chia-bls/src/cached_bls.rs
```diff
@@ -91,15 +88,16 @@ impl BLSCache {
...
+                if r_size < 1 {return Err(PyValueError::new_err("Cannot have a cache size less than one."))} else {
+                    return Ok(Self::new(NonZeroUsize::new(p_size.extract::<usize>().unwrap())))
+                }
+            },
```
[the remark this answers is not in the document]
Rigidity: And editor language support so it formats on save is really helpful too

crates/chia-bls/src/cached_bls.rs
```diff
@@ -232,4 +231,12 @@ pub mod tests {
         // assert first key has been removed
         assert!(bls_cache.cache.get(&h).is_none());
     }
+
+    pub fn test_empty_sig() {
```
arvidn, line 235: this should not be `pub` and it should be marked `#[test]`

crates/chia-bls/src/cached_bls.rs
arvidn: All test functions should be private. i.e. not `pub`, unless you actually use them from some other file. I don't expect you to though, as that would be a bit odd.

tests/test_blscache.py
```diff
@@ -0,0 +1,164 @@
...
+
+    with pytest.raises(OverflowError) as exc_info:
+        bls_cache = BLSCache(-100000)
+
+    assert str(exc_info.value) == "can't convert negative int to unsigned"
+
+    with pytest.raises(OverflowError) as exc_info:
+        bls_cache = BLSCache(-9223372036854775809)
+
+    assert str(exc_info.value) == "can't convert negative int to unsigned"
```
arvidn, line 164: instead of these checks, you can add another parameter to `pytest.raises()`, `match="can't convert negative int to unsigned"`

arvidn (approved):

PR #454, opened 2024-04-03
Add MerkleSet struct to Rust

matt-o-how:
Also includes Python bindings and tests for Rust and Python


.github/workflows/benchmark.yml
```diff
@@ -49,6 +49,8 @@ jobs:
         python -m pip install maturin
         rustup target add x86_64-unknown-linux-musl
         python -m pip install pytest
+        python -m pip install pytest-asyncio
```
arvidn, line 52: as far as I can tell, you don't actually need this dependency

.github/workflows/benchmark.yml
```diff
@@ -49,6 +49,8 @@ jobs:
...
+        python -m pip install trio
```
arvidn, line 53: what's trio for?

wheel/generate_type_stubs.py
```diff
@@ -447,5 +447,18 @@ def derive_child_pk_unhardened(pk: G1Element, index: int) -> G1Element: ...
         ],
     )
 
+    print_class(f, "MerkleSet",
```
arvidn, line 450: since this class doesn't derive from streamable or implement any of the special functions the other protocol classes implement, I don't think you should use `print_class()` here. Just print the type stub directly to the output.

wheel/chia_rs.pyi
```diff
@@ -345,6 +345,41 @@ class SpendBundleConditions:
...
+    def to_bytes(self) -> bytes: ...
+    def __bytes__(self) -> bytes: ...
+    def stream_to_bytes(self) -> bytes: ...
+    def get_hash(self) -> bytes32: ...
+    def to_json_dict(self) -> Dict[str, Any]: ...
+    @staticmethod
+    def from_json_dict(json_dict: Dict[str, Any]) -> MerkleSet: ...
+    def replace(self, *, nodes: Union[ List[uint64], _Unspec] = _Unspec(),
+        leafs: Union[ List[bytes], _Unspec] = _Unspec(),
+        hash_cache: Union[ List[bytes], _Unspec] = _Unspec()) -> MerkleSet: ...
```
arvidn, line 381: I don't think this type-stub is correct given your python binding.

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -0,0 +1,1362 @@
...
+const EMPTY_NODE_HASH: [u8; 32] = [
+    127, 156, 158, 49, 172, 130, 86, 202, 47, 37, 133, 131, 223, 38, 45, 188, 125, 111, 104, 242,
+    160, 48, 67, 213, 201, 154, 74, 229, 167, 57, 108, 233,
+];
+
+#[derive(Debug)]
+#[cfg_attr(feature = "py-bindings", pyclass(frozen, name = "SetError"))]
+pub struct SetError;
+
+pub fn deserialize_proof(proof: &[u8]) -> Result<MerkleSet, SetError> {
```
arvidn, line 52: this function should have a fuzzer

arvidn (changes requested):

.github/workflows/benchmark.yml
matt-o-how: running the tests locally wouldn't work without pytest-asyncio and trio. I'm not exactly sure why, if you can run them without these then I'll delete

crates/chia-consensus/src/merkle_set.rs
```diff
@@ -147,567 +185,4 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
...
-            hasher.finalize().into()
-        }
-        (hash, NodeType::Mid) => hash,
-        (hash, NodeType::MidDbl) => hash,
-        (_, NodeType::Empty) => panic!("unexpected"),
-    }
-}
-
-#[test]
-fn test_get_bit_msb() {
```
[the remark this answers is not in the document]
matt-o-how: The tests also test the new code at the same time, so to prevent circular imports and leave your code as intact as possible I moved them to the new file

crates/chia-consensus/src/merkle_set.rs
```diff
@@ -147,567 +185,4 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
     }
 }
 
-pub fn compute_merkle_set_root(leafs: &mut [[u8; 32]]) -> [u8; 32] {
```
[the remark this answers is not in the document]
matt-o-how: It makes some logical sense that it should come before `radix_sort` in the code as it calls `radix_sort` - it's not urgent, though it is one of Richard's preferences as well

crates/chia-consensus/src/merkle_set.rs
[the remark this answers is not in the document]
Rigidity: As far as I'm aware it's fine to have two files use items from each other in Rust, as long as they are part of the same crate.

wheel/chia_rs.pyi
matt-o-how: These were autogenerated from `print_class` and then copied. What seems to be wrong specifically?

crates/chia-consensus/src/merkle_set.rs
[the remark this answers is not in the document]
matt-o-how: Either way the tests have been expanded to test the new things so it makes to put them with the new things. There is a comment at the bottom of merkle_set pointing to where the tests now are

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -0,0 +1,1362 @@
...
+    }
+}
+
+impl MerkleSet {
+    pub fn get_merkle_root(&self) -> [u8; 32] {
+        self.hash_cache[self.hash_cache.len() - 1]
+    }
+
+    // returns a tuple of a bool representing if the value has been found, and if so bytes that represent the proof of inclusion
+    pub fn generate_proof(&self, included_leaf: [u8; 32]) -> Result<(bool, Vec<u8>), SetError> {
```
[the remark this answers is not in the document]
matt-o-how: This is the current return structure for the Python and I want to make as minimal changes as possible to the interface structure

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -0,0 +1,1362 @@
...
+        let mut proof = Vec::new();
+        let r = self.is_included(self.nodes_vec.len() - 1, included_leaf, &mut proof, 0)?;
+        Ok((r, proof))
+    }
+
+    pub fn is_included(
```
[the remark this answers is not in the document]
matt-o-how: Again, structure / name brought over from Python

.github/workflows/benchmark.yml
matt-o-how: nvm fixed

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -0,0 +1,1362 @@
...
+fn _deserialize(merkle_tree: &mut MerkleSet, proof: &[u8], pos: usize) -> Result<usize, SetError> {
+    if let Some(&t) = proof.get(pos) {
+        match t {
+            EMPTY => {
+                merkle_tree.nodes_vec.push(ArrayTypes::Empty);
+                merkle_tree.hash_cache.push(BLANK);
+                Ok(pos + 1)
+            }
+            TERMINAL => {
+                let hash: [u8; 32] = proof[pos + 1..pos + 33].try_into().map_err(|_| SetError)?;
```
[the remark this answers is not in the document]
matt-o-how: fixed

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -0,0 +1,1362 @@
...
+                merkle_tree.leaf_vec.push(hash);
+                merkle_tree.nodes_vec.push(ArrayTypes::Leaf {
+                    data: merkle_tree.leaf_vec.len() - 1,
+                });
+                merkle_tree.hash_cache.push(hash);
+                Ok(pos + 33)
+            }
+            TRUNCATED => {
+                let hash: [u8; 32] = proof[pos + 1..pos + 33].try_into().map_err(|_| SetError)?;
```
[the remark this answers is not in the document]
matt-o-how: fixed

crates/chia-consensus/src/merkle_set.rs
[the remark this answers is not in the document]
arvidn: it makes it very hard to review what happened to them, when they're moved and change in the same PR.

crates/chia-consensus/src/merkle_set.rs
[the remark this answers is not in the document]
arvidn: was it moved verbatim? if not, it's quite hard to review any changes

wheel/chia_rs.pyi
arvidn: `print_class()` assumes a rust `struct` that derives from `Streamable` and `PyStreamable`. That's where a lot of these member functions come from. I don't think your class should implement these functions, so the simplest thing would be to just hard code the type stub in the generation script

crates/chia-consensus/src/merkle_tree.rs
arvidn: a fuzzer is much better at finding edge cases. It's not random, it's code coverage guided.

crates/chia-consensus/src/merkle_set.rs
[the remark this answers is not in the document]
matt-o-how: It was not - I added in extra lines to check the new code as well

crates/chia-consensus/src/merkle_set.rs
```diff
@@ -37,13 +45,43 @@ fn hash(ltype: NodeType, rtype: NodeType, left: &[u8; 32], right: &[u8; 32]) ->
     hasher.finalize().into()
 }
 
-const BLANK: [u8; 32] = [
+pub const BLANK: [u8; 32] = [
```
Rigidity, line 48: Instead of having a public constant for this, you could just use `[0; 32]` anywhere you would use it. This fills a new 32 byte array with zeros.

wheel/Cargo.lock
Rigidity, line 1: We have a global `Cargo.lock` for the whole workspace which now includes the `wheel`. Why was this added?

Rigidity (commented):
Looks good, just a couple things

crates/chia-consensus/src/merkle_set.rs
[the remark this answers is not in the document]
matt-o-how: Disregard, my now deleted comment. Yes it was moved verbatim.

crates/chia-consensus/src/merkle_set.rs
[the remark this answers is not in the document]
matt-o-how: Ok, I've moved the tests back into merkle_set.rs

wheel/Cargo.lock
matt-o-how: probably a mistake - deleting it now

crates/chia-consensus/src/merkle_set.rs
Rigidity: Maybe just not public then, we don't want to clutter the public API with things only relevant to dependents

crates/chia-consensus/src/merkle_set.rs
```diff
@@ -147,567 +185,892 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
...
+
+            let (root, tree) = generate_merkle_tree(&mut random_data);
+            assert_eq!(tree.hash_cache.len(), tree.nodes_vec.len());
+            assert_eq!(root, tree.hash_cache[tree.nodes_vec.len() - 1]);
+            assert_eq!(root, compute_merkle_set_root(&mut random_data));
+            let mut rng = rand::thread_rng();
+            let index = rng.gen_range(0..random_data.len());
+            let (included, proof) = tree.generate_proof(random_data[index]).unwrap();
+            assert!(included);
+            let rebuilt = deserialize_proof(&proof[0..proof.len() - 2]);
```
arvidn, line 1072: these proofs are mostly valid, just truncated. I think a fuzzer that passes in truly arbitrary bytes would be very valuable.

crates/chia-consensus/src/merkle_set.rs
```diff
@@ -147,567 +185,892 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
...
+                tree.hash_cache[tree.hash_cache.len() - 1]
+            );
+        }
+    }
+
+    #[test]
+    fn test_bad_proofs() {
+        for _i in [1..1000000] {
+            // Create a random number generator
+            let mut small_rng = SmallRng::from_entropy();
```
arvidn, line 1051: I think our tests should be deterministic, so I think we should use a fixed seed for the rngs

crates/chia-consensus/src/merkle_set.rs
```diff
@@ -147,567 +185,892 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
...
+                let mut array: [u8; 32] = [0; 32];
+                small_rng.fill(&mut array);
+                random_data.push(array);
+            }
+
+            let (root, tree) = generate_merkle_tree(&mut random_data);
+            assert_eq!(tree.hash_cache.len(), tree.nodes_vec.len());
+            assert_eq!(root, tree.hash_cache[tree.nodes_vec.len() - 1]);
+            assert_eq!(root, compute_merkle_set_root(&mut random_data));
+            let mut rng = rand::thread_rng();
```
arvidn, line 1068: there's no threading here, right? my understanding is that this is an RNG whose state is kept in TLS. it seems unnecessary and a bit confusing to suggest threading is used. Also, using a fixed seed would be preferable I think. (in which case you also need to move these out of the loop)

crates/chia-consensus/src/merkle_set.rs
```diff
@@ -147,567 +185,892 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
...
+            let index = rng.gen_range(0..random_data.len());
```
arvidn, line 1036: what would you think of doing this in a loop, and test every leaf? Maybe the outer loop would need to be made a bit smaller (at least in debug mode)

crates/chia-consensus/src/merkle_set.rs
```diff
@@ -147,567 +185,892 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
...
+
+        let ArrayTypes::Middle { children } = tree.nodes_vec.last().unwrap() else {
+            panic!("expected middle node");
+        };
+        let ArrayTypes::Leaf { .. } = tree.nodes_vec[children.0] else {
+            panic!("expected leaf");
+        };
+        // generate proof of inclusion for every node
+        let (included, _proof) = tree.generate_proof(d).unwrap();
+        assert!(included);
```
arvidn, line 999:
```suggestion
        assert!(tree.generate_proof(d).unwrap().0);
```

arvidn: but maybe it would make sense to validate the actual proof as well

crates/chia-consensus/src/merkle_set.rs
```diff
@@ -147,567 +185,892 @@ fn radix_sort(range: &mut [[u8; 32]], depth: u8) -> ([u8; 32], NodeType) {
...
+        let (included, _proof) = tree.generate_proof(a).unwrap();
+        assert!(included);
+        let (included, proof) = tree.generate_proof(c).unwrap();
+        assert!(included);
+        assert_eq!(tree.hash_cache.len(), tree.nodes_vec.len());
+        let rebuilt = deserialize_proof(&proof).unwrap();
+        assert_eq!(
+            rebuilt.hash_cache[rebuilt.hash_cache.len() - 1],
+            tree.hash_cache[tree.hash_cache.len() - 1]
+        );
```
arvidn, line 1009: this is just checking the root hash, right? I would seem reasonable to have at least one test to check the uncle hashes as well

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -0,0 +1,1362 @@
...
+#[cfg(test)]
+use rand::rngs::SmallRng; // cargo says this isn't required but tests won't run without it
+#[cfg(test)]
+use rand::{Rng, SeedableRng}; // needed for PyBytes
+
+// the ArrayType is used to create a more lasting MerkleTree representation in the MerkleSet struct
+#[derive(PartialEq, Debug, Copy, Clone)]
+pub enum ArrayTypes {
+    Leaf { data: usize },                // indexes for the data_hash array
+    Middle { children: (usize, usize) }, // indexes for a Vec of ArrayTypes
```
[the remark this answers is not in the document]
arvidn: there's no cost associated with a cast, but there *is* cost associated with using memory.

crates/chia-consensus/src/merkle_tree.rs
arvidn: I commented in the tests; they pass in valid proofs (or valid but truncated proofs). when we deserialize proofs from the network, we can't trust them so we need a fuzzer to test potentially malicious input.

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -0,0 +1,516 @@
...
+            let hash: [u8; 32] = proof[pos + 1..pos + 33].try_into().map_err(|_| SetError)?;
+            merkle_tree.nodes_vec.push(ArrayTypes::Truncated);
+            merkle_tree.hash_cache.push(hash);
+            Ok(pos + 33)
+        }
+        MIDDLE => {
+            if proof.len() < pos + 1 {
+                return Err(SetError);
+            };
+            let new_pos = deserialize_recurse(merkle_tree, proof, pos + 1)?;
```
arvidn, line 102: something I would expect a fuzzer to find is the case where you pass a buffer of 4000 `0x03`-bytes. that should cause a stack overflow. It would probably be best to not implement this recursively, but keep the parse-stack explicitly in a `Vec`

crates/chia-consensus/src/merkle_tree.rs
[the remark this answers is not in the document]
arvidn: I agree that rationale makes sense for the python binding, but not for the rust API

crates/chia-consensus/src/merkle_tree.rs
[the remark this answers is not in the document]
arvidn: this isn't exported to python though, so I think it making sense is more important. But maybe there's a really good reason to call it this (it would warrant a comment though, I think). The convention in rust is that functions of the form `is_*()` query the state of a class.

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -0,0 +1,516 @@
...
+                if get_bit(&included_leaf, depth) {
+                    // bit is 1 so truncate left branch and search right branch
+                    self.other_included(
+                        children.0,
+                        included_leaf,
+                        proof,
+                        depth + 1,
+                        matches!(self.nodes_vec[children.1], ArrayTypes::Empty),
+                    )?;
+                    self.is_included(children.1, included_leaf, proof, depth + 1)
```
arvidn, line 188: is it possible to cause a stack overflow here? These trees cannot be deeper than 256 level I suppose, right?
(it might be good to comment on that, to help future readers of this code be convinced that it's valid)

arvidn (changes requested):

crates/chia-consensus/src/merkle_set.rs
matt-o-how: I'll set it to `pub(crate)` so that `merkle_tree.rs` can also use it.

crates/chia-consensus/src/merkle_set.rs
```diff
@@ -37,13 +45,43 @@ fn hash(ltype: NodeType, rtype: NodeType, left: &[u8; 32], right: &[u8; 32]) ->
...
+    }
+}
+
 // this function performs an in-place, recursive radix sort of the range.
 // as each level returns, values are hashed pair-wise and as a hash tree.
-// the return value is the merkle tree root that the values in the range form
+// It will also populate a MerkleTreeData struct at each level of the call
+// the return value is a tuple of:
+// - merkle tree root that the values in the range form
+// - the type of node that this is
```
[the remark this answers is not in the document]
matt-o-how: Moved and updated

crates/chia-consensus/src/merkle_tree.rs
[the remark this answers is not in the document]
matt-o-how: Ok, done

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -81,7 +81,7 @@ fn deserialize_recurse(
             let hash: [u8; 32] = proof[pos + 1..pos + 33].try_into().map_err(|_| SetError)?;
             merkle_tree.leaf_vec.push(hash);
             merkle_tree.nodes_vec.push(ArrayTypes::Leaf {
-                data: merkle_tree.leaf_vec.len() - 1,
+                data: merkle_tree.leaf_vec.len() as u32 - 1 ,
```
arvidn, line 84:
```suggestion
                data: merkle_tree.leaf_vec.len() as u32 - 1,
```

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -123,56 +123,60 @@ fn deserialize_recurse(
 }
 
 impl MerkleSet {
+    pub fn new(leafs: &mut [[u8; 32]]) -> MerkleSet {
```
arvidn, line 126: I don't think it makes sense to just add another (unused) function, and more tests for this function. If you would turn `generate_merkle_tree()` into this function, you'd just be renaming it. No need for any new tests

crates/chia-consensus/src/merkle_set.rs
matt-o-how: done

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -0,0 +1,982 @@
...
+        assert_eq!(tree.leaf_vec.len(), 5);
+        if let ArrayTypes::Middle { children } = tree.nodes_vec[tree.nodes_vec.len() - 1] {
+            if let ArrayTypes::Leaf { data } = tree.nodes_vec[children.1 as usize] {
+                assert_eq!(tree.leaf_vec[data as usize], d);
+            } else {
+                assert!(false) // node should be a leaf
+            }
+        } else {
+            assert!(false) // root node should be a Middle
+        }
```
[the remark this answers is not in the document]
arvidn: I think I mixed up the `panic!()` messages in my suggestion, but you get the gist

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -0,0 +1,982 @@
...
+        Err(SetError)
+    } else {
+        Ok(merkle_tree)
+    }
+}
+
+fn deserialize_recurse(
+    merkle_tree: &mut MerkleSet,
+    proof: &[u8],
+    pos: usize,
```
arvidn, line 66: I think it would be better to pass in a sub-slice of `proof` when recursing, and not pass `pos` at all. The return value could then be the number of bytes consumed by the call, and you add it up as you return from the recursion

arvidn (approved):

crates/chia-consensus/src/merkle_set.rs
arvidn: I can add these tests later

PR #456, opened 2024-04-05
Mypy and test fixes, black formatting

Rigidity:
This fixes mypy errors, type stubs, and formats all of the test files with black. Black formatting is checked in CI as well.

You can review one commit at a time, and mostly ignore the "Black formatting" commit, since no changes were made in it other than formatting and adding CI to enforce the formatting.


wheel/chia_rs.pyi
```diff
@@ -55,19 +55,19 @@ NO_UNKNOWN_OPS: int = ...
...
 def run_chia_program(
     program: bytes, args: bytes, max_cost: int, flags: int
-) -> Pair[int, LazyNode]: ...
+) -> Tuple[int, LazyNode]: ...
 
 class LazyNode:
-    def pair() -> Optional[Tuple[LazyNode, LazyNode]]: ...
-    def atom() -> bytes: ...
+    pair: Optional[Tuple[LazyNode, LazyNode]]
+    atom: bytes
```
arvidn, line 62: actually, it looks like this should be `Optional[bytes]`. Did some tool detect this and say it should be just `bytes`?

wheel/chia_rs.pyi
```diff
@@ -91,11 +91,8 @@ class G1Element:
     def get_fingerprint(self) -> int: ...
     def pair(self, other: G2Element) -> GTElement: ...
     @staticmethod
-    def from_bytes_unchecked(b: bytes) -> G1Element: ...
-    @staticmethod
     def generator() -> G1Element: ...
     def __str__(self) -> str: ...
-    def __repr__(self) -> str: ...
```
arvidn, line 98: are all the `__repr__()` and `from_bytes_unchecked()` that you're removing duplicates?

arvidn (commented):
it all looks good, except the type of the `atom` property should be `Optional[bytes]` as far as I can tell

wheel/chia_rs.pyi
Rigidity: yes

wheel/chia_rs.pyi
Rigidity: ok I'll fix that in a bit, the generate type stubs script hard codes this (another reason automatic generation would be nice for everything later)

arvidn (approved):

PR #458, opened 2024-04-11
NIL conditions

arvidn:
The items in the conditions list is expected to contain lists where the first item is the condition op-code.

This PR adds a feature where the list is allowed to be empty (i.e. a NIL condition). Allowing this is a hard fork, so the feature is disabled by default and enabled by a new flag, `ENABLE_NIL_CONDITIONS`. It's currently not enabled anywhere.

The rationale for this is to easily be able to inject calls to, for example, `secp256k1_verify` or `bls_verify` in the condition list. These calls return `NIL` or raise an exception. So treating NILs in the condition lists as no-ops simplifies this.

The alternatives are to filter the condition list to remove the NILs or to wrap them in the `REMARK` condition.


AmineKhaldi (dismissed):

Rigidity (approved):

richardkiss:
This is a hard fork. Maybe the idea is to hide this behind a future hard fork. But I don't actually think it's a good idea, as it would break a lot of assumptions in wallet code... assumptions that wouldn't actually be exercised until we add tests for it (and it would have to be a lot of new tests... essentially one for any code that attempts to parse conditions, including a LOT of third party code we have no control over). So I am quite strongly against this.

richardkiss:
> This is a hard fork. Maybe the idea is to hide this behind a future hard fork. But I don't actually think it's a good idea, as it would break a lot of assumptions in wallet code... assumptions that wouldn't actually be exercised until we add tests for it (and it would have to be a lot of new tests... essentially one for any code that attempts to parse conditions, including a LOT of third party code we have no control over). So I am quite strongly against this.

Maybe I've misunderstood this? On further review, it seems like this just adds a new `REM`-type condition with condition code 0 instead of 1. I've run out of time at the moment, but I will look again later.

richardkiss:
Okay, I guess this seem innocuous. I don't completely understand why this is necessary since it doesn't change any results, but I suppose based on your description above is it hints at how it might make future changes simpler.

Rigidity:
This is already allowed on-chain and a farmer could include a spend that outputs a nil condition today. This just allows it in the mempool validation logic, which would have no backwards incompatibility, so it's not a hard fork.

richardkiss:
> @richardkiss this is not a hard- or soft fork. Any unknown conditions have always been no-ops in consensus mode (but disallowed in mempool mode). This includes `nil`. `parse_opcode()` would just return `None` in this case, and then it would fail or continue depending on whether we're on mempool mode or not.
> 
> The idea is simply to "bless" `nil` as a no-op, and make it allowed in mempools.
> 
> The rationale is to make puzzles that use secp or custom BLS validation (and possibly future soft-forked validation functions) easier to use and cost fewer bytes and CLVM cost in puzzles.
> 
> [This](https://github.com/Chia-Network/chia-blockchain/blob/29ddd41beefff68d8c94fb077447f0b26af4acfd/chia/wallet/puzzles/p2_delegated_or_hidden_secp.clsp#L12) is a real world example. i.e. use the form:
> 
> ```
> (if (secp256r1_verify SECP_PK (sha256 delegated_puzzle_hash coin_id GENESIS_CHALLENGE HIDDEN_PUZZLE_HASH) signature)
>     (x)  ; this doesn't actually run because secp256_verify will raise on failure
>     (c ... )
> )
> ```
> 
> Another alternative would be to write it something like this (assuming I counted my parentheses right):
> 
> ```
> (c (REMARK (secp256r1_verify SECP_PK (sha256 delegated_puzzle_hash coin_id GENESIS_CHALLENGE HIDDEN_PUZZLE_HASH) signature))
>     (c ... )
> )
> ```
> 
> With this patch, it can simply be written as:
> 
> ```
> (c (secp256r1_verify SECP_PK (sha256 delegated_puzzle_hash coin_id GENESIS_CHALLENGE HIDDEN_PUZZLE_HASH) signature)
>     (c ... )
> )
> ```
> 
> This pattern also applies to the `secp256k1_verify` and `bls_verify`.

Maybe I have this wrong, but this patch does *not* enable this pattern. There's a difference between

EX1 `((condition-code-1 v1 v2) (0 x1 x2))`

and

EX2 `((condition-code-1 v1 v2) 0)`

The first thing EX1 is okay. The second thing EX2 is not because the lone `0` is not a cons. If `secp256r1_verify` is to return a `nil`/`0` value (and I'm assuming it would because then the `secp256r1` change would be a simple soft fork), you can't just drop in a `nil` into a condition list, and this patch wouldn't change that fact. But it's possible I'm missing something...

richardkiss:
My understanding is the condition list has to be a list of cons boxes, ie. a list of lists, each of which is a condition, and having an atom (including `nil`) in the condition list is an error. That's always been my working assumption.

Rigidity:
Yeah good point, in the condition parsing code it does appear to early exit if the condition is not a cons.

Rigidity:
so I suppose that means that
`(c 1 (secp256k1_verify ...))`
or with this PR
`(c () (secp256k1_verify ...))`
would be the proper form for this

arvidn:
> My understanding is the condition list has to be a list of cons boxes, ie. a list of lists, each of which is a condition, and having an atom (including nil) in the condition list is an error. That's always been my working assumption.

`nil` *is* a valid list though. it's the empty list.

a list of empty lists looks like this:

```
(0 . (0 . (0 . 0)))
```

understanding that `0` is the same as `()` or `""`.

richardkiss:
> > My understanding is the condition list has to be a list of cons boxes, ie. a list of lists, each of which is a condition, and having an atom (including nil) in the condition list is an error. That's always been my working assumption.
> 
> `nil` _is_ a valid list though. it's the empty list.
> 
> a list of empty lists looks like this:
> 
> ```
> (0 . (0 . (0 . 0)))
> ```
> 
> understanding that `0` is the same as `()` or `""`.

The idea of `list` is more of a conceptual idea, I suppose. It's really a union type of this form:

`list = 0 | cons x list` ie. a `list` is recursively defined as either a `nil` or a cons of anything with a `list`.

So what I think of as a condition (and what I assume conditions are when I've written my limited amount of wallet/condition parsing code) is really a non-empty list:

`condition = cons CONDITION_CODE list` where `list` is as above

An empty `nil` list is missing the `CONDITION_CODE` so I assumed it would be an error, ie. not pass the consensus condition parsing. That's what I had in my head. What consensus *actually* does, I'm not sure, but I think it makes sense to just have it keep doing what it's doing, and if it *does* allow `nil` as a condition (ie. there is no condition code), we probably ought to pump out some tests.

richardkiss:
(To be clear, I have no problem with `nil`/`0`/`()` as a condition code. I'm just not crazy about the idea of it being a condition. But I don't think this patch affects that.)

arvidn:
without this patch. the test I added passes in consensus mode and fails in mempool mode:

https://github.com/Chia-Network/chia_rs/pull/458/files#diff-c93029cf40a6e979ca6c9b02528cfc71cad2e7099818a2deeb92cb9cf5cf344bR2314

However, now that I think about it, it's not immediately obvious why. you'd think it's the `first(a, c)?` that fails, because `c` is `nil`. In which case it *would* be a consensus failure as well.

I'll investigate some more.

arvidn:
it turned out my test was wrong. Supporting NIL conditions is a hard fork. This PR has been updated to add a flag that enable support for it, ready to be enabled if we have another hard fork

richardkiss:
Okay, so if we're talking that `0`/`()` should be a valid condition, I strongly oppose this. It will break a lot of code that assumes that each condition that come out of running the generator will have a condition code, an assumption this hard fork makes incorrect. There is a lot of existing code, including third party code and even clvm code which cannot be updated which makes this assumption. For example, I suspect the CAT code iterates over conditions to look for announcements that it needs to morph and will almost certainly barf on a `nil` condition, which means inner puzzles that returns nil conditions can't be used in a cat (and if they always return a nil condition, could brick the cat coin).

IMO this fork gives us opportunity to shoot ourselves in the foot in a lot of different ways for not much benefit.

arvidn:
I'm not sure if it's worth pursuing anything down this path anymore

richardkiss:
> I'm not sure if it's worth pursuing anything down this path anymore

Probably not.

PR #460, opened 2024-04-15
Move `PySpend` and `PySpendBundleConditions` to `chia-consensus` and rename to `Owned*`

Rigidity:


wheel/src/run_generator.rs
```diff
@@ -117,8 +66,8 @@ fn convert_spend(a: &Allocator, spend: Spend) -> PySpend {
 pub fn convert_spend_bundle_conds(
```
arvidn, line 66: these conversion functions shouldn't be part of the `wheel` anymore either now, should they?
Also, would there be any benefits to make these implementations of the `From` trait instead?

arvidn (approved):

PR #466, opened 2024-04-16
Block prevalidation port

Rigidity:
This is an early WIP of porting block prevalidation to Rust. Only signature validation and some other scaffolding has been done so far.


Rigidity:
This is super out of date and will be replaced with individual PRs for each part of the port.

PR #473, opened 2024-04-18
Add leaf audit to proof deserialization

matt-o-how:
Creates a stack of bits vecs and pops them to match the flow of the while loop of deserialisation.

### DIFF
--- a/crates/chia-consensus/src/merkle_tree.rs
+++ b/crates/chia-consensus/src/merkle_tree.rs
@@ -665,7 +665,7 @@ mod tests {
         let mut array: [u8; 32] = [0; 32];
         rng.fill(&mut array);
         random_data.push(array);
-        
+
         let mut bad_proof: Vec<u8> = Vec::new();
         bad_proof.push(MIDDLE);
         bad_proof.push(TRUNCATED);
@@ -676,7 +676,7 @@ mod tests {
         bad_proof.push(TERMINAL);
         bad_proof.extend_from_slice(&[0x0; 32]);
         let rebuilt = MerkleSet::from_proof(&bad_proof[0..bad_proof.len()]);
-        assert!(matches!(rebuilt, Err(SetError)));  // this is failing the audit
+        assert!(matches!(rebuilt, Err(SetError))); // this is failing the audit
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -77,93 +77,114 @@ impl MerkleSet {
...
-                        }
-                        TRUNCATED => {
-                            let mut hash = [0; 32];
-                            proof.read_exact(&mut hash).map_err(|_| SetError)?;
-                            values.push((self.nodes_vec.len() as u32, NodeType::Mid));
-                            self.nodes_vec.push((ArrayTypes::Truncated, hash));
-                        }
-                        MIDDLE => {
-                            if depth > 256 {
+            if let Some(bits) = bits_stack.pop() {
```
arvidn, line 84: I think it would be a lot nicer to say:
```suggestion
            let Some(bits) = bits_stack.pop() else {
                return Err(SetError);
            }
```

arvidn: it saves an indentation level for a lot of code

arvidn (commented):
I think this could a done with a single (mutable) `Vec<bool>` since we're doing a depth-first search. But it's fine for now. I can look into improving performance later.

However, I think we need some tests exercising proofs that are valid except failing this audit check.

crates/chia-consensus/src/merkle_tree.rs
```diff
@@ -631,6 +652,33 @@ mod tests {
...
+        rng.fill(&mut array);
+        random_data.push(array);
+
+        let mut bad_proof: Vec<u8> = Vec::new();
+        bad_proof.push(MIDDLE);
+        bad_proof.push(TRUNCATED);
+        bad_proof.extend_from_slice(&random_data[0]);
+        bad_proof.push(MIDDLE);
+        bad_proof.push(TERMINAL);
+        bad_proof.extend_from_slice(&[0x1; 32]); // this ought to be on the right
```
arvidn, line 675: I don't think this comment is right. The two leafs you have start with `0b00000000` and `0b00000001`. i.e. they share the first 7 bits. The proof is expected to have 8 layers. This is a valid test (that proofs must be complete), but I don't think it tests what you meant to test.

arvidn (approved):

PR #479, opened 2024-04-22
Port UnfinishedHeaderBlock to Rust

Rigidity:
From https://github.com/Chia-Network/chia-blockchain/blob/698455ee7ebe8bba16c2acadd0114de5788a9389/chia/types/unfinished_header_block.py


crates/chia-protocol/src/header_block.rs
```diff
@@ -72,6 +73,18 @@ impl HeaderBlock {
     pub fn first_in_sub_slot(&self) -> bool {
         !self.finished_sub_slots.is_empty()
     }
+
+    pub fn into_unfinished_header_block(self) -> UnfinishedHeaderBlock {
```
arvidn, line 77: I don't think there's a counterpart to this in `chia-blockchain`, is there?

arvidn (commented):
looks good. It's not clear what the new conversion function is intended to be used for however.

crates/chia-protocol/src/header_block.rs
Rigidity: It's to do this, this helper function isn't needed in the Python code, but it makes sense for the Rust port imo:
https://github.com/Chia-Network/chia-blockchain/blob/main/chia/consensus/block_header_validation.py#L842

Rigidity: That's why I didn't write a binding for it

arvidn (approved):

PR #482, opened 2024-04-22
Make ErrorCode comprehensive

Rigidity:
Here is a permalink to the `chia-blockchain` error codes I copied these from:

https://github.com/Chia-Network/chia-blockchain/blob/698455ee7ebe8bba16c2acadd0114de5788a9389/chia/util/errors.py

I used an LLM to make copying these over less tedious. So for that reason in addition to normal human error, careful review would be appreciated.


crates/chia-consensus/src/gen/validation_error.rs
```diff
@@ -95,39 +204,145 @@ pub fn first(a: &Allocator, n: NodePtr) -> Result<NodePtr, ValidationErr> {
 impl From<ErrorCode> for u32 {
     fn from(err: ErrorCode) -> u32 {
         match err {
```
arvidn, line 206: if we make this error code type hold the union of all python and rust errors, maybe we don't even need this conversion function. Or maybe we can just relying on the error value just being the rank of the error code list.

crates/chia-consensus/src/gen/validation_error.rs
```diff
@@ -5,47 +5,156 @@ use thiserror::Error;
...
+    InvalidIccHashCc,
+    InvalidIccHashRc,
+    InvalidIccEosVdf,
+    InvalidSpIndex,
+    TooManyBlocks,
+    InvalidCcChallenge,
+    InvalidPrefarm,
+    AssertSecondsRelativeFailed,
+    BadCoinbaseSignature,
+    // InitialTransactionFreeze (removed)
```
arvidn, line 125: if you want to maintain the same enum values, you still need a place holder here

crates/chia-consensus/src/gen/validation_error.rs
```diff
@@ -95,39 +204,145 @@ pub fn first(a: &Allocator, n: NodePtr) -> Result<NodePtr, ValidationErr> {
...
+            ErrorCode::InvalidIccHashCc => 98,
+            ErrorCode::InvalidIccHashRc => 99,
+            ErrorCode::InvalidIccEosVdf => 100,
+            ErrorCode::InvalidSpIndex => 101,
+            ErrorCode::TooManyBlocks => 102,
+            ErrorCode::InvalidCcChallenge => 103,
+            ErrorCode::InvalidPrefarm => 104,
+            ErrorCode::AssertSecondsRelativeFailed => 105,
+            ErrorCode::BadCoinbaseSignature => 106,
+            ErrorCode::NoTransactionsWhileSyncing => 108,
```
arvidn, line 320: I suppose you compensate by skipping 107 here. I think you should add a comment here too though. It looks like a mistake and a future reader of the code may "fix" it.

crates/chia-consensus/src/gen/validation_error.rs
Rigidity: This is commented in chia-blockchain as well.

crates/chia-consensus/src/gen/validation_error.rs
Rigidity: Sure

arvidn (approved):

PR #483, opened 2024-04-22
Make chia-protocol modules private

Rigidity:


arvidn (approved):

crates/chia-consensus/src/gen/get_puzzle_and_solution.rs
```diff
@@ -1,5 +1,5 @@
 use crate::gen::validation_error::{atom, check_nil, first, next, rest, ErrorCode, ValidationErr};
-use ::chia_protocol::bytes::Bytes32;
+use ::chia_protocol::Bytes32;
```
arvidn, line 2: are the leading `::` necessary here?

arvidn (approved):

PR #491, opened 2024-04-26
Add clippy to all targets

Rigidity:


.github/workflows/build-test.yml
```diff
@@ -305,11 +305,7 @@ jobs:
 
       - name: workspace
         run: |
-            cargo clippy --workspace --all-features -- -Dwarnings
-      - name: wheel
```
Rigidity, line 309: `wheel` is now part of the workspace, so this is redundant.

arvidn (approved):

PR #499, opened 2024-04-29
Add OwnedBytes and SizedBytes to clvm-utils

Rigidity:


arvidn (commented):
I would have expected this crate to contain *traits*, and that the concrete implementations would be defined in some other crate

Rigidity:
> I would have expected this crate to contain *traits*, and that the concrete implementations would be defined in some other crate

Should these go in `clvm-utils`?

Rigidity:
These might be too redundant with `chia-protocol`, I'll hold off on this for now.

PR #500, opened 2024-04-30
expose aggregate_verify_gt() in chia-bls

arvidn:
to prepare for the BLS cache.

Validating an aggregate signature in BLS can be broken down into the following steps:

1. for each (public key, message)-pair
1.1 augment the message by pre-pending the public key. The public key is also known as G1.
1.2 map the augmented message to G2 (`hash_to_g2()`)
1.3 pair the G1 and G2 points to get a GT point
2. multiply all GT points together
3. pair the generator (a G1 point) with the signature (a G2 point)
4. if the aggregate GT point equals the signature pair, the signature is valid.

The pairing operation is quite expensive. As you can see from the benchmark, validating a signature against 1000 GT points is 2 orders of magnitude faster than validating 1000 (key, message)-pairs. The BLS cache caches the GT points to avoid having to compute the pairings multiple times.

The benchmark gives an indication on the speed-up of caching the pairings:

| benchmark | time |
| --- | --- |
| aggregate_verify_gt, small msg | 2.1471 ms |
| aggregate_verify, small msg | 349.57 ms |
| verify, small msg | 944.59 µs |
| verify, 4kiB msg | 949.05 µs |

### DIFF
--- a/crates/chia-bls/benches/verify.rs
+++ b/crates/chia-bls/benches/verify.rs
@@ -1,6 +1,8 @@
-use chia_bls::secret_key::SecretKey;
 use chia_bls::signature;
-use chia_bls::signature::sign;
+use chia_bls::{
+    aggregate_verify, aggregate_verify_gt, hash_to_g2, sign, GTElement, PublicKey, SecretKey,
+    Signature,
+};
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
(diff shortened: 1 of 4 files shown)


matt-o-how:
Looks good.
If you add some extra comments to the public facing functions on signature.rs then this will be perfect.

matt-o-how (approved):

PR #503, opened 2024-05-06
Add TreeHasher to simplify currying puzzle hashes

Rigidity:
Instead of having to special case untyped currying code with `curry_tree_hash` for each puzzle type, you can use `CurriedProgram` directly now. Here is an example:

```rs
let puzzle_hash = CurriedProgram {
    program: CAT_PUZZLE_HASH,
    args: CatArgs {
        mod_hash: CAT_PUZZLE_HASH.into(),
        tail_program_hash: CurriedProgram {
            program: GENESIS_BY_COIN_ID_TAIL_PUZZLE_HASH,
            args: GenesisByCoinIdTailArgs {
                genesis_coin_id: Bytes32::default(),
            },
        }
        .tree_hash()
        .into(),
        inner_puzzle: CurriedProgram {
            program: STANDARD_PUZZLE_HASH,
            args: StandardArgs {
                synthetic_key: PublicKey::default(),
            },
        },
    },
}
.tree_hash();
```

### DIFF
--- a/crates/chia-consensus/src/gen/get_puzzle_and_solution.rs
+++ b/crates/chia-consensus/src/gen/get_puzzle_and_solution.rs
@@ -46,7 +46,7 @@ pub fn get_puzzle_and_solution_for_coin(
         }
 
         let puzzle_hash = tree_hash(a, puzzle);
-        if puzzle_hash != find_ph.as_ref() {
+        if puzzle_hash != find_ph.into() {
             continue;
         }
 
--- a/crates/chia-puzzles/src/derive_synthetic.rs
+++ b/crates/chia-puzzles/src/derive_synthetic.rs
@@ -18,7 +18,7 @@ pub trait DeriveSynthetic {
     where
         Self: Sized,
     {
-        self.derive_synthetic_hidden(&DEFAULT_HIDDEN_PUZZLE_HASH)
+        self.derive_synthetic_hidden(&DEFAULT_HIDDEN_PUZZLE_HASH.to_bytes())
     }
 }
 
--- a/crates/chia-puzzles/src/puzzles/cat.rs
+++ b/crates/chia-puzzles/src/puzzles/cat.rs
@@ -1,7 +1,7 @@
 use chia_bls::PublicKey;
 use chia_protocol::{Bytes32, Coin};
 use clvm_traits::{FromClvm, ToClvm};
-use clvm_utils::{curry_tree_hash, tree_hash_atom};
+use clvm_utils::TreeHash;
 use hex_literal::hex;
 
 use crate::LineageProof;
@@ -49,26 +49,6 @@ pub struct CoinProof {
     pub parent_coin_info: Bytes32,
     pub inner_puzzle_hash: Bytes32,
     pub amount: u64,
-}
-
-pub fn cat_puzzle_hash(asset_id: Bytes32, inner_puzzle_hash: Bytes32) -> Bytes32 {
-    let mod_hash = tree_hash_atom(&CAT_PUZZLE_HASH);
-    let asset_id_hash = tree_hash_atom(&asset_id);
-    curry_tree_hash(
-        CAT_PUZZLE_HASH,
-        &[mod_hash, asset_id_hash, inner_puzzle_hash.to_bytes()],
-    )
-    .into()
-}
-
-pub fn everything_with_signature_asset_id(public_key: &PublicKey) -> Bytes32 {
-    let pk_tree_hash = tree_hash_atom(&public_key.to_bytes());
-    curry_tree_hash(EVERYTHING_WITH_SIGNATURE_TAIL_PUZZLE_HASH, &[pk_tree_hash]).into()
-}
-
-pub fn genesis_by_coin_id_asset_id(genesis_coin_id: Bytes32) -> Bytes32 {
-    let coin_id_hash = tree_hash_atom(&genesis_coin_id);
-    curry_tree_hash(GENESIS_BY_COIN_ID_TAIL_PUZZLE_HASH, &[coin_id_hash]).into()
 }
 
 /// This is the puzzle reveal of the [CAT2 standard](https://chialisp.com/cats) puzzle.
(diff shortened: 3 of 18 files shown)


crates/chia-puzzles/src/puzzles/offer.rs
```diff
@@ -45,11 +46,11 @@ pub const SETTLEMENT_PAYMENTS_PUZZLE_V1: [u8; 267] = hex!(
 ///
 /// **Warning:**
 /// It is recommended not to use settlement payments v1 for anything other than backwards compatibility (e.g. offer compression).
-pub const SETTLEMENT_PAYMENTS_PUZZLE_HASH_V1: [u8; 32] = hex!(
+pub const SETTLEMENT_PAYMENTS_PUZZLE_HASH_V1: TreeHash = TreeHash::new(hex!(
```
arvidn, line 49: what's the motivation behind changing the type to `TreeHash`?
Is it still possible to call `.to_bytes()` on a constant, given that `to_bytes()` consumes `self`?

crates/chia-puzzles/src/puzzles/cat.rs
```diff
@@ -250,77 +223,4 @@ mod tests {
         assert_puzzle_hash!(EVERYTHING_WITH_SIGNATURE_TAIL_PUZZLE => EVERYTHING_WITH_SIGNATURE_TAIL_PUZZLE_HASH);
         assert_puzzle_hash!(GENESIS_BY_COIN_ID_TAIL_PUZZLE => GENESIS_BY_COIN_ID_TAIL_PUZZLE_HASH);
     }
-
-    #[test]
-    fn curry_cat_tree_hash() {
```
arvidn, line 255: don't we still need these tests?

crates/chia-puzzles/src/derive_synthetic.rs
```diff
@@ -118,7 +118,7 @@ mod tests {
         for (index, hex) in hex_keys.iter().enumerate() {
             let key = intermediate
                 .derive_unhardened(index as u32)
-                .derive_synthetic(&DEFAULT_HIDDEN_PUZZLE_HASH);
+                .derive_synthetic(&DEFAULT_HIDDEN_PUZZLE_HASH.to_bytes());
```
arvidn, line 121: are there any circumstances where having this be a `TreeHash` improves something? e.g. ergonomics or type-safety

crates/chia-puzzles/src/puzzles/cat.rs
```diff
@@ -250,77 +223,4 @@ mod tests {
...
-        let actual_tree_hash = hex::encode(cat_puzzle_hash(
-            asset_id,
-            standard_puzzle_hash(&inner_args.synthetic_key),
-        ));
-
-        assert_eq!(expected_tree_hash, actual_tree_hash);
-    }
-
-    #[test]
-    fn curry_everything_with_signature() {
```
arvidn, line 288: I would think we would still want these tests, just using the new mechanism to compute the hashes

crates/clvm-utils/src/hash_encoder.rs
```diff
@@ -0,0 +1,167 @@
...
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct TreeHash([u8; 32]);
+
+impl TreeHash {
+    pub const fn new(hash: [u8; 32]) -> Self {
+        Self(hash)
+    }
+
+    pub const fn to_bytes(self) -> [u8; 32] {
```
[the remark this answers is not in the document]
Rigidity: I didn't put much thought into it - is it more performant to take a reference to self here?

crates/chia-puzzles/src/puzzles/cat.rs
Rigidity: The special cased functions for currying these puzzles are gone, so I removed the tests for them as well. We could technically test the use of `CurriedProgram` but it seemed redundant to me since it's the same as using the `Allocator` now, just with a different encoder.

And you could theoretically use it with all of the puzzles now. Open to other ideas for testing though

crates/chia-puzzles/src/derive_synthetic.rs
Rigidity: Yeah, the idea is you can pass a puzzle hash directly into a generic inner puzzle field, and it will act as pre-hashed (ie as if you put the preimage there instead). If it were `[u8; 32]` you could easily mistakenly curry in a list of 32 bytes instead of the hash as-is.

crates/chia-puzzles/src/puzzles/cat.rs
Rigidity: I can add them back if you'd like, but it might make sense to have more comprehensive tests for all of the puzzles if so, to be consistent. I got rid of these since there's nothing special with functions like `cat_puzzle_hash` now.

crates/chia-puzzles/src/puzzles/offer.rs
Rigidity: Preventing mistakes with passing `[u8; 32]` instead of `TreeHash` in a generic inner puzzle field

It seems like taking self as a reference works at compile time too, for what it's worth

crates/chia-tools/src/bin/run-spend.rs
```diff
@@ -135,7 +136,7 @@ fn print_puzzle_info(a: &Allocator, puzzle: NodePtr, solution: NodePtr) {
         return;
     };
 
-    match tree_hash(a, uncurried.program) {
```
arvidn, line 138: should `tree_hash()` return a `TreeHash`? it would seem to strengthen type-safety

arvidn (approved):

crates/chia-consensus/src/gen/get_puzzle_and_solution.rs
```diff
@@ -46,7 +46,7 @@ pub fn get_puzzle_and_solution_for_coin(
         }
 
         let puzzle_hash = tree_hash(a, puzzle);
-        if puzzle_hash != find_ph.as_ref() {
+        if puzzle_hash != find_ph.into() {
```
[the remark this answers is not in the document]
Rigidity: I think mod hashes in chia-wallet are the most important, but may be worth revisiting later

PR #513, opened 2024-05-14
Add standard puzzle and CAT constructors

Rigidity:


crates/chia-puzzles/src/puzzles/cat.rs
```diff
@@ -312,22 +345,17 @@ mod tests {
...
         let allocated_tree_hash = hex::encode(tree_hash(&a, curried_ptr));
 
-        let tree_hash = hex::encode(
-            CurriedProgram {
-                program: GENESIS_BY_COIN_ID_TAIL_PUZZLE_HASH,
-                args: GenesisByCoinIdTailArgs { genesis_coin_id },
-            }
-            .tree_hash(),
-        );
+        let tree_hash = hex::encode(GenesisByCoinIdTailArgs::curry_tree_hash(genesis_coin_id));
```
arvidn, line 358: `GenesisByCoinIdTailArgs` is now tied to the program puzzle hash `GENESIS_BY_COIN_ID_TAIL_PUZZLE_HASH`. I suppose this is on purpose, but does the name suffix `Args` still make sense?
I think it's soon time to document how these types fit together to build puzzles and puzzle hashes.

arvidn (approved):

crates/chia-puzzles/src/puzzles/cat.rs
Rigidity: Yeah that would be ideal - maybe I can follow up on these PRs with another that has some documented examples on how to use these puzzles?

Rigidity: If you have a better idea for the API than adding it as an associated function on the args, let me know

PR #514, opened 2024-05-14
Add NFT puzzle constructors

Rigidity:


crates/chia-puzzles/src/puzzles/nft.rs
```diff
@@ -24,6 +46,32 @@ pub struct NftStateLayerArgs<I, M> {
     pub inner_puzzle: I,
 }
 
+impl<I, M> NftStateLayerArgs<I, M> {
+    pub fn new(metadata: M, inner_puzzle: I) -> Self {
+        Self {
+            mod_hash: NFT_STATE_LAYER_PUZZLE_HASH.into(),
```
arvidn, line 52: I suppose we'd want to make `mod_hash` a `TreeHash` at some point, right?
(as well as `metadata_updater_puzzle_hash` and `launcher_puzzle_hash`)

crates/chia-puzzles/src/puzzles/nft.rs
```diff
@@ -57,6 +135,36 @@ pub struct NftRoyaltyTransferPuzzleArgs {
     pub trade_price_percentage: u16,
 }
 
+impl NftRoyaltyTransferPuzzleArgs {
+    pub fn new(
+        launcher_id: Bytes32,
+        royalty_puzzle_hash: Bytes32,
```
arvidn, line 141: is this not a `TreeHash`?

arvidn (approved):
looks reasonable. Can you think of any ways to test or validate that these types produce the correct puzzle hashes? It seems you have somewhat of a protocol (with the name `curry_tree_hash()`). Perhaps a macro could generate a test that ensures the `tree_hash()` call generates the same hash as generating the full puzzle and tree-hashing it.

It might even be possible to have tests that execute (generated) puzzles with their (generated) solutions and ensure we get the conditions back that we expect.

crates/chia-puzzles/src/puzzles/nft.rs
Rigidity: No, since these need to be encoded as atoms rather than the value of the tree hash's preimage.

crates/chia-puzzles/src/puzzles/nft.rs
Rigidity: It has to be encoded as the puzzle hash atom rather than the puzzle itself

PR #516, opened 2024-05-17
Bump pyo3 to 0.21.2

Rigidity:
I followed my editor's deprecation warnings and the [pyo3 migration guide](https://pyo3.rs/v0.21.0/migration).

### DIFF
--- a/crates/chia-bls/src/cached_bls.rs
+++ b/crates/chia-bls/src/cached_bls.rs
@@ -113,10 +113,14 @@ impl BLSCache {
         msgs: &pyo3::Bound<PyList>,
         sig: &Signature,
     ) -> PyResult<bool> {
-        let pks_r = pks.iter().map(|item| item.extract::<PublicKey>().unwrap());
+        let pks_r = pks
+            .iter()
+            .unwrap()
+            .map(|item| item.unwrap().extract::<PublicKey>().unwrap());
         let msgs_r = msgs
             .iter()
-            .map(|item| item.extract::<PyBackedBytes>().unwrap());
+            .unwrap()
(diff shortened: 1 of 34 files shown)


crates/chia-bls/src/cached_bls.rs
```diff
@@ -107,12 +109,18 @@ impl BLSCache {
...
+        pks: &pyo3::Bound<PyList>,
+        msgs: &pyo3::Bound<PyList>,
         sig: &Signature,
     ) -> PyResult<bool> {
-        let pks_r = pks.iter().map(|item| item.extract::<PublicKey>().unwrap());
-        let msgs_r = msgs.iter().map(|item| item.extract::<&[u8]>().unwrap());
+        let pks_r = pks
+            .iter()
+            .unwrap()
+            .map(|item| item.unwrap().extract::<PublicKey>().unwrap());
```
arvidn, line 119: I suppose there used to be an `unwrap()` here before your change, but now there are two. I would expect we prefer to raise a python exception instead of panic if something's wrong. Do you feel confident that `unwrap()` is the best choice?

arvidn (dismissed):

crates/chia-bls/src/cached_bls.rs
Rigidity: Didn't you say panics get converted to Python errors? I suppose it's just cleaner to return an actual error anyways though, I'll look into it

arvidn (approved):

crates/chia-bls/src/cached_bls.rs
arvidn: yeah, they do. but panics don't propagate gracefully through rust code

PR #517, opened 2024-05-17
Bump wasm-bindgen to 0.2.92

Rigidity:
Resolves the following warning about `wasm-bindgen`:

```
warning: the following packages contain code that will be rejected by a future version of Rust: wasm-bindgen v0.2.87
note: 
To solve this problem, you can try the following approaches:


- Some affected dependencies have newer versions available.
You may want to consider updating them to a newer version to see if the issue has been fixed.

wasm-bindgen v0.2.87 has the following newer versions available: 0.2.89, 0.2.90, 0.2.91, 0.2.92


- If the issue is not solved by updating the dependencies, a fix has to be
implemented by those dependencies. You can help with that by notifying the
maintainers of this problem (e.g. by creating a bug report) or by proposing a
fix to the maintainers (e.g. by creating a pull request):

  - wasm-bindgen@0.2.87
  - Repository: https://github.com/rustwasm/wasm-bindgen
  - Detailed warning command: `cargo report future-incompatibilities --id 2 --package wasm-bindgen@0.2.87`

- If waiting for an upstream fix is not an option, you can use the `[patch]`
section in `Cargo.toml` to use your own version of the dependency. For more
information, see:
https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section
        
note: this report can be shown with `cargo report future-incompatibilities --id 1`
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2500,9 +2500,9 @@ checksum = "9c8d87e72b64a3b4db28d11ce29237c246188f4f51057d65a7eab63b7987e423"
 
 [[package]]
 name = "wasm-bindgen"
-version = "0.2.87"
+version = "0.2.92"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7706a72ab36d8cb1f80ffbf0e071533974a60d0a308d01a5d0375bf60499a342"
+checksum = "4be2531df63900aeb2bca0daaaddec08491ee64ceecbee5076636a3b026795a8"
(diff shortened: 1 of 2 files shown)


Rigidity:
If you're wondering why I removed the `=`, it's because we now include `wasm` in the workspace, so it's included in `Cargo.lock`. Therefore this pin is redundant.

arvidn (approved):

PR #528, opened 2024-05-18
Improve API and test readability of bls cache

Rigidity:


crates/chia-bls/src/cached_bls.rs
```diff
@@ -1,253 +1,260 @@
...
+use sha2::{Digest, Sha256};
 
-#[cfg_attr(feature = "py-bindings", pyclass(name = "BLSCache"))]
-#[derive(Clone)]
-pub struct BLSCache {
-    cache: LruCache<[u8; 32], GTElement>, // LRUCache of hash(pubkey + message) -> GTElement
+use crate::{aggregate_verify_gt, hash_to_g2};
+use crate::{GTElement, PublicKey, Signature};
+
+/// This cache is a bit weird because it's trying to account for validating
```
Rigidity, line 10: This makes the comment attached to the bls cache itself as a doc comment, so anyone using the public API on the Rust side can see it in their editor or on docs.rs.

crates/chia-bls/src/cached_bls.rs
```diff
@@ -1,253 +1,260 @@
...
+/// mempool signatures versus block signatures. When validating block signatures,
+/// there's not much point in caching the pairings because we're probably not going
+/// to see them again unless there's a reorg. However, a spend in the mempool
+/// is likely to reappear in a block later, so we can save having to do the pairing
+/// again. So caching is primarily useful when synced and monitoring the mempool in real-time.
+#[cfg_attr(feature = "py-bindings", pyo3::pyclass(name = "BLSCache"))]
+#[derive(Debug, Clone)]
+pub struct BlsCache {
```
Rigidity, line 18: Renamed this to `BlsCache` to follow [the Rust convention](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case).

crates/chia-bls/src/cached_bls.rs
```diff
@@ -1,253 +1,260 @@
...
+    // sha256(pubkey + message) -> GTElement
+    cache: LruCache<[u8; 32], GTElement>,
 }
 
-impl Default for BLSCache {
+impl Default for BlsCache {
     fn default() -> Self {
-        Self::new(None)
+        Self::new(NonZeroUsize::new(50000).unwrap())
```
Rigidity, line 25: Moving the default logic to `Default`, so `new` doesn't have to be concerned with this.

crates/chia-bls/src/cached_bls.rs
```diff
@@ -1,253 +1,260 @@
...
-    >(
+    pub fn is_empty(&self) -> bool {
+        self.cache.is_empty()
+    }
+
+    pub fn aggregate_verify(
         &mut self,
-        pks: P,
-        msgs: M,
+        pks: impl IntoIterator<Item = impl Borrow<PublicKey>>,
```
Rigidity, line 46: Personal preference, `impl Trait` is syntax sugar for generics, so this doesn't change behavior at all.

crates/chia-bls/src/cached_bls.rs
```diff
@@ -1,253 +1,260 @@
...
-            self.cache.put(h, pairing.clone());
+            let hash: [u8; 32] = hasher.finalize().into();
+
+            let pairing = aug_hash.pair(pk.borrow());
+            self.cache.put(hash, pairing.clone());
             pairing
         });
-        agg_ver_gt(sig, iter)
+
+        aggregate_verify_gt(sig, iter)
```
Rigidity, line 76: I don't think we should alias an import if it's only used in one place, it's just confusing to readers

crates/chia-bls/src/cached_bls.rs
```diff
@@ -1,253 +1,260 @@
...
     }
 }
 
-// Python Functions
+#[cfg(feature = "py-bindings")]
```
Rigidity, line 80: This is a new thing I'm going to be moving our code to if it works out - it keeps all of the imports contained into a Python binding module rather than repeating `#[cfg(feature = "py-bindings")]` over and over.

Rigidity (commented):
Here's some context on the changes

crates/chia-bls/src/cached_bls.rs
arvidn: I think this comment could be more succinct and precise in describing what this class is meant to do. I don't think saying that it's "weird" helps describe this.
But, that's not why you're here :)

crates/chia-bls/src/cached_bls.rs
```diff
@@ -1,253 +1,260 @@
...
-                    Err(PyValueError::new_err(
-                        "Cannot have a cache size less than one.",
-                    ))
-                } else {
-                    Ok(Self::new(NonZeroUsize::new(p_size as usize)))
-                }
-            }
-            None => Ok(Self::default()),
+    #[pymethods]
+    impl BlsCache {
```
arvidn, line 93: I imagine this refers the `BlsCache` in the parent module because you `use super::*;` above, right?
Is the reason why you're allowed to add another `impl` block for this type here because you're still in the same file? Because normally a module that doesn't define the type isn't allowed to open an `impl` block for it, as I understand it.

arvidn (approved):
looks good to me. feel free to close resolve my comments. It's probably a good idea for Matt to have a chance to take a look as well, before landing.

crates/chia-bls/src/cached_bls.rs
Rigidity: You can write an impl for a struct anywhere within the same crate

crates/chia-bls/src/cached_bls.rs
matt-o-how: I disagree with this because BLS is an acronym, not a word

Rigidity: This is what the convention says though:
> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.

crates/chia-bls/src/cached_bls.rs
matt-o-how: Can you retain the `expect` and message instead of reverting to `unwrap`?

crates/chia-bls/src/cached_bls.rs
Rigidity: Other conventions seem to include this as well (for acronyms longer than 2 characters at least):
https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers

PR #531, opened 2024-05-21
extend BLS cache with items() and update()

arvidn:
to support the current use case of serializing and returning the cache entries from the TX validation worker process back to the main process


matt-o-how (approved):

Rigidity (commented):
Now you shouldn't have to use hacks like imports inside of functions or fully qualified import paths, since the Python bindings are in their own module

crates/chia-bls/src/cached_bls.rs
```diff
@@ -130,6 +133,37 @@ mod python {
         pub fn py_len(&self) -> PyResult<usize> {
             Ok(self.len())
         }
+
+        #[pyo3(name = "items")]
+        pub fn py_items(&self, py: pyo3::Python) -> PyResult<PyObject> {
+            use pyo3::prelude::*;
```
Rigidity, line 139: This is what I meant, but it's not that big of a deal (the `pyo3::Python` isn't necessary either technically)

Rigidity (approved):

crates/chia-bls/src/cached_bls.rs
arvidn: what do you mean by `pyo3::Python` isn't necessary? the `PyList::new_bound()` takes a `py` object so the parameter is required, right?

arvidn: regarding importing this at function scope; If I remove this line, I get this error:
```
error[E0599]: no method named `append` found for struct `pyo3::Bound<'_, PyList>` in the current scope
...
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
85  +     use pyo3::prelude::PyListMethods;
    |
```

If I import `PyListMethods` at the module level, I get this error:

```
error[E0277]: the `?` operator can only be applied to values that implement `Try`
   --> crates/chia-bls/src/cached_bls.rs:120:23
    |
120 |               let pks = pks
    |  _______________________^
121 | |                 .iter()?
    | |________________________^ the `?` operator cannot be applied to type `BoundListIterator<'_>`
    |
    = help: the trait `Try` is not implemented for `BoundListIterator<'_>`
```

Which is unrelated code, which apparently breaks when exposed to this trait.

I could change this line to just import the trait I need here. Maybe that's an improvement at least.

matt-o-how (approved):

Rigidity (approved):

PR #532, opened 2024-05-21
PrettyPrint encoder for CLVM

Rigidity:
Because `ClvmEncoder` only gets access to `encode_pair` invocations, it can't know if it's working with the start of a list, or one of the pairs in the middle or end. Therefore there's no easy way I know of to treat operators differently.

### DIFF
--- a/crates/clvm-traits/src/pretty_printer.rs
+++ b/crates/clvm-traits/src/pretty_printer.rs
@@ -1,0 +1,121 @@
+use num_bigint::BigInt;
+
+use crate::{ClvmEncoder, ToClvm};
+
+const PRINTABLE: &str = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!\"#$%&'()*+,-./:;<=>?@[]^_\\`{|}~ \t\r\n";
+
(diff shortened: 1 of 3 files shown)


crates/clvm-traits/src/pretty_printer.rs
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            return Ok(Pretty(BigInt::from_signed_bytes_be(bytes).to_string()));
+        }
+
+        if !self.atom_strings {
+            return Ok(Pretty(format!("0x{}", hex::encode(bytes))));
+        }
+
+        let text = String::from_utf8_lossy(bytes);
```
arvidn, line 60: it seems a bit scary to interpret the string as utf8 before ensuring all characters are printable

arvidn (commented):
nice. I suppose we don't care so much about performance for this, because allocating all those strings is not efficient.

crates/clvm-traits/src/pretty_printer.rs
Rigidity: The replacement character for invalid UTF-8 isn't in the printable list

Rigidity:
Its main purpose is for tests, though I'm not sure this is sufficient - might need some more thinking / work

PR #533, opened 2024-05-21
CHIA-595: Introduce a soft-fork flag to disallow infinity G1 points as public keys in `AGG_SIG_*` conditions

arvidn:
And validate public keys earlier, in parse_conditions().

This PR adds a new flag, `DISALLOW_INFINITY_G1`. When set, `AGG_SIG_*` conditions are not allowed to have infinity G1 public keys. Enabling this flag is a soft-fork. When activated, Chia's BLS behaves like the IETF specification.

Infinity G1 public keys don't truly behave like public keys. They always produce the infinity G2 signature, regardless of the message. They do not provide any security.

Currently, public keys are validated (i.e. parsed from raw bytes into a valid `PublicKey` object) close to the python binding, where the results from running a generator is about to be passed back to python. In order to validate keys for infinity, this conversion is moved closer into the parsing of conditions. When this flag is enabled, it lets the conditions parser fail earlier, in case an invalid public key is detected.


Rigidity (dismissed):
lgtm

richardkiss (approved):
Seems reasonable

PR #535, opened 2024-05-22
Add TEST_CONSTANTS to consensus constants

matt-o-how:


crates/chia-protocol/src/bytes.rs
```diff
@@ -44,6 +44,12 @@ impl Bytes {
     }
 }
 
+impl Bytes32 {
```
Rigidity, line 47: I think it would be fine to change the main `BytesImpl` constructor to be `const` instead of this

crates/chia-consensus/src/consensus_constants.rs
```diff
@@ -132,3 +132,71 @@ pub struct ConsensusConstants {
...
+        0xcc, 0xd5, 0xbb, 0x71, 0x18, 0x35, 0x32, 0xbf, 0xf2, 0x20, 0xba, 0x46, 0xc2, 0x68, 0x99,
+        0x1a, 0x3f, 0xf0, 0x7e, 0xb3, 0x58, 0xe8, 0x25, 0x5a, 0x65, 0xc3, 0x0a, 0x2d, 0xce, 0x0e,
+        0x5f, 0xbb,
+    ]),
+    genesis_pre_farm_pool_puzzle_hash: Bytes32::const_new([
+        0xd2, 0x3d, 0xa1, 0x46, 0x95, 0xa1, 0x88, 0xae, 0x57, 0x08, 0xdd, 0x15, 0x22, 0x63, 0xc4,
+        0xdb, 0x88, 0x3e, 0xb2, 0x7e, 0xde, 0xb9, 0x36, 0x17, 0x8d, 0x4d, 0x98, 0x8b, 0x8f, 0x3c,
+        0xe5, 0xfc,
+    ]),
+    genesis_pre_farm_farmer_puzzle_hash: Bytes32::const_new([
```
Rigidity, line 174: Would much prefer using the `hex!("...")` macro from the `hex-literal` crate for all of these `Bytes32` values

crates/chia-protocol/src/bytes.rs
matt-o-how: I initially tried this and had a problem with cargo not liking it due to the variable size of the array.

arvidn: in `BytesImpl`, `N` is not variable, it's a constant

crates/chia-consensus/src/consensus_constants.rs
```diff
@@ -132,3 +132,71 @@ pub struct ConsensusConstants {
...
+        0x3d, 0x87, 0x65, 0xd3, 0xa5, 0x97, 0xec, 0x1d, 0x99, 0x66, 0x3f, 0x6c, 0x98, 0x16, 0xd9,
+        0x15, 0xb9, 0xf6, 0x86, 0x13, 0xac, 0x94, 0x00, 0x98, 0x84, 0xc4, 0xad, 0xda, 0xef, 0xcc,
+        0xe6, 0xaf,
+    ]),
+    max_vdf_witness_size: 64,
+
+    mempool_block_buffer: 10,
+
+    max_coin_amount: ((1u128 << 64) - 1) as u64,
```
arvidn, line 183: there's `u64::MAX` you can use here instead

crates/chia-consensus/src/consensus_constants.rs
```diff
@@ -132,3 +132,71 @@ pub struct ConsensusConstants {
...
 }
+
+pub const TEST_CONSTANTS: ConsensusConstants = ConsensusConstants {
+    slot_blocks_target: 32,
+    min_blocks_per_challenge_block: 16,
+    max_sub_slot_blocks: 128,
+    num_sps_sub_slot: 64,
+    sub_slot_iters_starting: u64::pow(2, 27),
+
+    difficulty_constant_factor: u128::pow(2, 67),
```
arvidn, line 143: is this preferable to `1_u128 << 67`?

Rigidity (approved):
looks fine to me

crates/chia-consensus/src/consensus_constants.rs
matt-o-how: In the python DEFAULT_CONSTANTS we used `uint128(2**67)` so I kept it equivalent to that

crates/chia-protocol/src/bytes.rs
```diff
@@ -48,6 +48,12 @@ impl Bytes {
...
```
arvidn, line 51: shouldn't this be removed now?

crates/chia-protocol/src/bytes.rs
```diff
@@ -48,6 +48,12 @@ impl Bytes {
...
+    pub const fn const_new(data: [u8; 32]) -> Self {
+        BytesImpl(data)
+    }
+}
+
```
arvidn, line 56:
```suggestion
```

arvidn (approved):

PR #538, opened 2024-05-23
Use cargo workspaces for publish

Rigidity:


.github/workflows/build-crate-and-npm.yml
```diff
@@ -54,122 +54,13 @@ jobs:
...
-    - name: publish to crates.io if tagged (chia)
-      continue-on-error: true
+    - name: publish to crates.io if tagged
       if: startsWith(github.event.ref, 'refs/tags')
       env:
         CARGO_REGISTRY_TOKEN: ${{ secrets.cargo_registry_token }}
       run: |
-        cargo publish
+        cargo install cargo-workspaces --locked
+        cargo ws publish --publish-as-is --allow-dirty
```
arvidn, line 63: I seem to recall that `--allow-dirty` is risky. Does it not include the dirty files in the package?

.github/workflows/build-crate-and-npm.yml
```diff
@@ -54,122 +54,13 @@ jobs:
...
-    - name: publish to crates.io if tagged (chia)
-      continue-on-error: true
+    - name: publish to crates.io if tagged
       if: startsWith(github.event.ref, 'refs/tags')
       env:
         CARGO_REGISTRY_TOKEN: ${{ secrets.cargo_registry_token }}
       run: |
-        cargo publish
+        cargo install cargo-workspaces --locked
+        cargo ws publish --publish-as-is
```
arvidn, line 63: the only thing the readme has to offer about this option is: "Publish crates from the current commit without versioning". What does that mean? Surely the crates that are published will have versions.

arvidn:
this is a good improvement

.github/workflows/build-crate-and-npm.yml
Rigidity: Confusing wording, it just means it wont bump version automatically since we already have

arvidn (approved):

PR #541, opened 2024-05-29
Add functions for multithreaded validation to the Rust side.

matt-o-how:
[DRAFT PR]


crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,243 @@
...
+    max_cost: u64,
+    constants: ConsensusConstants,
+    height: u32,
+    syncing: bool,
+    cache: Arc<Mutex<BlsCache>>,
+) -> Result<(OwnedSpendBundleConditions, Duration), ErrorCode> {
+    let start_time = Instant::now();
+    let additional_data = constants.agg_sig_me_additional_data;
+    let program: BlockGenerator = simple_solution_generator(&spend_bundle)?;
+    let npcresult = get_name_puzzle_conditions(program, max_cost, true, height, constants);
```
arvidn, line 73: I think you could simplify this quite a bit by, instead of implementing the full `get_name_puzzle_conditions()`, take a look at `run_block_generator2()` ([here](https://github.com/Chia-Network/chia_rs/blob/main/crates/chia-consensus/src/gen/run_block_generator.rs#L113)). Now imagine you never have any block references (which we don't in the mempool) and also imagine that you already have all the spends as a list of puzzle-reveals and solutions (as we do in `spend_bundle`).

You don't actually need to generate a "simple solution" nor do you need to serialize and deserialize it.

You just need this section, but simplified, since you don't need to parse it out of the CLVM structure.

https://github.com/Chia-Network/chia_rs/blob/main/crates/chia-consensus/src/gen/run_block_generator.rs#L156-L192

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,243 @@
...
+    match npcresult {
+        Err(e) => return Err(e),
+        Ok(unwrapped) => {
+            let npcresult = unwrapped;
+            let (pks, msgs) = pkm_pairs(npcresult.clone(), additional_data.as_slice())?;
```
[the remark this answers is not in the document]
matt-o-how: Yeah, unfortunately or it moves into `pkm_pairs` and can't be returned later and `pkm_pairs` needs a clone to loop through or you get references that don't live

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,73 @@
...
+// }
+
+pub fn get_name_puzzle_conditions(
+    generator: BlockGenerator,
+    max_cost: u64,
+    mempool_mode: bool,
+    height: u32,
+    constants: ConsensusConstants,
+) -> Result<OwnedSpendBundleConditions, ErrorCode> {
+    let mut flags = get_flags_for_height_and_constants(height, constants.clone());
```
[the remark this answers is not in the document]
matt-o-how: I get a moved value error on line 47 if I don't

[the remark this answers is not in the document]
arvidn: the function should take this by reference, not by value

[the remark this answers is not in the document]
matt-o-how: Actually this works if Constants is a reference

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,73 @@
...
+//     error: Option<u16>,
+//     conds: Option<OwnedSpendBundleConditions>,
+// }
+
+pub fn get_name_puzzle_conditions(
+    generator: BlockGenerator,
+    max_cost: u64,
+    mempool_mode: bool,
+    height: u32,
+    constants: ConsensusConstants,
```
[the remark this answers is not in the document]
matt-o-how: Done

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,243 @@
...
+//                 Ok((new_spend, owned_conditions.0))
+//             },
+//             Err(e) => {
+//                 Err(e)
+//             }
+//         }
+//     }
+// }
+
+pub fn simple_solution_generator(bundle: &SpendBundle) -> Result<BlockGenerator, ErrorCode> {
```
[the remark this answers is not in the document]
matt-o-how: `build_generator` expects the tuple `(Coin, AsRef<[u8]>, AsRef<[u8]>)` so this function is needed to convert a SpendBundle into that form

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+    for spend in conditions.spends {
+        let condition_items_pairs = [
+            (AGG_SIG_PARENT, spend.clone().agg_sig_parent),
+            (AGG_SIG_PUZZLE, spend.clone().agg_sig_puzzle),
+            (AGG_SIG_AMOUNT, spend.clone().agg_sig_amount),
+            (AGG_SIG_PUZZLE_AMOUNT, spend.clone().agg_sig_puzzle_amount),
+            (AGG_SIG_PARENT_AMOUNT, spend.clone().agg_sig_parent_amount),
+            (AGG_SIG_PARENT_PUZZLE, spend.clone().agg_sig_parent_puzzle),
+            (AGG_SIG_ME, spend.clone().agg_sig_me),
```
arvidn, line 37: I don't think you should clone the `Spend` here. just store references in the array

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,136 @@
...
+    let mut result = msg.to_vec();
+    result.extend(addendum);
+    if let Some(additional_data) = agg_sig_additional_data.get(&opcode) {
+        result.extend(additional_data.clone());
+    }
+
+    result
+}
+
+fn u64_to_bytes(val: u64) -> Bytes {
```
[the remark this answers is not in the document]
arvidn: oh, I see, we use this for `amount` as well. never mind

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,137 @@
...
+use sha2::{Digest, Sha256};
+use std::collections::HashMap;
+
+pub fn pkm_pairs(
+    conditions: OwnedSpendBundleConditions,
+    additional_data: &[u8],
+) -> Result<(Vec<PublicKey>, Vec<Vec<u8>>), ErrorCode> {
+    let mut pks = Vec::<PublicKey>::new();
+    let mut msgs = Vec::<Vec<u8>>::new();
+    let disallowed = agg_sig_additional_data(additional_data);
```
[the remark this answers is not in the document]
matt-o-how: Well, they're only constant on the same network. I agree broadly, but we ought to be calculating it somewhere

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+    let block_program = solution_generator(spends);
+    match block_program {
+        Ok(bp) => {
+            let program = Program::from_bytes(bp.as_slice());
+            match program {
+                Ok(p) => Ok(BlockGenerator {
+                    program: p,
+                    generator_refs: Vec::<Program>::new(),
+                    block_height_list: Vec::<u32>::new(),
```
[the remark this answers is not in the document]
matt-o-how: Eventually though we're going to use `generator_refs` though right? They're for the sake of potential future optimisations

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,73 @@
...
+//     py_pickle
+// )]
+// #[streamable]
+// pub struct NPCResult {
+//     error: Option<u16>,
+//     conds: Option<OwnedSpendBundleConditions>,
+// }
+
+pub fn get_name_puzzle_conditions(
+    generator: BlockGenerator,
```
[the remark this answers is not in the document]
matt-o-how: This causes move problems in the for loop as Program cannot implement Copy

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -715,7 +716,7 @@ impl Spend {
 // spend bundle level, like reserve_fee and absolute time locks. Other
 // conditions are per spend, like relative time-locks and create coins (because
 // they have an implied parent coin ID).
-#[derive(Debug, Default)]
+#[derive(Default, Debug)]
```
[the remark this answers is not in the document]
Rigidity: Personally this is my preferred order:

Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, ToClvm, FromClvm

Open to other ideas though

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,137 @@
...
+};
+use crate::gen::owned_conditions::{OwnedSpend, OwnedSpendBundleConditions};
+use crate::gen::validation_error::ErrorCode;
+use chia_bls::PublicKey;
+use chia_protocol::Bytes;
+use chia_protocol::Coin;
+use sha2::{Digest, Sha256};
+use std::collections::HashMap;
+
+pub fn pkm_pairs(
```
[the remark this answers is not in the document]
Rigidity: If you return an iterator over Result you can collect it into a single Result, returning the first error collected

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,137 @@
...
+fn agg_sig_additional_data(agg_sig_data: &[u8]) -> HashMap<ConditionOpcode, Vec<u8>> {
+    let mut ret: HashMap<ConditionOpcode, Vec<u8>> = HashMap::new();
+    for code in [
+        AGG_SIG_PARENT,
+        AGG_SIG_PUZZLE,
+        AGG_SIG_AMOUNT,
+        AGG_SIG_PUZZLE_AMOUNT,
+        AGG_SIG_PARENT_AMOUNT,
+        AGG_SIG_PARENT_PUZZLE,
+    ] {
```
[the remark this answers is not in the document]
Rigidity: Isn't it represented the same way in memory? Arrays are contiguous

crates/chia-consensus/src/gen/condition_tools.rs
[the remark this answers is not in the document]
Rigidity: Maybe a static cache (with a proper lock) could work? Not sure if that's a good solution or not

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,137 @@
...
+use crate::gen::owned_conditions::{OwnedSpend, OwnedSpendBundleConditions};
+use crate::gen::validation_error::ErrorCode;
+use chia_bls::PublicKey;
+use chia_protocol::Bytes;
+use chia_protocol::Coin;
+use sha2::{Digest, Sha256};
+use std::collections::HashMap;
+
+pub fn pkm_pairs(
+    conditions: OwnedSpendBundleConditions,
```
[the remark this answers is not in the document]
Rigidity: In some cases it's a pain or impossible to pass the Allocator around when using multithreading - worth keeping in mind

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,243 @@
...
+use clvmr::{ENABLE_BLS_OPS_OUTSIDE_GUARD, ENABLE_FIXED_DIV};
+use std::time::{Duration, Instant};
+
+// currently in multiprocess_validation.py
+// called via blockchain.py from full_node.py when a full node wants to add a block or batch of blocks
+fn pre_validate_blocks_multiprocessing() {}
+
+// currently in multiprocess_validation.py
+// called in threads from pre_validate_blocks_multiprocessing
+fn batch_pre_validate_blocks() {}
```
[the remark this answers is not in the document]
Rigidity: Yeah I have been / will be working on these in a separate PR. I need some of the same prerequisites such as NPCResult so I may wait for this PR to land so I can rebase.

Alternatively this could be split up into multiple PRs

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,243 @@
...
+
+// currently in mempool_manager.py
+// called in full_node.py when adding a transaction
+pub fn pre_validate_spendbundle(
+    new_spend: SpendBundle,
+    max_cost: u64,
+    constants: ConsensusConstants,
+    peak_height: u32,
+    syncing: bool,
+    cache: Arc<Mutex<BlsCache>>,
```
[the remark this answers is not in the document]
Rigidity: If you need to update it, you can't pass it by immutable reference without interior mutability.

I would say something like `&mut MutexGuard` or `&Mutex` could work though depending on the needs. That way you don't need to pass it an `Arc`.

I would personally recommend using `parking_lot` (sync) or `tokio` (async) for `Mutex` since it can't be poisoned, but don't have strong feelings on it.

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,243 @@
...
+                    let data: Vec<(&PublicKey, &[u8])> = pks
+                        .iter()
+                        .zip(msgs.iter().map(|msg| msg.as_slice()))
+                        .collect();
+                    aggregate_verify(&spend_bundle.aggregated_signature, data)
+                } else {
+                    // if we're fully synced then use the cache
+                    cache.lock().unwrap().aggregate_verify(
+                        pks,
+                        msgs,
```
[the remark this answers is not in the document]
Rigidity: Maybe even IntoIterator for maximum flexibility?

crates/chia-consensus/src/multiprocess_validation.rs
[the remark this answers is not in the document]
Rigidity: Even if that's the case, we shouldn't add complexity that we don't yet need

crates/chia-consensus/src/lib.rs
```diff
@@ -6,5 +6,8 @@ pub mod error;
 pub mod fast_forward;
 pub mod gen;
 pub mod generator_rom;
+pub mod generator_types;
 pub mod merkle_set;
 pub mod merkle_tree;
+pub mod multiprocess_validation;
```
Rigidity, line 12: Imo we should name this something more clear, like `spend_bundle_validation`. We aren't using multi-processing anymore.

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,238 @@
...
+                .unwrap()
+                .aggregate_verify(pks, msgs, &spend_bundle.aggregated_signature)
+        }
+    } {
+        return Err(ErrorCode::InvalidSpendBundle);
+    }
+    Ok((npcresult, start_time.elapsed()))
+}
+
+// #[cfg(feature = "py-bindings")]
```
Rigidity, line 97: Is this going to be uncommented/removed later?

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,238 @@
...
+
+pub fn simple_solution_generator(bundle: &SpendBundle) -> Result<BlockGenerator, ErrorCode> {
+    let mut spends = Vec::<(Coin, Vec<u8>, Vec<u8>)>::new();
+    for cs in &bundle.coin_spends {
+        let puzzle_reveal = cs.puzzle_reveal.to_vec();
+        let solution_reveal = cs.solution.to_vec();
+        spends.push((cs.coin, puzzle_reveal, solution_reveal));
+    }
+    let block_program = solution_generator(spends);
+    match block_program {
```
Rigidity, line 139: Instead of nesting matches like this it's best to use `let`..`else` and early return errors, so there's less indentation.

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,74 @@
...
+        flags = flags | MEMPOOL_MODE
+    };
+    let mut block_args = Vec::<Vec<u8>>::new();
+    for gen in generator.generator_refs {
+        block_args.push(gen.to_vec());
+    }
+    let mut a2 = make_allocator(LIMIT_HEAP);
+    let sbc_result: Result<SpendBundleConditions, ValidationErr> =
+        if height >= constants.hard_fork_fix_height {
+            run_block_generator2::<_, EmptyVisitor>(
```
Rigidity, line 48: nit: you could conditionally assign these functions to a variable then call the variable as a function instead of duplicating this

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,74 @@
...
+        } else {
+            run_block_generator::<_, EmptyVisitor>(
+                &mut a2,
+                generator.program.into_inner().as_slice(),
+                &block_args,
+                max_cost,
+                flags,
+            )
+        };
+    match sbc_result {
```
Rigidity, line 64: Another case where things like `let`..`else`, `map_err`, and `?` can make this more readable

crates/chia-consensus/src/npc_result.rs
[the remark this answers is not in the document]
arvidn: you should not be copying or cloning this object

crates/chia-consensus/src/multiprocess_validation.rs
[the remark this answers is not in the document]
arvidn: I think the fact that python shares so much of the mempool validation with block validation is a net-increase of complexity. By separating them, the mempool validation can be a lot simpler and faster

crates/chia-consensus/src/multiprocess_validation.rs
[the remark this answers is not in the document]
arvidn: you can convert that in this function, but it's only one line. Something like this: `.iter().map(|cs| (cs.coin, cs.puzzle_reveal, cs.solution))`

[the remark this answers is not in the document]
arvidn: but as I mentioned in my last comment. I don't think you should be building a generator at all

crates/chia-consensus/src/gen/condition_tools.rs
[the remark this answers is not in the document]
arvidn: here, `result` is a `Vec<Vec<u8>>`, which is 8 heap allocations, one of which contains pointers to the other 7.
I think a better solution, though, is to extend `ConsensusConstants` to keep all the `agg_sig_*_additional_data` as separate fields.

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,243 @@
...
+    spend_bundle: &SpendBundle,
+    max_cost: u64,
+    constants: ConsensusConstants,
+    height: u32,
+    syncing: bool,
+    cache: Arc<Mutex<BlsCache>>,
+) -> Result<(OwnedSpendBundleConditions, Duration), ErrorCode> {
+    let start_time = Instant::now();
+    let additional_data = constants.agg_sig_me_additional_data;
+    let program: BlockGenerator = simple_solution_generator(&spend_bundle)?;
```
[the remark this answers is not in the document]
matt-o-how: Once again the lack of `copy` on SpendBundle due to #[streamable] means that the iterator would cause a move

crates/chia-consensus/src/generator_types.rs
```diff
@@ -0,0 +1,13 @@
+use chia_protocol::Program;
+use chia_streamable_macro::streamable;
+
+#[cfg(feature = "py-bindings")]
+#[cfg_attr(feature = "py-bindings", pyo3::pyclass(module = "chia_rs"))]
+#[streamable]
+pub struct BlockGenerator {
```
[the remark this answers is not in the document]
matt-o-how: I've made get_name_puzzle_conditions() take just the program, but I think we may still need the full generator later

crates/chia-consensus/src/gen/condition_tools.rs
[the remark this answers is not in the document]
arvidn: right, once we need to cross the thread boundary, we need to turn it into `Owned...`. But we should defer that to the last moment

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,96 @@
...
+    let env = a.nil();
+    let Reduction(clvm_cost, mut all_spends) = run_program(&mut a, &dialect, program, env, cost_left)?;
+
+    subtract_cost(&a, &mut cost_left, clvm_cost)?;
+    all_spends = first(&a, all_spends)?;
+    let mut ret = SpendBundleConditions::default();
+    let mut state = ParseState::default();
+    let mut cache = HashMap::<NodePtr, TreeHash>::new();
+
+    while let Some((spend, rest)) = a.next(all_spends) {
```
arvidn, line 61: you may have noticed that you start with a `SpendBundle` (a list of coins, puzzle reveals and solutions). You turn that into a block generator using `simple_solution_generator()`. In this function you then execute that generator, and right here you iterate over the coins, puzzle reveals and solutions.

What I was suggesting was that you skip the steps of creating a block generator and executing the generator to extract the spends. One way would be to pass in the `SpendBundle` directly to this function, but this name may be a bit misleading since it's used to run block generators on the python side. You could also just inline this function into the one place where you call it.

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        let Reduction(clvm_cost, conditions) =
+            run_program(&mut a, &dialect, puzzle, solution, cost_left)?;
+
+        subtract_cost(&a, &mut cost_left, clvm_cost)?;
+
+        let buf = tree_hash_cached(&a, puzzle, &HashSet::<NodePtr>::new(), &mut cache);
+        let puzzle_hash = a.new_atom(&buf)?;
+
+        process_single_spend::<EmptyVisitor>(
```
arvidn, line 75: when validating transactions for the mempool, we want to use the `MempoolVisitor` here, not the `EmptyVisitor`.
This is to perform some additional checks for `ELIGIBLE_FOR_DEDUP` and `ELIGIBLE_FOR_FF`.

See the current path:
https://github.com/Chia-Network/chia_rs/blob/main/wheel/src/run_generator.rs#L45-L49

crates/chia-bls/src/cached_bls.rs
```diff
@@ -78,6 +78,40 @@ impl BlsCache {
 
         aggregate_verify_gt(sig, iter)
     }
+
+    pub fn aggregate_verify_with_iter(
```
arvidn, line 82: rather than duplicating `aggregate_verify()`, I think you could just change it to take an iterator of pairs, instead of two iterators. basically, replace it with this function. Tests and the python bindings would have to be updated, but you'd need tests for this function anyway

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,93 @@
...
+//     py_uppercase,
+//     py_pickle
+// )]
+// #[streamable]
+// pub struct NPCResult {
+//     error: Option<u16>,
+//     conds: Option<OwnedSpendBundleConditions>,
+// }
+
+pub fn get_name_puzzle_conditions(
```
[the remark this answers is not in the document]
matt-o-how: Eventually we'll want to use it for other purposes though right? The mempool param isn't taking up a lot of overhead

crates/chia-bls/src/cached_bls.rs
matt-o-how: Done

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,93 @@
...
+    let mut ret = SpendBundleConditions::default();
+    let mut state = ParseState::default();
+    let mut cache = HashMap::<NodePtr, TreeHash>::new();
+
+    for coin_spend in spend_bundle.coin_spends.clone() {
+        // process the spend
+        let puz = node_from_bytes(&mut a, coin_spend.puzzle_reveal.as_slice())?;
+        let sol = node_from_bytes(&mut a, coin_spend.solution.as_slice())?;
+        let parent = a.new_atom(coin_spend.coin.parent_coin_info.as_slice())?;
+        let amount = a.new_atom(u64_to_bytes(coin_spend.coin.amount).as_slice())?;
```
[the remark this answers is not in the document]
matt-o-how: Done

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,281 @@
...
+            cache,
+        )?;
+        Ok(result)
+    }
+}
+
+// currently in mempool_manager.py
+// called in threads from pre_validate_spend_bundle()
+// returns (error, cached_results, new_cache_entries, duration)
+fn validate_clvm_and_signature(
```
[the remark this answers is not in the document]
matt-o-how: Is this not all the concern of the pybindings?

Duration is how Rust stores it so we can cast to Python in the binding, right?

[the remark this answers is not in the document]
arvidn: yeah, that's a good point. That's a good place to do the conversion

crates/chia-bls/src/cached_bls.rs
```diff
@@ -44,13 +43,13 @@ impl BlsCache {
...
-    pub fn aggregate_verify(
+    pub fn aggregate_verify<Pk: Borrow<PublicKey>, Msg: AsRef<[u8]>>(
         &mut self,
-        pks: impl IntoIterator<Item = impl Borrow<PublicKey>>,
-        msgs: impl IntoIterator<Item = impl AsRef<[u8]>>,
+        pks_msgs: impl IntoIterator<Item = (Pk, Msg)>,
         sig: &Signature,
-    ) -> bool {
-        let iter = pks.into_iter().zip(msgs).map(|(pk, msg)| -> GTElement {
+    ) -> (bool, HashMap<[u8; 32], GTElement>) {
```
arvidn, line 50: The python counterpart has a function that returns the content of the cache, rather than making `aggregate_verify()` return all the entries that were added. I can imagine a future, when we've replaced the worker processes with threads, where the BLS cache is truly shared between the block validation threads and the TX validation threads. At that point we no longer need to return the cached items.

Right now, asking for the items and inserting the items is done in the python binding. What do you think about moving those functions into the proper rust type instead?

Also, I think returning a `Vec<([u8; 32], GTElement)>` would be quite a lot cheaper too.

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,86 @@
...
+            constants.agg_sig_parent_puzzle_additional_data.as_slice(),
+        ]
+        .concat(),
+        AGG_SIG_ME => {
+            let coin: Coin = Coin::new(spend.parent_id, spend.puzzle_hash, spend.coin_amount);
+            [
+                coin.coin_id().as_slice(),
+                constants.agg_sig_me_additional_data.as_slice(),
+            ]
+            .concat()
```
arvidn, line 57: it would be a bit more efficient to `.extend()` the vector twice, rather than first concatenating the buffers, and then extend by the result. You get one more allocation the way you do it right now.

In fact, even better would be to allocate `Vec<u8>::with_capacity()` and specify the exact number of bytes, then `.extend()` the `msg` and the two or three parameters. Then you'd only have a single heap allocation.

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,86 @@
...
+        }
+        _ => Vec::<u8>::new(),
+    });
+
+    result
+}
+
+pub fn u64_to_bytes(val: u64) -> Bytes {
```
arvidn, line 65: this should have a unit test

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,86 @@
+use crate::consensus_constants::ConsensusConstants;
+use crate::gen::opcodes::{
+    ConditionOpcode, AGG_SIG_AMOUNT, AGG_SIG_ME, AGG_SIG_PARENT, AGG_SIG_PARENT_AMOUNT,
+    AGG_SIG_PARENT_PUZZLE, AGG_SIG_PUZZLE, AGG_SIG_PUZZLE_AMOUNT,
+};
+use crate::gen::owned_conditions::OwnedSpend;
+use chia_protocol::Bytes;
+use chia_protocol::Coin;
+
+pub fn make_aggsig_final_message(
```
arvidn, line 10: this should have a unit test. I would recommend `rstest` so you can parameterize.

crates/chia-consensus/src/generator_types.rs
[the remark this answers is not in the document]
arvidn: I think you can remove this now

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,509 @@
...
+        return Err(ErrorCode::InvalidSpendBundle);
+    }
+    Ok((npcresult, added, start_time.elapsed()))
+}
+
+pub fn get_flags_for_height_and_constants(height: u32, constants: &ConsensusConstants) -> u32 {
+    let mut flags: u32 = 0;
+    if height >= constants.soft_fork2_height {
+        flags |= NO_RELATIVE_CONDITIONS_ON_EPHEMERAL;
+    }
```
arvidn, line 105: this is in the pipeline of being removed as well. https://github.com/Chia-Network/chia-blockchain/pull/18135
In fact, in the latest version of `chia_rs`, this constant has been removed.

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,509 @@
...
+    #[test]
+    fn test_validate_aggsig_me() {
+        let sk_hex = "52d75c4707e39595b27314547f9723e5530c01198af3fc5849d9a7af65631efb";
+        let sk = SecretKey::from_bytes(&<[u8; 32]>::from_hex(sk_hex).unwrap()).unwrap();
+        //let pk: PublicKey = sk.public_key(); //0x997cc43ed8788f841fcf3071f6f212b89ba494b6ebaf1bda88c3f9de9d968a61f3b7284a5ee13889399ca71a026549a2
+        // panic!("{:?}", pk);
+
+        let full_puz = Bytes32::new(tree_hash_atom(&[1_u8]).to_bytes());
+        let test_coin = Coin::new(
+            hex::decode("4444444444444444444444444444444444444444444444444444444444444444")
```
arvidn, line 252: you can use `hex!()` here instead, since it's a string literal.

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,509 @@
...
+                .unwrap()
+                .try_into()
+                .unwrap(),
+            full_puz,
+            1,
+        );
+
+        let solution = "ffff32ffb0997cc43ed8788f841fcf3071f6f212b89ba494b6ebaf1bda88c3f9de9d968a61f3b7284a5ee13889399ca71a026549a2ff8568656c6c6f8080";
```
arvidn, line 260: I think all of these hex string would benefit from using `hex!()`, to make them binary blobs instead

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,509 @@
...
+            &spend_bundle,
+            TEST_CONSTANTS.max_block_cost_clvm,
+            &TEST_CONSTANTS,
+            TEST_CONSTANTS.hard_fork_height + 1,
+            Arc::new(Mutex::new(BlsCache::default())),
+        );
+        match result {
+            Ok(_) => return,
+            Err(e) => panic!("{:?}", e),
+        }
```
arvidn, line 367:
```suggestion
        validate_clvm_and_signature(
            &spend_bundle,
            TEST_CONSTANTS.max_block_cost_clvm,
            &TEST_CONSTANTS,
            TEST_CONSTANTS.hard_fork_height + 1,
            Arc::new(Mutex::new(BlsCache::default())),
        ).expect("validation is supposed to pass");
```

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,509 @@
...
+        let result = validate_clvm_and_signature(
+            &spend_bundle,
+            TEST_CONSTANTS.max_block_cost_clvm,
+            &TEST_CONSTANTS,
+            TEST_CONSTANTS.hard_fork_height - 1,
+            Arc::new(Mutex::new(BlsCache::default())),
+        );
+        if let Ok(_) = result {
+            panic!("height too low!")
+        };
```
arvidn, line 496:
```suggestion
        validate_clvm_and_signature(
            &spend_bundle,
            TEST_CONSTANTS.max_block_cost_clvm,
            &TEST_CONSTANTS,
            TEST_CONSTANTS.hard_fork_height - 1,
            Arc::new(Mutex::new(BlsCache::default())),
        ).expect("height too low!");
```

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,509 @@
...
+            &spend_bundle,
+            TEST_CONSTANTS.max_block_cost_clvm,
+            &TEST_CONSTANTS,
+            TEST_CONSTANTS.hard_fork_height + 1,
+            Arc::new(Mutex::new(BlsCache::default())),
+        );
+        match result {
+            Ok(_) => return,
+            Err(e) => panic!("{:?}", e),
+        }
```
arvidn, line 507: this could also use `.expect()`

crates/chia-consensus/src/npc_result.rs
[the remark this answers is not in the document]
arvidn: Most of the suggestions I've made to simplify this was based on the assumption that we would use a separate function for full blocks.

The cost of making this function work both for mempool (i.e. single spend bundles) and full blocks is:
1. We would need to turn the spend bundle into a block before calling this function (in order to have a uniform interface). This function would then need to run said block generator function to get back to the individual spends. As it is now, we can skip that part.
3. We would need to support both pre- and post- hard fork (since blocks need to validate with the pre-hard fork rules). Transactions going into the mempool only need to be validated with the post-hard fork rules however.

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,92 @@
...
+            flags,
+            &mut cost_left,
+            constants,
+        )?;
+    }
+
+    validate_conditions(&a, &ret, state, a.nil(), flags)?;
+
+    ret.cost = max_cost - cost_left;
+    let Ok(osbc) = OwnedSpendBundleConditions::from(&a, ret) else {
```
arvidn, line 88: if you rebase onto `main` now, this conversion cannot fail. This simplifies this code a bit.

wheel/src/api.rs
```diff
@@ -365,6 +371,47 @@ fn fast_forward_singleton<'p>(
...
+pub fn py_validate_clvm_and_signature(
+    new_spend: SpendBundle,
+    max_cost: u64,
+    constants: ConsensusConstants,
+    peak_height: u32,
+    cache: BlsCache,
+) -> PyResult<(
+    SpendBundle,
+    OwnedSpendBundleConditions,
+    HashMap<[u8; 32], GTElement>,
```
arvidn, line 385: I think it would be preferable to return a `Vec<>` here, rather than a hashmap. All we're doing with it is pickling, unpickling and adding it to the main BLS cache

wheel/src/api.rs
```diff
@@ -365,6 +371,47 @@ fn fast_forward_singleton<'p>(
...
+            HashMap<[u8; 32], GTElement>,
+            Duration,
+        ),
+        ErrorCode,
+    > = validate_clvm_and_signature(
+        &new_spend,
+        max_cost,
+        &constants,
+        peak_height,
+        Arc::new(Mutex::new(cache)),
```
arvidn, line 400: this will copy the entire BLS cache. I don't think that's practical

crates/chia-consensus/src/gen/condition_tools.rs
Rigidity: I am very worried about conversions like this here and in my own code. I think we should have extensive fuzz testing to compare the results against the Allocator too. Maybe out of scope of this PR though

Rigidity: The main concern is leading zeros being added or not, for example

matt-o-how: I agree and will implement some unit tests, but this code was adapted from `Coin.coin_id()` which is used in many critical places

Rigidity: Yeah if it's the same, probably not too urgent but better test coverage of that existing code would be good too in the future

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,467 @@
...
+    fn test_validate_no_pks() {
+        let test_coin = Coin::new(
+            Bytes32::new(hex!(
+                "4444444444444444444444444444444444444444444444444444444444444444"
+            )),
+            Bytes32::new(hex!(
+                "3333333333333333333333333333333333333333333333333333333333333333"
+            )),
+            1,
+        );
```
arvidn, line 159: I think you can simplify this to:
```suggestion
        let test_coin = Coin::new(
            hex!("4444444444444444444444444444444444444444444444444444444444444444").into(),
            hex!("3333333333333333333333333333333333333333333333333333333333333333").into(),
            1,
        );
```

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,467 @@
...
+            hex!(
+                "ff\
+ff33\
+ffa02222222222222222222222222222222222222222222222222222222222222222\
+ff01\
+80\
+80"
+            )
+            .to_vec(),
+        );
```
arvidn, line 171: I'm pretty sure you can leave this as a `&[u8]` and convert it directly to a `Program` below, with the `into()` conversion.

```suggestion
        let solution = hex!(
                "ff\
ff33\
ffa02222222222222222222222222222222222222222222222222222222222222222\
ff01\
80\
80"
        );
```

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,467 @@
...
+        let spend = CoinSpend::new(
+            test_coin,
+            Program::new(vec![1_u8].into()),
+            Program::new(solution),
```
arvidn, line 175:
```suggestion
            [1_u8].into(),
            solution.into(),
```

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,467 @@
...
+        let result = validate_clvm_and_signature(
+            &spend_bundle,
+            TEST_CONSTANTS.max_block_cost_clvm,
+            &TEST_CONSTANTS,
+            TEST_CONSTANTS.hard_fork_height - 1,
+            Arc::new(Mutex::new(BlsCache::default())),
+        );
+        if let Ok(_) = result {
+            panic!("height too low!")
+        };
```
arvidn, line 333: I could be wrong, but it looks like this would be simpler to just:

```suggestion
        assert!_eq(validate_clvm_and_signature(
            &spend_bundle,
            TEST_CONSTANTS.max_block_cost_clvm,
            &TEST_CONSTANTS,
            TEST_CONSTANTS.hard_fork_height - 1,
            Arc::new(Mutex::new(BlsCache::default())),
        ).unwrap_err(), <expected_error>);
```

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,467 @@
...
+        let _result = validate_clvm_and_signature(
```
arvidn, line 334:
```suggestion
        validate_clvm_and_signature(
```

wheel/src/api.rs
```diff
@@ -365,6 +371,45 @@ fn fast_forward_singleton<'p>(
...
+    ); // TODO: use cache properly
+    match sbc {
+        Ok((owned_conditions, additions, duration)) => Ok((
+            new_spend,
+            owned_conditions,
+            additions,
+            duration.as_secs_f32(),
+        )),
+        Err(e) => Err(PyErr::new::<PyTypeError, _>(e as u32)),
+    }
```
arvidn, line 410:
```suggestion
    let (owned_conditions, additions, duration) = validate_clvm_and_signature(
        &new_spend,
        max_cost,
        &constants,
        peak_height,
        Arc::new(Mutex::new(cache)),
    ).map_err(|e| PyErr::new::<PyTypeError, _>(e as u32))?; // TODO: use cache properly
    Ok((
        new_spend,
        owned_conditions,
        additions,
        duration.as_secs_f32(),
    )),
```

arvidn (commented):
I think there are two separate changes that could be pulled out of this PR, as independent ones.

* porting `get_flags_for_height_and_constants()` to rust and exposing it to python. This is in itself a pretty siginifcant change as it moves the logic that's used to define soft-fork activation.
* Updating the API to BLS cache to take an iterator of pairs rather than two separate iterators

crates/chia-bls/src/bls_cache.rs
```diff
@@ -42,13 +41,22 @@ impl BlsCache {
     pub fn is_empty(&self) -> bool {
         self.cache.is_empty()
     }
+    
+    pub fn update(&mut self, new_items: Vec<([u8; 32], GTElement)>) {
```
arvidn, line 45: I don't think you should take a vector by value. There's no need to move it in here. In fact, you probably want to take an iterator of `GTElement`, so the python binding can avoid generating an intermediate vector.

Also, as you add this, it probably would make sense to update the python binding to use these functions as well.

https://github.com/Chia-Network/chia_rs/blob/main/crates/chia-bls/src/bls_cache.rs#L139-L167

arvidn (commented):
I think it's a little bit odd to make the `aggregate_verify()` member function return the cache misses.
On the python side, we construct an empty `BLSCache` run `aggregate_verify()` and then pick out the cached items. This is not the most efficient way of doing it, but it's simple and reuses the BLS cache as the "local stash" of the pairings.

Your approach is more efficient. For example, we know that every single key will be a cache miss, so there's not point in looking it up in the cache. However, when validating blocks, we don't really want to do the extra work of returning the cache misses. So your approach would work best if we decide to use different functions for validating `SpendBundle`s and validating blocks. You could go the route of not using the BLS cache at all, in your patch. Instead, just implement a *free function* `aggregate_verify()` that doesn't use a cache but returns all the pairings it computed as a list.

This would improve performance some, and maybe simplify the two different code paths as well, since they don't need to share as much code.

In the long run, however, perhaps we would like to share the cache as well. I'm not sure though, maybe it's reasonable to aim for what's simplest now and figure out how to port the whole thread pool to rust when we get there/

crates/chia-consensus/src/multiprocess_validation.rs
matt-o-how: It needs the `to_vec`, but we can save the `Bytes::new()` with `.into()`

crates/chia-consensus/src/multiprocess_validation.rs
matt-o-how: This test has been removed now anyway as we removed the fork height

crates/chia-consensus/src/multiprocess_validation.rs
matt-o-how: Done

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,90 @@
...
+//     conds: Option<OwnedSpendBundleConditions>,
+// }
+
+pub fn get_name_puzzle_conditions(
+    spend_bundle: &SpendBundle,
+    max_cost: u64,
+    mempool_mode: bool,
+    height: u32,
+    constants: &ConsensusConstants,
+) -> Result<OwnedSpendBundleConditions, ValidationErr> {
```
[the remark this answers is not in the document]
matt-o-how: We need to translate lots of the Spend data out of `NodePtr`s anyway, and send it back to Python as an `OwnedSpendBundleConditions` so in my opinion this is not worth it

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,413 @@
...
...ff82095180ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095280ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095380ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095480ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095580ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095680ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095780ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095880ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095980ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095a80ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095b80ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095c80ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095d80ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095e80ffff33ffa09dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2ff82095f8080").to_vec();
```
arvidn, line 206: this is a very long line. Maybe this should be included from a binary file (or at least a separate file) instead. see `!include_str()` and `!include_bytes()` macros. I would like a comment to explain what it is also.

crates/chia-consensus/src/npc_result.rs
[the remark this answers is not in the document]
arvidn: Not a biggie; but the python boundary will keep moving away from this code. But we can address it then

crates/chia-consensus/src/npc_result.rs
```diff
@@ -0,0 +1,72 @@
...
+use clvmr::chia_dialect::LIMIT_HEAP;
+use clvmr::reduction::Reduction;
+use clvmr::run_program::run_program;
+use clvmr::serde::node_from_bytes;
+use std::collections::{HashMap, HashSet};
+
+pub fn get_name_puzzle_conditions(
+    spend_bundle: &SpendBundle,
+    max_cost: u64,
+    mempool_mode: bool,
```
arvidn, line 24: this parameter should be removed. You're only dealing with transactions, i.e. mempool mode

tests/test_blscache.py
```diff
@@ -1,11 +1,100 @@
...
+from chia_rs.sized_ints import uint8, uint16, uint32, uint64, uint128
 from typing import List
 from chia.util.hash import std_hash
 from chia.util.lru_cache import LRUCache
+from chia.types.blockchain_format.program import Program as ChiaProgram
 from chia.util import cached_bls as cached_bls_old
 import pytest
 
 
+DEFAULT_CONSTANTS = ConsensusConstants(
```
arvidn, line 25: it looks like this is only used by tests. I believe we already have a `TEST_CONSTANTS` that you could use. If that doesn't work, could you at least only define this when building with `#[cfg(test)]`?

wheel/generate_type_stubs.py
```diff
@@ -305,6 +305,22 @@ def confirm_not_included_already_hashed(
...
 
+def validate_clvm_and_signature(
+    new_spend: SpendBundle,
+    max_cost: int,
+    constants: ConsensusConstants,
+    peak_height: int,
+    cache: Optional[BLSCache],
+) -> Tuple[SpendBundleConditions, float]: ...
+
+def get_name_puzzle_conditions(
```
arvidn, line 316: you need to export the `get_flags_for_height_and_constants()` as well. We need that function on the python side.

wheel/generate_type_stubs.py
```diff
@@ -338,7 +354,7 @@ class BLSCache:
     def __init__(self, cache_size: Optional[int] = 50000) -> None: ...
     def len(self) -> int: ...
     def aggregate_verify(self, pks: List[G1Element], msgs: List[bytes], sig: G2Element) -> bool: ...
-    def items(self) -> List[Tuple[bytes, bytes]]: ...
+    def items(self) -> List[Tuple[bytes, GTElement]]: ...
```
arvidn, line 357: I don't think this change would work. We still pass this type across the python process boundary, so it needs to be pickled.

arvidn (changes requested):

tests/test_blscache.py
Rigidity: This is Python code - that said we may be able to export it as a binding for test purposes?

crates/chia-consensus/src/lib.rs
matt-o-how: Done

crates/chia-consensus/src/multiprocess_validation.rs
matt-o-how: Done

crates/chia-consensus/src/npc_result.rs
matt-o-how: Done

wheel/generate_type_stubs.py
matt-o-how: GTElement is a Rust pybinding in Python right? So passing it across should be straightforward. Or am I missing something?

crates/chia-consensus/src/multiprocess_validation.rs
matt-o-how: Done

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,467 @@
...
+            .flat_map(|(condition, items)| {
+                let spend = spend.clone();
+                items.iter().map(move |(pk, msg)| {
+                    (
+                        pk,
+                        make_aggsig_final_message(*condition, msg.as_slice(), &spend, constants),
+                    )
+                })
+            })
+            .collect::<Vec<_>>()
```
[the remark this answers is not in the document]
matt-o-how: Done

crates/chia-consensus/src/multiprocess_validation.rs
```diff
@@ -0,0 +1,281 @@
...
+use chia_bls::BlsCache;
+use chia_protocol::SpendBundle;
+use clvmr::{ENABLE_BLS_OPS_OUTSIDE_GUARD, ENABLE_FIXED_DIV};
+use std::sync::{Arc, Mutex};
+// use std::thread;
+use std::time::{Duration, Instant};
+
+// currently in mempool_manager.py
+// called in full_node.py when adding a transaction
+pub fn pre_validate_spendbundle(
```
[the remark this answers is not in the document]
matt-o-how: Removed

wheel/generate_type_stubs.py
arvidn: the way current python code is written, suggests that we have to manually convert objects into something that supports pickle.
looking at `crates/chia_py_streamable_macro/src/lib.rs`, it seems like we implement the pickle protocol in our bindings though. So, as far as I can tell, this ends up serializing the `GTElement` into bytes.
I don't believe we have any test coverage or round-trips of the pickle support though

wheel/src/api.rs
```diff
@@ -398,19 +398,20 @@ pub fn py_validate_clvm_and_signature(
 }
 
 #[pyfunction]
-#[pyo3(name = "get_name_puzzle_conditions")]
-pub fn py_get_name_puzzle_conditions(
+#[pyo3(name = "get_conditions_from_spendbundle")]
```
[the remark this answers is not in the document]
matt-o-how: Yes. It is required to fix a bug with the fast-forwarding code

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -0,0 +1,395 @@
+use crate::consensus_constants::ConsensusConstants;
```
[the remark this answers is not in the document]
matt-o-how: Should be fixed now

crates/chia-consensus/src/npc_result.rs
```diff
@@ -7,7 +7,7 @@ use crate::gen::flags::MEMPOOL_MODE;
 use crate::gen::owned_conditions::OwnedSpendBundleConditions;
 use crate::gen::run_block_generator::subtract_cost;
 use crate::gen::validation_error::ValidationErr;
-use crate::spend_bundle_validation::get_flags_for_height_and_constants;
+use crate::spendbundle_validation::get_flags_for_height_and_constants;
```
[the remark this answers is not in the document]
matt-o-how: Done

crates/chia-consensus/src/npc_result.rs
```diff
@@ -21,14 +21,12 @@ use std::collections::{HashMap, HashSet};
...
     max_cost: u64,
-    mempool_mode: bool,
     height: u32,
     constants: &ConsensusConstants,
 ) -> Result<OwnedSpendBundleConditions, ValidationErr> {
     let mut flags = get_flags_for_height_and_constants(height, constants);
-    if mempool_mode {
-        flags |= MEMPOOL_MODE;
-    };
+    flags |= MEMPOOL_MODE;
```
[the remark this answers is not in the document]
matt-o-how: Done

crates/chia-consensus/src/gen/condition_tools.rs
matt-o-how: Done

PR #554, opened 2024-05-31
Validate AGG_SIG_UNSAFE messages to not end in one of the banned suffixes

arvidn:
as part of the condition parsing.

The message of `AGG_SIG_UNSAFE` conditions are not allowed to end with any of the suffixes used by `AGG_SIG_*` conditions (those suffixes are derived from `ADDITIONAL_DATA`, i.e. a unique identifier for the network).

This is currently validated in `pkm_pairs()` in `chia-blockchain`. This PR moves this validation into the condition parsing, to reject invalid conditions earlier. This requires `ConsensusConstants` to be passed in to run_block_generator and run_puzzle.

In `chia-blockchain`, the suffixes for the `AGG_SIG_*` conditions are computed on-the-fly. This patch also move the agg_sig_*_additional_data into ConsensusConstants instead.

### DIFF
--- a/crates/chia-consensus/src/consensus_constants.rs
+++ b/crates/chia-consensus/src/consensus_constants.rs
@@ -74,8 +74,15 @@ pub struct ConsensusConstants {
     /// We override this value based on the chain being run (testnet0, testnet1, mainnet, etc).
     genesis_challenge: Bytes32,
 
-    /// Forks of chia should change this value to provide replay attack protection.
+    /// Forks of chia should change these values to provide replay attack protection.
     agg_sig_me_additional_data: Bytes32,
+    /// By convention, the below additional data is derived from the agg_sig_me_additional_data
+    agg_sig_parent_additional_data: Bytes32,
+    agg_sig_puzzle_additional_data: Bytes32,
+    agg_sig_amount_additional_data: Bytes32,
+    agg_sig_puzzle_amount_additional_data: Bytes32,
+    agg_sig_parent_amount_additional_data: Bytes32,
+    agg_sig_parent_puzzle_additional_data: Bytes32,
 
     /// The block at height must pay out to this pool puzzle hash.
     genesis_pre_farm_pool_puzzle_hash: Bytes32,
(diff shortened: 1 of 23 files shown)


crates/chia-consensus/src/consensus_constants.rs
```diff
@@ -163,6 +170,24 @@ pub const TEST_CONSTANTS: ConsensusConstants = ConsensusConstants {
...
+    )),
+    agg_sig_puzzle_amount_additional_data: Bytes32::new(hex!(
+        "0f7d90dff0613e6901e24dae59f1e690f18b8f5fbdcf1bb192ac9deaf7de22ad"
+    )),
+    agg_sig_parent_amount_additional_data: Bytes32::new(hex!(
+        "585796bd90bb553c0430b87027ffee08d88aba0162c6e1abbbcc6b583f2ae7f9"
+    )),
+    agg_sig_parent_puzzle_additional_data: Bytes32::new(hex!(
+        "2ebfdae17b29d83bae476a25ea06f0c4bd57298faddbbc3ec5ad29b9b86ce5df"
+    )),
```
arvidn, line 190: these were computed the same way they are in `chia-blockchain`, but these are also just test constants, so it doesn't really matter.

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -253,6 +254,43 @@ pub enum Condition {
...
+fn check_agg_sig_unsafe_message(
+    a: &Allocator,
+    msg: NodePtr,
+    constants: &ConsensusConstants,
+) -> Result<(), ValidationErr> {
+    if a.atom_len(msg) < 32 {
+        return Ok(());
+    }
+    let buf = a.atom(msg);
+    if buf
```
Rigidity, line 266: I'd prefer this iterate over the constants rather than be repeated like this, if possible.

crates/chia-tools/src/bin/analyze-chain.rs
```diff
@@ -161,9 +162,14 @@ fn main() {
         let start_conditions = SystemTime::now();
         // we pass in what's left of max_cost here, to fail early in case the
         // cost of a condition brings us over the cost limit
-        let Ok(conds) =
-            parse_spends::<MempoolVisitor>(&a, generator_output, ti.cost - clvm_cost, MEMPOOL_MODE)
-        else {
+        // TODO: Use real constants
```
Rigidity, line 165: Perhaps we should port the real constants over as well. Out of scope for this PR though, so feel free to resolve this comment

crates/chia-consensus/src/gen/run_puzzle.rs
```diff
@@ -12,6 +13,7 @@ use clvmr::run_program::run_program;
 use clvmr::serde::{node_from_bytes, node_from_bytes_backrefs};
 use std::sync::Arc;
 
+#[allow(clippy::too_many_arguments)]
```
Rigidity, line 16: I think we should find a way to simplify this function somehow, but maybe it's not worth doing in this PR

crates/chia-consensus/src/gen/conditions.rs
arvidn: I suppose it would be unrolled anyway

crates/chia-tools/src/bin/analyze-chain.rs
arvidn: yeah, at some point we should. I want to defer additional complexity from that though.

crates/chia-consensus/src/gen/run_puzzle.rs
arvidn: I agree. I suspect as we move the rust-python boundary forward, more opportunities will open up for that

Rigidity (dismissed):

Rigidity (approved):

PR #555, opened 2024-05-31
Add missing workspace dependencies

Rigidity:


arvidn (approved):
I suspect `cargo fuzz add` may not create Crate files like this, but we could fix them up manually in that case.

PR #556, opened 2024-05-31
Bump BLST to 0.3.12 and remove patch

Rigidity:
We will need to ideally update `clvmr` to use the new `chia-bls` version, then bump `clvmr` in `chia_rs`.

But I think in the lockfile, it's already replaced it since it's a semver compatible version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,8 +224,9 @@ dependencies = [
 
 [[package]]
 name = "blst"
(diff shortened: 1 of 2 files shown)


arvidn (dismissed):

Rigidity:
Rebased against main

emlowe (approved):

PR #557, opened 2024-05-31
remove the NO_RELATIVE_CONDITIONS_ON_EPHEMERAL flag

arvidn:
And unconditionally enable. This soft-fork has activated.

This is OK once https://github.com/Chia-Network/chia-blockchain/pull/18075 lands.


crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -1330,7 +1329,7 @@ pub fn validate_conditions(
     ret: &SpendBundleConditions,
     state: ParseState,
     spends: NodePtr,
-    flags: u32,
+    _flags: u32,
```
Rigidity, line 1332: Does this need to be here still?

Rigidity (approved):

crates/chia-consensus/src/gen/conditions.rs
arvidn: it doesn't. I started removing it but left it in as I think it's quite nice to have the infrastructure to affect conditions parsing by new flags

PR #558, opened 2024-05-31
CHIA-658: Make linting much more strict

Rigidity:


crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -1951,10 +1939,7 @@ fn test_message_strict_args_count(
         flags | ENABLE_SOFTFORK_CONDITION | ENABLE_MESSAGE_CONDITIONS,
     );
     if flags == 0 {
-        if let Err(e) = ret {
-            println!("{:?}", e);
-        }
-        assert!(ret.is_ok());
```
arvidn, line 1957: did clippy figure out that this transform is OK? It's pretty impressive in that case

crates/chia-protocol/fuzz/fuzz_targets/streamable.rs
```diff
@@ -7,16 +7,13 @@ use sha2::{Digest, Sha256};
...
-        }
-        Ok(o) => o,
-    };
+    let obj2 = T::from_bytes(&bytes).unwrap_or_else(|_| {
+        panic!(
+            "failed to parse input: {}, from object: {:?}",
+            hex::encode(&bytes),
+            &obj
+        )
+    });
```
arvidn, line 16: this looks like it would be slightly simpler as:
```suggestion
    let Ok(obj2) = T::from_bytes(&bytes) else {
        panic!(
            "failed to parse input: {}, from object: {:?}",
            hex::encode(&bytes),
            &obj
        )
    };
```

arvidn (approved):

crates/chia-consensus/src/gen/conditions.rs
Rigidity: Did it by hand here

arvidn (approved):

PR #559, opened 2024-06-01
CHIA-659: Rework and expand puzzle types

Rigidity:


crates/chia-puzzles/src/proof.rs
```diff
@@ -22,6 +22,15 @@ pub struct LineageProof {
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 #[clvm(list)]
 pub struct EveProof {
+    pub parent_parent_coin_id: Bytes32,
```
arvidn, line 25: this is usually called `parent_coin_info`, not a great name, but it introduces an inconsistency to use a different name

crates/chia-puzzles/src/puzzles/nft.rs
```diff
@@ -132,33 +132,33 @@ pub struct NftOwnershipLayerSolution<I> {
 pub struct NftRoyaltyTransferPuzzleArgs {
     pub singleton_struct: SingletonStruct,
     pub royalty_puzzle_hash: Bytes32,
-    pub trade_price_percentage: u16,
+    pub royalty_percentage: u16,
```
arvidn, line 135: somewhat unrelated, but my recollection is that this is expressed in thousands, right? if so it should really be called `royalty_permille`

crates/chia-puzzles/src/proof.rs
arvidn: oh, I misread. this is the parent's parent

arvidn (approved):

arvidn (approved):

PR #560, opened 2024-06-01
CHIA-660: Bump dependencies

Rigidity:


crates/chia-ssl/src/ca.rs
```diff
@@ -6,11 +6,14 @@ pub const CHIA_CA_CRT: &str = include_str!("../chia_ca.crt");
 
 lazy_static! {
     pub static ref CHIA_CA: Certificate = load_ca_cert();
+    pub static ref CHIA_CA_KEY_PAIR: KeyPair =
+        KeyPair::from_pem(CHIA_CA_KEY).expect("could not load CA keypair");
```
arvidn, line 10: this looks unrelated. can you describe what API changed in the crate prompting this change?

arvidn (commented):
just one question.
Also, does dependabot suggest upgrades whenever they're available, or is there any kind of security audit as well? It makes me a bit nervous to "blindly" upgrade to new versions of our dependencies (without a strong reason to)

crates/chia-ssl/src/ca.rs
Rigidity: https://github.com/rustls/rcgen/blob/main/rcgen/docs/0.12-to-0.13.md
This is the migration guide for `rcgen` - they changed the API around a bit so I had to rewrite how the SSL cert generation is done

arvidn (approved):

PR #563, opened 2024-06-04
CHIA-896: CHIP-0026 mempool update messages

Rigidity:
Adds the new CHIP-0026 mempoool update messages, which can be used to receive events when mempool items are added or removed that match your puzzle hash and coin subscriptions.


crates/chia-protocol/src/wallet_protocol.rs
```diff
@@ -303,3 +303,50 @@ impl chia_traits::ChiaToPython for RejectStateReason {
         Ok(pyo3::IntoPy::into_py(*self, py).bind(py).clone())
     }
 }
+
+#[repr(u8)]
+#[cfg_attr(feature = "py-bindings", derive(PyJsonDict, PyStreamable))]
+#[derive(Streamable, Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+pub enum MempoolRemoveReason {
+    Conflict = 1,
```
arvidn, line 312: do we really start our enum at 1 and not 0?

wheel/stubtest.allowlist
```diff
@@ -29,6 +29,9 @@ chia_rs\.PrivateKey\.__init__
 # TODO: expects *args
 chia_rs\.RequestPeers\.__init__
 
+# TODO: expects *args
```
arvidn, line 32: I don't think this comment is right. This is an empty struct, that's why it has to be exempt (afaict)

arvidn (approved):

wheel/stubtest.allowlist
Rigidity: I just copied the existing stubtest comments on similar empty structs - I think "expects *args" is the warning that's created by being an empty struct for some reason. @altendky probably knows more about this

altendky: The full error is `chia_rs.RequestCostInfo.__init__ is inconsistent, stub does not have *args argument "args"`.  The Rust form of `RequestCostInfo` matches that of `RequestPeers` above with a similar comment.  It's a todo because it would be nice to figure out what is going on at some point.  I'm ok with continuing this ignore pattern with this addition.

arvidn: this being a `TODO`-comment makes it look like there's some improvement that can be made here, that we're punting on. It's not clear to me what that improvement would be though (at least not from this comment). I suppose we could improve the stub generation to support empty structs.

PR #574, opened 2024-06-09
Python type stub generation with macros

Rigidity:
This makes it a bit less hacky to generate Python type stubs. It's done at compile time with macros, instead of scraping and parsing Rust files by hand. It also allows you to add additional stubs or stubs for non-streamable types within the Rust code. And enforces that stubs exist for all bindings.

However, since it uses macros, it likely slows compile times down a bit, increases the complexity of our setup even more, and is still error prone way of doing this. I'm hoping `pyo3` and `maturin` will eventually have a solution for doing this 100% automatically (though our sized int/byte types would still be an issue there).

This is just an initial draft.


Rigidity:
I don't think this is the solution we should go with, but was worth exploring.

PR #579, opened 2024-06-16
[CHIA-780] member functions for deriving hardened and unhardened keys

arvidn:
for the python bindings.

Suggested by @Quexington


wheel/python/chia_rs/chia_rs.pyi
```diff
@@ -211,6 +212,9 @@ class PrivateKey:
     def sign_g2(self, msg: bytes, dst: bytes) -> G2Element: ...
     def get_g1(self) -> G1Element: ...
     def __str__(self) -> str: ...
+    def public_key(self) -> G1Element: ...
```
Rigidity, line 215: Can we get rid of `get_g1` then?

Rigidity (approved):

wheel/python/chia_rs/chia_rs.pyi
arvidn: `chia-blockchain` still uses that name. if we want to transition all of them, I'm find with removing `get_g1()`

Quexington (approved):

Issue #349, opened 2023-12-16
Freebsd 13.2 ./install.sh failure


SweetAndLow:
Cloned the latest chia as of yesterday into my jail on freebsd 13.2, created my virtual environment and executed the bash ./install.sh script. Got this error. I did a little poking around to see if i could just install via pip or build manually but didn't have any luck,

`Collecting chia-rs==0.2.13 (from chia-blockchain==2.1.2)
  Using cached chia_rs-0.2.13.tar.gz (299 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [12 lines of output]
      💥 maturin failed
        Caused by: Cargo metadata failed. Does your crate compile with `cargo build`?
        Caused by: `cargo metadata` exited with an error: error: failed to load manifest for workspace member `/tmp/pip-install-r8vvmcgg/chia-rs_93784bac3d044b36b6413a2c3ed8292d/chia-bls/fuzz`

      Caused by:
        failed to read `/tmp/pip-install-r8vvmcgg/chia-rs_93784bac3d044b36b6413a2c3ed8292d/chia-bls/fuzz/Cargo.toml`

      Caused by:
        No such file or directory (os error 2)
      Error running maturin: Command '['maturin', 'pep517', 'write-dist-info', '--metadata-directory', '/tmp/pip-modern-metadata-tz7m0cpv', '--interpreter', '/root/chia-2.1.0/chia-blockchain/venv/bin/python']' returned non-zero exit status 1.
      Checking for Rust toolchain....
      Running `maturin pep517 write-dist-info --metadata-directory /tmp/pip-modern-metadata-tz7m0cpv --interpreter /root/chia-2.1.0/chia-blockchain/venv/bin/python`
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.`

hoffmang9:
I've tried the basics on 13.2 including replication.

I can also build the crate in the root dir of chia_rs and `pip install .` in the wheel/ dir and that all works.

Not sure why it is chocking on the bls fuzzer.

Others can look at it during the week.

[...]

PR #585, opened 2024-06-25
Add python source to `Cargo.toml`

Quexington:
I made this because it allows me to specify a specific commit to install `chia_rs` from with pip.  Without this entry, the stub file doesn't come along and mypy doesn't function properly.  The `chia-blockchain` CI has some difficulties with this but I think that's because it is manually building it when installing from the commit rather than installing from a published version.  I'm not entirely sure how to check that though so some input on how reasonable this seems would be appreciated.

### DIFF
--- a/wheel/Cargo.toml
+++ b/wheel/Cargo.toml
@@ -17,6 +17,9 @@ name = "chia_rs"
 crate-type = ["cdylib"]
 path = "src/lib.rs"
 
+[package.metadata.maturin]
+python-source = "python"
+
 [dependencies]
 clvmr = { workspace = true }
 hex = { workspace = true }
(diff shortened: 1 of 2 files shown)


Rigidity:
Source installations via pip working is blocked on this issue:
https://github.com/Chia-Network/chia_rs/issues/349

Quexington:
> Source installations via pip working is blocked on this issue: #349

I don't follow this thread so well this is not my domain of expertise but I don't think this _requires_ anything new to happen during your run of the mill build process, it just _enables_ pip installing from source to work (on at least some platforms).  I don't know though.

Rigidity (approved):
I don't see anything wrong with this, but it's marked as draft - so that's why I haven't approved sooner.

Quexington:
As long as no one sees any unexpected side effects of this then go ahead and merge it at your convenience

PR #590, opened 2024-06-26
Expose `from_seed` in `SecretKey` to python as well

Quexington:
This is another method it would be helpful to have exposed on the python side


Rigidity:
Two things - you'll need to update generate_type_stubs and run the file to update the chia_rs.pyi file rather than doing it by hand

And `Self::from_seed` is probably what you're trying to do here - this is an associated function on the struct rather than an instance method

Rigidity:
Otherwise, lgtm

Rigidity (approved):

arvidn (approved):

PR #592, opened 2024-07-02
CHIA-886: Redesign ToClvm and FromClvm to be generic over Encoder and Decoder

Rigidity:
Makes a few changes:
* `ToClvm` is generic over `Encoder` rather than `Node`.
* `FromClvm` is generic over `Decoder` rather than `Node`.
* `ToNodePtr` and `FromNodePtr` have been removed since they are no longer needed.
* `Program` can now implement `ToClvm<Allocator>` and `FromClvm<Allocator>` directly.
* `Encoder` now has a method for `encode_bigint` and `encode_atom` takes a `clvmr::Atom` instead of `&[u8]`.


crates/clvm-traits/src/clvm_encoder.rs
```diff
@@ -1,17 +1,33 @@
...
         first: Self::Node,
         rest: Self::Node,
     ) -> Result<Self::Node, ToClvmError>;
 
+    fn encode_bigint(&mut self, number: BigInt) -> Result<Self::Node, ToClvmError> {
+        let bytes = number.to_signed_bytes_be();
+        let mut slice = bytes.as_slice();
+
+        // Remove leading zeros.
+        while !slice.is_empty() && slice[0] == 0 {
```
arvidn, line 21: this doesn't look right. If you want to encode a *positive* integer whose MSB is 1, you need one leading zero. .e.g. 255.

It would be good to have unit tests covering this

crates/clvm-traits/src/match_byte.rs
```diff
@@ -5,19 +5,14 @@ use crate::{ClvmDecoder, ClvmEncoder, FromClvm, FromClvmError, ToClvm, ToClvmErr
 #[derive(Debug, Copy, Clone)]
 pub struct MatchByte<const BYTE: u8>;
 
-impl<N, const BYTE: u8> ToClvm<N> for MatchByte<BYTE> {
-    fn to_clvm(&self, encoder: &mut impl ClvmEncoder<Node = N>) -> Result<N, ToClvmError> {
-        if BYTE == 0 {
-            return encoder.encode_atom(&[]);
-        }
-        let number = BigInt::from(BYTE);
-        let bytes = number.to_signed_bytes_be();
```
arvidn, line 14: the old code here looks a bit suspicious too, it doesn't handle leading zeroes based on whether the number of positive or negative at all. But you're changing this behavior now, afaict.

crates/clvm-traits/src/match_byte.rs
```diff
@@ -5,19 +5,14 @@ use crate::{ClvmDecoder, ClvmEncoder, FromClvm, FromClvmError, ToClvm, ToClvmErr
...
-        encoder.encode_atom(&bytes)
+impl<N, E: ClvmEncoder<Node = N>, const BYTE: u8> ToClvm<E> for MatchByte<BYTE> {
+    fn to_clvm(&self, encoder: &mut E) -> Result<N, ToClvmError> {
+        encoder.encode_bigint(BigInt::from(BYTE))
```
arvidn, line 10: it seems like quite a detour to go via a `BitInt` here. In most cases, you'll most likely end up with the same byte as you started with, right?
Maybe it would make sense to do something like:
```
if BYTE == 0 { encoder.encode_atom(&[]) }
else if BYTE < 0x80 { encoder.encode_atom(&[BYTE]); } 
else { encoder.encode_atom(&[0, BYTE]); } 
```

arvidn (commented):
I really like this direction. There were some details that looked fishy

crates/clvm-traits/src/match_byte.rs
Rigidity: Changed it to this

crates/clvm-traits/src/match_byte.rs
Rigidity: Fixed for decoding and added a test for each byte value with a macro

crates/clvm-traits/src/clvm_encoder.rs
Rigidity: There's a subsequent check `slice.len() > 1 && (slice[1] & 0x80 == 0x80)` which breaks the loop.

This code was taken directly from `clvm_rs` here:
https://github.com/Chia-Network/clvm_rs/blob/main/src/number.rs#L10

crates/clvm-traits/src/clvm_encoder.rs
```diff
@@ -43,6 +63,10 @@ impl ClvmEncoder for Allocator {
     ) -> Result<Self::Node, ToClvmError> {
         self.new_pair(first, rest).or(Err(ToClvmError::OutOfMemory))
     }
+
+    fn encode_bigint(&mut self, number: BigInt) -> Result<Self::Node, ToClvmError> {
+        self.new_number(number).or(Err(ToClvmError::OutOfMemory))
+    }
```
arvidn: this is nice. I envision that we may want an `encode_g1()` and `encode_g2()` at some point in the future as well (when the allocator supports those types natively)

crates/clvm-traits/src/match_byte.rs
arvidn: this is still very expensive compared to just `encode_atom()` with a single byte. Anything involving creating and destroying a `BigInt` is expensive. It's one of the main bottlenecks in `clvmr`

arvidn (approved):

PR #598, opened 2024-07-08
CHIA-899: Add Python binding to scalar_multiply

Rigidity:


arvidn (commented):
we don't have a very comprehensive test for BLS on the python side, but for new python bindings, I think it would be good to also add some simple tests in python, exercising the bindings themselves

crates/chia-bls/src/signature.rs
```diff
@@ -1203,6 +1216,27 @@ mod tests {
...
+        g2.scalar_multiply(&data);
+        assert_eq!(
+            hex::encode(g2.to_bytes()),
+            "
+            ae4d384a25c51283b8be8c6546d23e1555995c87fbcc0fe12169b63a052c4a1f
+            c9d3a020e8e010d4be619e3c0980a1f213b951fe75375012c5df6a690548a637
+            ef25f8da1e4f9c8f4d2062531ce688c040258a76543831abde774872e00af74b
+            "
+            .replace([' ', '\n'], "")
+        );
```
arvidn, line 1237: you should have a comment with a reference where this test vector comes from, or you could turn it into a property-check, to ensure the result satisfies some expected properties of scalar multiplication

arvidn (commented):
proper test in rust (and maybe even a property test in a fuzzer) would be warranted

Rigidity:
I'm not sure this is necessary at the moment

PR #600, opened 2024-07-08
Subclassable spend bundle

Rigidity:


wheel/generate_type_stubs.py
```diff
@@ -336,6 +339,7 @@ def get_puzzle_and_solution_for_coin(program: ReadableBuffer, args: ReadableBuff
 
 class BLSCache:
     def __init__(self, cache_size: Optional[int] = 50000) -> None: ...
+    def __new__(cls, cache_size: Optional[int] = 50000) -> BLSCache: ...
```
arvidn, line 342: do we need both `__init__` and `__new__`?

Rigidity: I don't think we necessarily need this PR anymore, I'll put it into draft when I'm back home

Rigidity:
Replaced by another PR

PR #605, opened 2024-07-10
Resolve dependabot alerts

Rigidity:


jack60612 (approved):
concerned by coverage but approving

PR #609, opened 2024-07-10
CHIA-622 switch to use clvmr hashing only + clvm_rs 0.8.0

jack60612:
one edge case for loading entropy in key gen however, so sha2 is still used

- [ ] clvm_rs release 0.8.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ chia-traits = { workspace = true, optional = true }
 chia-puzzles = { workspace = true, optional = true }
 clvm-traits = { workspace = true, optional = true }
 clvm-utils = { workspace = true, optional = true }
+clvmr = { workspace = true}
 
 [features]
 default = [
@@ -83,6 +84,8 @@ puzzles = ["dep:chia-puzzles"]
 clvm-traits = ["dep:clvm-traits"]
 clvm-utils = ["dep:clvm-utils"]
 
+openssl = ["clvmr/openssl"] # enable openssl for clvmr on unix
+
 [profile.release]
 lto = "thin"
 
@@ -101,13 +104,12 @@ clvm-utils = { path = "./crates/clvm-utils", version = "0.10.0" }
 clvm-derive = { path = "./crates/clvm-derive", version = "0.10.0" }
 chia-fuzz = { path = "./crates/chia-consensus/fuzz", version = "0.10.0" }
 blst = { version = "0.3.12", features = ["portable"] }
-clvmr = "0.7.0"
+clvmr = "0.8.0"
 syn = "2.0.70"
 quote = "1.0.32"
 proc-macro2 = "1.0.84"
 proc-macro-crate = "1.3.1"
 anyhow = "1.0.86"
-sha2 = "0.10.8"
 hkdf = "0.12.0"
 hex = "0.4.3"
 thiserror = "1.0.61"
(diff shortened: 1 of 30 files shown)


jack60612:
requires clvm_rs release

crates/chia-bls/Cargo.toml
```diff
@@ -19,7 +19,8 @@ arbitrary = ["dep:arbitrary"]
 chia-traits = { workspace = true }
 chia_py_streamable_macro = { workspace = true, optional = true }
 anyhow = { workspace = true }
-sha2 = { workspace = true }
+sha2 = {version = "0.10.8"}
```
Rigidity, line 22: Use `{ workspace = true }` here. We use [workspace dependencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table)

Cargo.toml
```diff
@@ -101,13 +104,12 @@ clvm-utils = { path = "./crates/clvm-utils", version = "0.10.0" }
...
 chia-fuzz = { path = "./crates/chia-consensus/fuzz", version = "0.10.0" }
 blst = { version = "0.3.12", features = ["portable"] }
-clvmr = "0.7.0"
+clvmr = "0.8.0"
 syn = "2.0.70"
 quote = "1.0.32"
 proc-macro2 = "1.0.84"
 proc-macro-crate = "1.3.1"
 anyhow = "1.0.86"
-sha2 = "0.10.8"
```
Rigidity, line 110: If we're still using this in a couple places for now, this will need to be here still.

Cargo.toml
```diff
@@ -83,6 +84,8 @@ puzzles = ["dep:chia-puzzles"]
 clvm-traits = ["dep:clvm-traits"]
 clvm-utils = ["dep:clvm-utils"]
 
+openssl = ["clvmr/openssl"] # enable openssl for clvmr on unix
```
Rigidity, line 87: This will be decided by dependents and/or the CI here, not this manifest. So this comment is unnecessary imo.

Cargo.toml
```diff
@@ -59,6 +59,7 @@ chia-traits = { workspace = true, optional = true }
 chia-puzzles = { workspace = true, optional = true }
 clvm-traits = { workspace = true, optional = true }
 clvm-utils = { workspace = true, optional = true }
+clvmr = { workspace = true}
```
Rigidity, line 62: nit: space

crates/chia-puzzles/src/derive_synthetic.rs
```diff
@@ -1,9 +1,8 @@
+use crate::standard::DEFAULT_HIDDEN_PUZZLE_HASH;
```
Rigidity, line 1: Did you mean to reorder this?

I usually put crate dependencies below the rest (above is fine too, but at the very least separated via whitespace is nice).

Cargo.toml
jack60612: ok ill move back from workspace

crates/chia-puzzles/src/derive_synthetic.rs
jack60612: my ide does things

crates/chia-bls/Cargo.toml
jack60612: changed

Cargo.toml
jack60612: removed

Cargo.toml
jack60612: fixed

Rigidity:
clvmr has been updated

PR #610, opened 2024-07-11
CHIA-622 Upgrade version to 0.11.0 

jack60612:
This is still pending, but we should only merge after at least #609, and ideally later


Rigidity:
This has already been done now

PR #611, opened 2024-07-14
CHIA-878: New peer implementation

Rigidity:
A new `Peer` implementation that:
* Keeps track of `PeerId` and `IpAddr`
* Has improved handling of messages
* Resolves some previous TODO items
* Higher code quality

You can run the following examples to test this out:
* `RUST_LOG=info cargo run -p chia-client --example peer_connection`
* `RUST_LOG=info cargo run -p chia-client --example peer_discovery`


Rigidity:
This is still a work in progress. I'm trying to get rid of potential race conditions and deadlocks.

crates/chia-client/src/request_map.rs
```diff
@@ -0,0 +1,65 @@
...
+pub struct RequestMap {
+    items: Mutex<HashMap<u16, Request>>,
+    semaphore: Arc<Semaphore>,
+}
+
+impl RequestMap {
+    pub fn new() -> Self {
+        Self {
+            items: Mutex::new(HashMap::new()),
+            semaphore: Arc::new(Semaphore::new(u16::MAX as usize)),
```
arvidn, line 28: can you add a comment explaining why you initialize the semaphore to `u16::MAX`?
it seems like some arbitrarily picked upper limit

crates/chia-client/src/request_map.rs
```diff
@@ -0,0 +1,65 @@
...
+impl Request {
+    pub fn send(self, message: Message) {
+        self.sender.send(message).ok();
+    }
+}
+
+#[derive(Debug)]
+pub struct RequestMap {
+    items: Mutex<HashMap<u16, Request>>,
+    semaphore: Arc<Semaphore>,
```
arvidn, line 21: `Mutex` is a tokio mutex, to be used with coroutines. `Arc` is the standard library primitive to be used with threads. It seems suspicious to mix these.

arvidn: It would be good to add a comment explaining what the semaphore is used for. It's kind of an odd primitive that was used in the early days, before mutexes and condition variables were invented.

crates/chia-client/src/request_map.rs
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    pub async fn insert(&self, sender: oneshot::Sender<Message>) -> u16 {
+        let permit = self
+            .semaphore
+            .clone()
+            .acquire_owned()
+            .await
+            .expect("semaphore closed");
```
arvidn, line 38: I imagine the semaphore is use to ensure no more than 65k coroutines enter this function simultaneously. Although, the mutex lock below would serialize them all anyway, it ensures there's always a free slot in the hash map.

This is subtle enough to warrant a comment I think

crates/chia-client/src/request_map.rs
```diff
@@ -0,0 +1,65 @@
...
+
+impl Request {
+    pub fn send(self, message: Message) {
+        self.sender.send(message).ok();
+    }
+}
+
+#[derive(Debug)]
+pub struct RequestMap {
+    items: Mutex<HashMap<u16, Request>>,
```
arvidn, line 20: This isn't really used as a hash map. I imagine the reason you picked this data structure was to save space in the common case of not having many messages in flight. It looks like it could also be:
```
Mutex<[Option<Request>; u16::MAX]>
```

crates/chia-client/src/request_map.rs
```diff
@@ -0,0 +1,65 @@
...
+        let mut items = self.items.lock().await;
+
+        let mut index = None;
+
+        for i in 0..=u16::MAX {
+            if !items.contains_key(&i) {
+                index = Some(i);
+                break;
+            }
+        }
```
arvidn, line 49: one simple optimization over a linear search is to store a `free_slot: Option<u16>,` member, which is set every time you release a slot, and before this linear search, you could just see if that slot is available.

Another, more heavy weight, optimization would be to have a stack of all free slots that you push and pop from, but given that the common case is (likely) to have few messages in flight, that is probably not worth it.

crates/chia-client/src/peer.rs
```diff
@@ -1,363 +1,341 @@
...
-            };
+        match message {
+            Text(text) => {
+                log::warn!("Received unexpected text message: {text}");
+            }
+            Close(close) => {
+                log::warn!("Received close: {close:?}");
+                break;
+            }
+            Ping(_ping) => {}
```
arvidn, line 312: are we not expected to (explicitly) respond with a pong in this case?

crates/chia-client/src/peer.rs
```diff
@@ -1,363 +1,341 @@
...
+        self.0.sink.lock().await.send(message).await?;
+        Ok(receiver.await?)
     }
+}
 
-    pub fn receiver(&self) -> &broadcast::Receiver<PeerEvent> {
-        &self.event_receiver
+impl Drop for PeerInner {
+    fn drop(&mut self) {
+        self.inbound_handle.abort();
```
arvidn, line 290: this looks a bit suspicious. If we haven't joined this coroutine by the time this is dropped, isn't it too late?

crates/chia-client/src/peer.rs
```diff
@@ -1,363 +1,341 @@
...
+                    })?;
+                    continue;
+                };
+
+                let Some(request) = requests.remove(id).await else {
+                    log::warn!(
+                        "Received {:?} message with untracked id {id}",
+                        message.msg_type
+                    );
+                    return Err(Error::UnexpectedMessage(message.msg_type));
```
arvidn, line 330: This is disconnecting this one peer, right?

crates/chia-client/src/request_map.rs
arvidn: why does this need to be held by an `Arc`? The whole `RequestMap` is already held by an `Arc`

arvidn (commented):
as far as I can tell, there are no threads involved. The fact that you use `Arc` instead of `Rc` suggests that there are threads. If I'm not mistaken, I think you should use `Rc` instead

crates/chia-client/src/request_map.rs
Rigidity: In the Chia protocol, message ids are 16 bit unsigned integers

crates/chia-client/src/peer.rs
Rigidity: If you drop an `AbortHandle` to a task, the default behavior is to leave it running in the background. The more sensible behavior here is to just kill the task.

That said, I think the underlying `Stream` might close if the `Sink` is destroyed, so this might not do much.

crates/chia-client/src/peer.rs
Rigidity: Tungstenite (the WebSocket library) handles pings automatically, so I'm not sure why this is even a message type. I don't think it's ever received.

crates/chia-client/src/request_map.rs
Rigidity: That seems like it could allocate a lot of memory. Most of the time you only have one or two requests pending, so this whole system is for an edge case which is unlikely to happen anyways

crates/chia-client/src/request_map.rs
Rigidity: Because you can't get owned semaphore guards from a non-`Arc` `Semaphore`

crates/chia-client/src/peer.rs
Rigidity: Yeah currently this crate only handles single peer connections

Rigidity:
I'm going to close this for now

PR #612, opened 2024-07-15
Install openssl on benchmark runner

Rigidity:


arvidn:
this won't make a difference yet, right? at least in `clvm_rs` you still need to say `--features openssl` to build against it

Rigidity:
I'll close for now

PR #616, opened 2024-07-16
CHIA-922: CHIA-938: Add signing and verification methods

Rigidity:
Adds `G1Element::verify` and `SecretKey::sign` in Python as part of the wallet signer effort.


matt-o-how:
Non-blocking comment - can you add some descriptions of what and why for this PR?

AmineKhaldi (approved):

crates/chia-bls/src/secret_key.rs
```diff
@@ -242,8 +242,15 @@ impl SecretKey {
     #[classattr]
     pub const PRIVATE_KEY_SIZE: usize = 32;
 
-    pub fn sign_g2(&self, msg: &[u8]) -> crate::Signature {
-        crate::sign(self, msg)
+    pub fn sign(&self, msg: &[u8], final_pk: Option<PublicKey>) -> crate::Signature {
```
arvidn, line 245: I understand the python interface has this optional public key, but do we ever actually use it? If not, it would be cleaner to just not have it

PR #624, opened 2024-07-22
`chia-blockchain==2.4.2` in `build-test.yml`

altendky:


Rigidity:
This is intentionally pinned to test fidelity against old versions of things which were ported to Rust. So this would be a larger change than just the version

altendky:
Gotcha, alrighty, thanks.  So I guess I should say why I was doing this.  Awhile back over in https://github.com/Chia-Network/chia_rs/pull/427 I was adding testing against 3.12.  That fails because blockchain didn't support 3.12 until 2.3.0 if I checked right.  So the actual interest is to test functionality on 3.12.  I'll close this and maybe you could comment over on #427 or just message me?

PR #637, opened 2024-07-31
Extend the BLSCache to add in a rust Pybinding for validate_clvm_and_signature for use in the mempool

matt-o-how:


crates/chia-bls/src/bls_cache.rs
```diff
@@ -122,9 +129,9 @@ impl BlsCache {
         let msgs = msgs
             .iter()?
             .map(|item| item?.extract())
-            .collect::<PyResult<Vec<PyBackedBytes>>>()?;
```
arvidn, line 125: what's the rationale for making this change?

crates/chia-bls/src/bls_cache.rs
```diff
@@ -138,26 +145,19 @@ impl BlsCache {
         use pyo3::types::PyBytes;
         let ret = PyList::empty_bound(py);
         for (key, value) in &self.cache {
-            ret.append((
-                PyBytes::new_bound(py, key),
-                PyBytes::new_bound(py, &value.to_bytes()),
-            ))?;
+            ret.append((PyBytes::new_bound(py, key), value.clone().into_py(py)))?;
```
arvidn, line 148: we talked about going back to using plan `bytes` here, since `GTElement` doesn't support pickle. (maybe it should in the future, but let's make that a separate PR)

crates/chia-bls/src/bls_cache.rs
```diff
@@ -64,7 +71,7 @@ impl BlsCache {
 
             // Otherwise, we need to calculate the pairing and add it to the cache.
             let mut aug_msg = pk.borrow().to_bytes().to_vec();
-            aug_msg.extend_from_slice(msg.as_ref());
```
arvidn, line 67: I would expect `extend_from_slice()` to be more efficient. Is there a good reason to make this change?

wheel/generate_type_stubs.py
```diff
@@ -305,6 +305,27 @@ def confirm_not_included_already_hashed(
     proof: bytes,
 ) -> bool: ...
 
+def validate_clvm_and_signature(
+    new_spend: SpendBundle,
+    max_cost: int,
+    constants: ConsensusConstants,
+    peak_height: int,
+    cache: Option[BLSCache],
```
arvidn, line 313: I don't think you should pass in the cache here. If we want to make that change, I think it should be considered independently of porting this to rust

wheel/generate_type_stubs.py
```diff
@@ -305,6 +305,27 @@ def confirm_not_included_already_hashed(
...
+) -> Tuple[SpendBundleConditions, List[Tuple[SpendBundleConditions, List[Tuple[Bytes32, bytes]]]], float]: ...
+
+def get_name_puzzle_conditions(
+    spend_bundle: SpendBundle,
+    max_cost: int,
+    constants: ConsensusConstants,
+    height: int,
+) -> SpendBundleConditions: ...
```
arvidn, line 321: It doesn't look like this function is defined in this PR. Did you mean `get_conditions_from_spendbundle()`?
It would be good to understand why this wasn't caught by CI as well. Perhaps there's no python test for this function, or we don't run `mypy`

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -0,0 +1,400 @@
...
+
+// currently in mempool_manager.py
+// called in threads from pre_validate_spend_bundle()
+// returns (error, cached_results, new_cache_entries, duration)
+pub fn validate_clvm_and_signature(
+    spend_bundle: &SpendBundle,
+    max_cost: u64,
+    constants: &ConsensusConstants,
+    height: u32,
+    cache: &Arc<Mutex<BlsCache>>,
```
arvidn, line 26: I don't think you should pass in the cache here

tests/test_blscache.py
```diff
@@ -126,27 +215,21 @@ def test_cached_bls_flattening():
...
     assert len(cached_bls.items()) == n_keys
-    gts = [
-        bytes(pk.pair(AugSchemeMPL.g2_from_message(bytes(pk) + b"foobar")))
-        for pk in pks
-    ]
+    gts = [pk.pair(AugSchemeMPL.g2_from_message(bytes(pk) + b"foobar")) for pk in pks]
     for key, value in cached_bls.items():
         assert isinstance(key, bytes)
-        assert isinstance(value, bytes)
+        assert isinstance(value, GTElement)
```
arvidn, line 221: we agreed to these as `bytes`, right?

arvidn (changes requested):

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,194 @@
+use crate::consensus_constants::ConsensusConstants;
+use crate::gen::opcodes::{
+    ConditionOpcode, AGG_SIG_AMOUNT, AGG_SIG_ME, AGG_SIG_PARENT, AGG_SIG_PARENT_AMOUNT,
+    AGG_SIG_PARENT_PUZZLE, AGG_SIG_PUZZLE, AGG_SIG_PUZZLE_AMOUNT,
+};
+use crate::gen::owned_conditions::OwnedSpend;
+use chia_protocol::Bytes;
+use chia_protocol::Coin;
+
+pub fn make_aggsig_final_message(
```
arvidn, line 10: this is the only public function in this file. Would it not make more sense to call the file `make_aggsig_final_message.rs`?

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -0,0 +1,400 @@
...
+    use chia_bls::{sign, G2Element, SecretKey, Signature};
+    use chia_protocol::{Bytes, Bytes32};
+    use chia_protocol::{Coin, CoinSpend, Program};
+    use clvm_utils::tree_hash_atom;
+    use hex::FromHex;
+    use hex_literal::hex;
+    use std::sync::Arc;
+
+    #[test]
+    fn test_validate_no_pks() {
```
arvidn, line 119: there seems to be a lot of repetitions across these tests. did you try to factor something out or make it a parameterized test?

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -0,0 +1,400 @@
...
+    let unsafe_items = npcresult
+        .agg_sig_unsafe
+        .iter()
+        .map(|(pk, msg)| (pk, msg.as_slice().to_vec()));
+    let iter = iter.chain(unsafe_items);
+    // Verify aggregated signature
+    let (result, added) = cache
+        .lock()
+        .unwrap()
+        .aggregate_verify(iter, &spend_bundle.aggregated_signature);
```
arvidn, line 67: this function doesn't really need the cache. the python implementation just uses the cache as a way to get at the resulting pairings. I think we could disentangle this function from the cache entirely by just computing the pairings, validating them and returning them

wheel/src/api.rs
```diff
@@ -385,6 +437,11 @@ pub fn chia_rs(_py: Python<'_>, m: &Bound<'_, PyModule>) -> PyResult<()> {
     m.add_function(wrap_pyfunction!(confirm_included_already_hashed, m)?)?;
     m.add_function(wrap_pyfunction!(confirm_not_included_already_hashed, m)?)?;
 
+    // multithread validattion
```
arvidn, line 440: I think this comment is misleading. These functions are just plain functions, there's no threading involved here

crates/chia-consensus/src/gen/condition_tools.rs
matt-o-how: there's also u64_to_bytes in here

tests/test_blscache.py
matt-o-how: Done

wheel/generate_type_stubs.py
matt-o-how: Fixed the immediate problem, but I agree the testing for the pyi file is lacking

Rigidity: Manually generating this file is I hope only a temporary solution

Pyo3 is working on automatic type stub generation (though I don't know how we'll do our custom integer types)

crates/chia-consensus/src/gen/condition_tools.rs
arvidn: but it shouldn't be public, should it?

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,194 @@
...
+            result.extend(coin.coin_id().as_slice());
+            result.extend(constants.agg_sig_me_additional_data.as_slice());
+        }
+        _ => return result,
+    };
+
+    result
+}
+
+pub fn u64_to_bytes(val: u64) -> Bytes {
```
arvidn, line 58: this seems to be an internal function that probably shouldn't be public/exported

wheel/generate_type_stubs.py
arvidn: we do run `mypy`, so I would have expected this to be caught if there was a test attempting to use this function (under either this name or the true name)

crates/chia-bls/src/bls_cache.rs
matt-o-how: Fixed

crates/chia-consensus/src/gen/condition_tools.rs
matt-o-how: It's used in `spendbundle_validation.rs`'s tests and is fairly useful to keep around

crates/chia-consensus/src/gen/condition_tools.rs
matt-o-how: It's used in spendbundle_validation.rs's tests and I suspect will be useful in more places as the Rust ports continue

crates/chia-bls/src/bls_cache.rs
matt-o-how: It felt better passing a pubkey in rather than serializing and deserializing, but I wasn't aware that that was just causing it to be pickled, same as with the GTElement change. 
I've reverted it back now.

crates/chia-consensus/src/spendbundle_validation.rs
arvidn: it's probably simpler to mimic the python implementation for now

crates/chia-consensus/src/gen/condition_tools.rs
arvidn: it should probably be `pub(crate)` then

arvidn (commented):
I think you should revert `crates/chia-bls/src/bls_cache.rs`, `crates/chia-bls/benches/cache.rs`, `crates/chia-bls/src/lib.rs`.

You can define `PairingInfo` in `wheel/src/api.rs` if you like a name for it.

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -0,0 +1,416 @@
...
+    if !result {
+        return Err(ErrorCode::BadAggregateSignature);
+    }
+    Ok((npcresult, added, start_time.elapsed()))
```
arvidn, line 74:
```suggestion
    Ok((npcresult, cache.items(), start_time.elapsed()))
```

This change requires you to add the `items()` function to the cache (which currently only exists in the python binding). Alternatively (but maybe a bigger change) would be to replace the call to `aggregate_verify()` with just a loop over computing the pairings, and return those. There's already an `aggregate_verify_gt()` for the signature validation

crates/chia-consensus/src/gen/condition_tools.rs
matt-o-how: Done

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,194 @@
...
+    #[case(AGG_SIG_PUZZLE_AMOUNT, b"puzzle_amount_message", hex!("4444444444444444444444444444444444444444444444444444444444444444").into(), hex!("3333333333333333333333333333333333333333333333333333333333333333").into(), 410)]
+    #[case(AGG_SIG_PARENT_AMOUNT, b"parent_amount_message", hex!("4444444444444444444444444444444444444444444444444444444444444444").into(), hex!("3333333333333333333333333333333333333333333333333333333333333333").into(), 909)]
+    #[case(AGG_SIG_PARENT_PUZZLE, b"parent_puzzle_message", hex!("4444444444444444444444444444444444444444444444444444444444444444").into(), hex!("3333333333333333333333333333333333333333333333333333333333333333").into(), 10_061_997)]
+    #[case(AGG_SIG_ME, b"me_message", hex!("4444444444444444444444444444444444444444444444444444444444444444").into(), hex!("3333333333333333333333333333333333333333333333333333333333333333").into(), 1303)]
+    fn test_make_aggsig_final_message(
+        #[case] opcode: ConditionOpcode,
+        #[case] msg: &[u8],
+        #[case] parent_id: Vec<u8>,
+        #[case] puzzle_hash: Vec<u8>,
```
arvidn, line 116: every test case uses the same values for `parent_id` and `puzzle_hash`. I don't think you need to make these parameters to the test.

I think you could make `msg` a constant in the tests as well, since you don't need to test different values of it (I think).

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,194 @@
...
+            result.extend(coin.coin_id().as_slice());
+            result.extend(constants.agg_sig_me_additional_data.as_slice());
+        }
+        _ => return result,
+    };
+
+    result
+}
+
+pub(crate) fn u64_to_bytes(val: u64) -> Bytes {
```
arvidn, line 58: we already have two copies of this exact function (with this name) in the repository. The other copies are only available to tests however. So I still think it would make sense to make this private (not public) and use one of the existing copies for the test.

https://github.com/Chia-Network/chia_rs/pull/645

crates/chia-consensus/src/gen/condition_tools.rs
```diff
@@ -0,0 +1,194 @@
...
+                        .agg_sig_parent_puzzle_additional_data
+                        .as_slice(),
+                );
+            }
+            AGG_SIG_ME => {
+                expected_result.extend(coin.coin_id().as_slice());
+                expected_result.extend(TEST_CONSTANTS.agg_sig_me_additional_data.as_slice());
+            }
+            _ => {}
+        };
```
arvidn, line 179: rather than computing the expected value programatically, I think it would be easier to read this test (and be confident it's correct) if the expected output was a parameter instead. Since these are binary strings, it could be simplified by either altering the test constants to make the additional data ascii, or to change the messages to be hex and then specify the expected output as hex. Given that the amount will be binary either way, that might be the better option

crates/chia-consensus/src/gen/condition_tools.rs
matt-o-how: Done

PR #639, opened 2024-08-02
update BLSCache::aggregate_verify() function signature

arvidn:
to take an iterator of pairs rather than two separate iterators.

This allows for interleaving work in an iterator, which isn't possible with *two* iterators


matt-o-how (approved):
Looks good to me

Rigidity (approved):

PR #643, opened 2024-08-05
Only pass coin spends to get_conditions_from_spendbundle

Rigidity:


arvidn (approved):
The only reason I can think of to not do this is if we're planning to make `get_conditions_from_spendbundle()` validate the signature in the future

Rigidity:
> The only reason I can think of to not do this is if we're planning to make `get_conditions_from_spendbundle()` validate the signature in the future

Personally it's preferable to keep them separate since you often want to validate transactions, compute the cost, or unit test them without actually having valid signatures

Maybe it could be a flag though

arvidn:
> Personally it's preferable to keep them separate since you often want to validate transactions, compute the cost, or unit test them without actually having valid signatures

Yeah, there are quite a lot of places where we trust the block to be valid (e.g. if we find it in our DB) but we just want to run it to get at the conditions. Then we don't want to also validate the signatures.

(this needs a rebase now btw)

Rigidity:
Closing for now

PR #646, opened 2024-08-06
Remove cache from validation

matt-o-how:
Creating a new branch and PRing into `validate_clvm_and_sigs` as this change is rather big.

### DIFF
--- a/crates/chia-consensus/src/spendbundle_validation.rs
+++ b/crates/chia-consensus/src/spendbundle_validation.rs
@@ -9,9 +9,10 @@ use crate::gen::opcodes::{
 use crate::gen::owned_conditions::OwnedSpendBundleConditions;
 use crate::gen::validation_error::ErrorCode;
 use crate::spendbundle_conditions::get_conditions_from_spendbundle;
-use chia_bls::BlsCache;
 use chia_bls::PairingInfo;
+use chia_bls::{aggregate_verify_gt, hash_to_g2, BlsCache};
 use chia_protocol::SpendBundle;
+use clvmr::sha2::Sha256;
 use clvmr::{ENABLE_BLS_OPS_OUTSIDE_GUARD, ENABLE_FIXED_DIV, LIMIT_HEAP};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
@@ -31,7 +32,8 @@ pub fn validate_clvm_and_signature(
     let sbc = get_conditions_from_spendbundle(&mut a, spend_bundle, max_cost, height, constants)
         .map_err(|e| e.1)?;
     let npcresult = OwnedSpendBundleConditions::from(&a, sbc);
-    let iter = npcresult.spends.iter().flat_map(|spend| {
+    let spends = npcresult.spends.clone();
+    let iter = spends.iter().flat_map(|spend| {
         let condition_items_pairs = [
             (AGG_SIG_PARENT, &spend.agg_sig_parent),
             (AGG_SIG_PUZZLE, &spend.agg_sig_puzzle),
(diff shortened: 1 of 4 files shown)


crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -24,14 +24,14 @@ pub fn validate_clvm_and_signature(
...
     height: u32,
-    cache: &Arc<Mutex<BlsCache>>,
 ) -> Result<(OwnedSpendBundleConditions, Vec<PairingInfo>, Duration), ErrorCode> {
     let start_time = Instant::now();
     let mut a = make_allocator(LIMIT_HEAP);
     let sbc = get_conditions_from_spendbundle(&mut a, spend_bundle, max_cost, height, constants)
         .map_err(|e| e.1)?;
     let npcresult = OwnedSpendBundleConditions::from(&a, sbc);
-    let iter = npcresult.spends.iter().flat_map(|spend| {
+    let spends = npcresult.spends.clone();
```
arvidn, line 33: I don't think you need to clone the spends here

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -46,32 +46,54 @@ pub fn validate_clvm_and_signature(
...
+                        constants,
+                    );
+                    aug_msg.extend_from_slice(msg.as_ref());
+                    let aug_hash = hash_to_g2(&aug_msg);
+                    let pairing = aug_hash.pair(pk);
+
+                    let mut hasher = Sha256::new();
+                    hasher.update(pk.to_bytes());
+                    hasher.update(msg.as_slice());
+                    let hash: [u8; 32] = hasher.finalize();
```
arvidn, line 63: it would probably make sense to factor this out as a separate function, as this code is repeated.

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -46,32 +46,54 @@ pub fn validate_clvm_and_signature(
             .flat_map(move |(condition, items)| {
                 let spend_clone = spend.clone();
                 items.iter().map(move |(pk, msg)| {
```
arvidn, line 48: is there a good reason to *move* the spends into this lambda function? That's probably why you have to clone the spends in the first place.

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -46,32 +46,54 @@ pub fn validate_clvm_and_signature(
...
+                    let hash: [u8; 32] = hasher.finalize();
+                    (hash, pairing)
                 })
             })
     });
-    let unsafe_items = npcresult
-        .agg_sig_unsafe
-        .iter()
-        .map(|(pk, msg)| (pk, msg.as_slice().to_vec()));
+    let unsafes = npcresult.agg_sig_unsafe.clone();
```
arvidn, line 68: I don't think you should clone this either

arvidn (commented):
looks good. just a few comments

crates/chia-consensus/src/spendbundle_validation.rs
matt-o-how: There's an unfortunate move that happens which means we can't return `npcresult` unless we clone it earlier

arvidn: I would expect you the lambda to take the entry by reference, and return a reference

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -46,32 +46,52 @@ pub fn validate_clvm_and_signature(
...
     let iter = iter.chain(unsafe_items);
+
     // Verify aggregated signature
-    let (result, added) = cache
-        .lock()
-        .unwrap()
-        .aggregate_verify(iter, &spend_bundle.aggregated_signature);
+    let result = aggregate_verify_gt(
+        &spend_bundle.aggregated_signature,
+        iter.clone().map(|tuple| tuple.1),
```
arvidn, line 77: Cloning the iterator here probably means you'll perform the work in the lambda twice, each time you iterate over it. It might be a good reason to `collect()` the `GTElement`s first. Alternatively, build the return vector as part of this lambda function. That would be the most efficient.

crates/chia-bls/src/gtelement.rs
```diff
@@ -13,7 +13,7 @@ use std::ops::{Mul, MulAssign};
     pyo3::pyclass,
     derive(chia_py_streamable_macro::PyStreamable)
 )]
-#[derive(Clone)]
+#[derive(Clone, Copy)]
```
arvidn, line 16: I believe the `GTElement` is more than 500 bytes large. I think it's pretty good to require explicit cloning, to see where we pay that cost.

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -41,50 +47,39 @@ pub fn validate_clvm_and_signature(
...
-    });
-    let iter = iter.chain(unsafe_items);
+        pairs.push((hash_pk_and_msg(&pk.to_bytes(), msg), pairing));
+    }
 
     // Verify aggregated signature
     let result = aggregate_verify_gt(
         &spend_bundle.aggregated_signature,
-        iter.clone().map(|tuple| tuple.1),
+        pairs.iter().map(|tuple| tuple.1),
```
arvidn, line 75:
```suggestion
        pairs.iter().map(|tuple| &tuple.1),
```
does this work?

arvidn (commented):
looks good. I would think the `&` allows you to not make `GTElement` derive from `Copy` though

PR #651, opened 2024-08-07
add higher level python binding for `get_puzzle_and_solution_for_coin()`

arvidn:
This new function (`get_puzzle_and_solution_for_coin2()`):
1. takes a `Program` as the generator, rather than `bytes`.
2. takes a list of block references rather than a CLVM encoding of the deserializer mod and the block references.
3. returns a tuple of `Program` rather than `bytes`, for the puzzle and solution
4. Takes a `Coin` rather than parent, amount and puzzle-hash.

Internally, the rust function to actually search the generator output for the specified coin can now also take advantage of caching the tree hash of nodes that have backrefs. This will likely speed up computing puzzle hashes somewhat.


matt-o-how (approved):

crates/chia-consensus/src/gen/get_puzzle_and_solution.rs
```diff
@@ -24,6 +25,7 @@ pub fn parse_coin_spend(
 pub fn get_puzzle_and_solution_for_coin(
     a: &Allocator,
     generator_result: NodePtr,
+    backrefs: &HashSet<NodePtr>,
     find_parent: Bytes32,
```
Rigidity, line 29: Can we just pass in a `Coin` here, rather than the 3 individual fields?

wheel/src/api.rs
```diff
@@ -170,6 +181,64 @@ pub fn get_puzzle_and_solution_for_coin<'a>(
...
+// The old version was written when Program was a python type had to be
+// serialized to bytes through rust boundary.
+#[allow(clippy::too_many_arguments)]
+#[pyfunction]
+pub fn get_puzzle_and_solution_for_coin2<'a>(
+    py: Python<'a>,
+    generator: &Program,
+    block_refs: &Bound<'a, PyList>,
+    max_cost: Cost,
+    find_parent: Bytes32,
```
Rigidity, line 195: Same here, using `Coin` would be less parameters

crates/chia-consensus/src/gen/get_puzzle_and_solution.rs
arvidn: yes! I forgot that `Coin` is also a rust type now :)

Rigidity (dismissed):

Rigidity (approved):

PR #656, opened 2024-08-08
CHIA-1034 Rename `Spend` to `SpendConditions` and `OwnedSpend` to `OwnedSpendConditions`

matt-o-how:


crates/chia-consensus/src/gen/owned_conditions.rs
```diff
@@ -14,7 +14,7 @@ use chia_py_streamable_macro::{PyJsonDict, PyStreamable};
     pyo3::pyclass(name = "Spend", get_all, frozen),
```
arvidn, line 14: you don't think the python name should be changed as well?

matt-o-how: Done, with the caveat that the chia-blockchain change is gonna have to land first

arvidn: you could also export it under both names

arvidn: in fact, I think it would be problematic to bump the `chia-blockchain` dependency. The version we use right now was carefully chosen to still contain python implementations of the classes we have ported to rust, so we can test against them

arvidn (changes requested):
`wheel/python/chia_rs/chia_rs.pyi` has line-ending changes, or someething

arvidn (approved):
it's important to squash this PR when landing it

PR #667, opened 2024-08-16
CHIA-1171: Make DID recovery list hash optional

Rigidity:
If the recovery list is empty, this saves 32 bytes of space in the puzzle reveal.


arvidn (approved):
the way this is decoded relies in matching the number of items and maybe that a `u64` is smaller than a `bytes32`, right?

Rigidity:
> the way this is decoded relies in matching the number of items and maybe that a `u64` is smaller than a `bytes32`, right?

Option is encoded as nil or bytes32 in this case

PR #668, opened 2024-08-17
CHIA-1169: Serde support for common types

Rigidity:


crates/chia-protocol/src/bytes.rs
```diff
@@ -19,10 +19,19 @@ use pyo3::prelude::*;
 #[cfg(feature = "py-bindings")]
 use pyo3::types::PyBytes;
 
+#[cfg(not(feature = "serde"))]
 #[derive(Default, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub struct Bytes(Vec<u8>);
 
+#[cfg(feature = "serde")]
```
Rigidity, line 27: I'm not a fan of duplicating the definition of `Bytes` and `BytesImpl`. It might be better to get rid of `serde_with` entirely and just use the manual implementation like I did for `PublicKey`

Cargo.toml
```diff
@@ -49,6 +49,7 @@ cast_possible_wrap = "allow"
 cast_lossless = "allow"
 similar_names = "allow"
 implicit_hasher = "allow"
+unsafe_derive_deserialize = "allow"
```
Rigidity, line 52: Not sure why it was giving this warning when deriving on types like `Coin`. Maybe the Python bindings are tripping it up

PR #673, opened 2024-08-21
`chia-datalayer`

altendky:
~Maybe this actually grows into something.  For now it is an opportunity to practice Rust within an existing Chia project.~

it grew...

partial testing of swappability with the python implementation is tested at https://github.com/Chia-Network/chia-blockchain/pull/18716.


crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,42 @@
+use std::collections::HashMap;
+
+type Index = usize;
+type Key = Vec<u8>;
```
arvidn, line 4: is the length of the key fixed? e.g. a 32 byte hash.
If so, you can make it `[u8; 32]`, which makes it allocated "in-line". i.e. not heap allocated.

altendky: Yeah, I think it does end up being that in this case.  it's really the key hash.  Thanks.

crates/chia-datalayer/src/lib.rs
```diff
@@ -1,42 +1,155 @@
...
+
+#[derive(Debug, PartialEq)]
 pub struct NodeMetadata {
     pub node_type: NodeType,
+    pub dirty: bool,
+}
+
+impl NodeMetadata {
+    // TODO: how do i say what i'm passing in is length two if i [u8; 2] here
+    pub fn load(blob: &[u8]) -> Self {
```
Rigidity, line 127: you can just make the argument `[u8; 2]` and pass it by value rather than reference, it'd be an array instead of a slice

crates/chia-datalayer/src/lib.rs
```diff
@@ -1,42 +1,155 @@
...
 
+// TODO: fill out related to the serializations
+const METADATA_SIZE: u32 = 2;
+const DATA_SIZE: u32 = 0;
+const SPACING: u32 = METADATA_SIZE + DATA_SIZE;
+
+impl MerkleBlob {
+    pub fn get_raw_node(&self, index: TreeIndex) -> RawMerkleNode {
+        // TODO: handle invalid indexes?
+        // TODO: handle overflows?
```
Rigidity, line 58: Overflows and invalid indices will cause a panic which will raise an error in Python and crash in Rust by default. You can instead use `.get(index)` to get an `Option` for example, and handle errors that way.

crates/chia-datalayer/src/lib.rs
```diff
@@ -1,42 +1,155 @@
-use std::collections::HashMap;
+// use std::collections::HashMap;
 
-type Index = usize;
-type Key = Vec<u8>;
+type TreeIndex = u32;
```
Rigidity, line 3: Instead of a type alias, you could make a newtype like this:
```rs
#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
struct TreeIndex(32);
```
This has the benefit that it will not be compatible with plain `u32` integers by default for type safety, and the actual index isn't exposed publicly. You can of course make the index `pub` if you need, or add an `fn index(&self) -> u32` method.

Docs on the newtype idiom:
https://doc.rust-lang.org/rust-by-example/generics/new_types.html

crates/chia-datalayer/src/lib.rs
```diff
@@ -1,42 +1,155 @@
...
+// type Key = Vec<u8>;
+type Hash = [u8; 32];
```
Rigidity, line 5: I don't think this type alias is useful

crates/chia-datalayer/src/lib.rs
```diff
@@ -1,42 +1,155 @@
...
+type KVId = Hash;
```
Rigidity, line 6: Similarly to `TreeIndex`, you could make a newtype called `KvId([u8; 32])` instead of a type alias.

Note the acronym `Kv` is PascalCase too:
https://rust-lang.github.io/api-guidelines/naming.html

crates/chia-datalayer/src/lib.rs
```diff
@@ -1,42 +1,155 @@
...
+
+#[derive(Debug, Hash, Eq, PartialEq)]
 
 pub enum NodeType {
```
Rigidity, line 10: This should also derive `Clone, Copy, Hash` IMO

crates/chia-datalayer/src/lib.rs
```diff
@@ -1,42 +1,155 @@
...
 }
 
+impl NodeType {
+    pub fn load(&value: &u8) -> Self {
+        // TODO: identify some useful structured serialization tooling we use
+        // TODO: find a better way to tie serialization values to enumerators
+        match value {
+            0 => NodeType::Internal,
+            1 => NodeType::Leaf,
+            other => panic!("unknown NodeType value: {}", other),
```
Rigidity, line 22: Instead of `panic` for deserialization, you can return `Option<Self>`

crates/chia-datalayer/src/lib.rs
```diff
@@ -1,42 +1,155 @@
...
+        key_value: KVId,
+        hash: Hash,
+        // TODO: kinda feels questionable having it be aware of its own location
+        index: TreeIndex,
+    },
+}
+
+impl RawMerkleNode {
+    // TODO: how do i say what i'm passing in is length two if i [u8; 2] here
+    pub fn load(metadata: NodeMetadata, index: TreeIndex, blob: &[u8]) -> Self {
```
Rigidity, line 98: Instead of `unwrap` everywhere when deserializing (which panics on errors), you could return a `Result` or `Option` to indicate an error case and let the caller handle it properly. Rust doesn't have exceptions so it's up to the library to propagate errors.

crates/chia-datalayer/src/lib.rs
altendky: I did want the type safety aspect but wasn't convinced by the newtype wrapper approach quite yet.  I do expect to try it out at some point and see how it feels for this.

crates/chia-datalayer/src/lib.rs
altendky: It might end up being a newtype as well, in time.  We'll see.

crates/chia-datalayer/src/lib.rs
altendky: case change incoming

crates/chia-datalayer/src/lib.rs
altendky: I have been adding the derives as I find a need.  I expect once I'm more familiar I will pre-emptively add some as well.

crates/chia-datalayer/src/lib.rs
altendky: I believe overflows only panic in debug builds?

Is indexing a case where I'm allowed to code a thing that might panic and I'm not given any warning to handle it instead?  So I should just always `.get()` instead of `[]` for indexing?  Though in this case the 'indexes' are used for slicing so I guess I should check how slices are handled and how to get them more safely.

crates/chia-datalayer/src/lib.rs
altendky: I grabbed this pattern from somewhere else in chia_rs just to get something working here.  It looked likely not a great way to do it so I'll come back at some point.

crates/chia-datalayer/src/lib.rs
altendky: changed.

altendky (commented):
thanks

crates/chia-datalayer/src/lib.rs
Rigidity: Overflows definitely always panic, otherwise it would be undefined behavior and a memory safety issue.
There are unchecked versions that don't panic, but they are marked as `unsafe` for those reasons.

I'd say if you don't know the length beforehand, just don't use indexing yeah.

crates/chia-datalayer/src/lib.rs
Rigidity: It's a really common pattern. For example, a `NodePtr` is a just a `u32`:
https://github.com/Chia-Network/clvm_rs/blob/main/src/allocator.rs#L14

The `.0` and whatnot is annoying at first but I got used to it.

crates/chia-datalayer/src/lib.rs
Rigidity: I did make a `TreeHash` type in `clvm-utils` for a similar reason :P

crates/chia-datalayer/src/lib.rs
Rigidity: Doesn't matter if nobody's using it I guess but a good habit to get into

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+        let mut block = Block::from_bytes(
+            self.get_block_bytes(old_leaf.index)?,
+            new_internal_node_index,
+        )?;
+        block.node.parent = Some(new_internal_node_index);
+        self.insert_entry_to_blob(old_leaf.index, block.to_bytes())?;
+
+        let mut old_parent_block =
+            Block::from_bytes(self.get_block_bytes(old_parent_index)?, old_parent_index)?;
+        match old_parent_block.node.specific {
```
arvidn, line 600: this would probably be a bit simpler as a let-else.

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+
+    pub fn delete(&mut self, key_value: KvId) -> Result<(), String> {
+        let leaf_index = *self.kv_to_index.get(&key_value).unwrap();
+        let leaf = self.get_node(leaf_index).unwrap();
+
+        match leaf.specific {
+            // TODO: blech
+            NodeSpecific::Leaf { .. } => (),
+            NodeSpecific::Internal { .. } => panic!(),
+        };
```
arvidn, line 632: I think this could be something like:
```
assert!(matches!(NodeSpecific::Leaf { ... }, leaf.specific));
```
But if there's a way to make the code just type-safe by construction, that would be ideal I think.

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+        match self.free_indexes.pop() {
+            None => {
+                self.last_allocated_index += 1;
+                self.last_allocated_index - 1
+            }
+            Some(new_index) => new_index,
+        }
+    }
+
+    fn get_random_leaf_node_from_bytes(&self, seed_bytes: Vec<u8>) -> Result<Node, String> {
```
arvidn, line 748: this looks like a test-only function. If it is, you could add a `#[cfg(test)]` to its definition

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+            dirty: Self::dirty_from_bytes(blob)?,
+        })
+    }
+
+    pub fn to_bytes(&self) -> [u8; METADATA_SIZE] {
+        [self.node_type.to_u8(), u8::from(self.dirty)]
+    }
+
+    pub fn node_type_from_bytes(blob: [u8; METADATA_SIZE]) -> Result<NodeType, String> {
+        NodeType::from_u8(blob[0])
```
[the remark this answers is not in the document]
altendky: i intend to review all panics, unwraps, expects, direct indexing.  i poke and did some verbose error handling for a few lines early on and have since been lazy.  i need to take another pass at getting myself thinking sensibly about the error handling.

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+    }
+
+    pub fn push(&mut self, mut other: DotLines) {
+        self.nodes.append(&mut other.nodes);
+        self.connections.append(&mut other.connections);
+        self.pair_boxes.append(&mut other.pair_boxes);
+    }
+
+    pub fn dump(&mut self) -> String {
+        // TODO: consuming itself, secretly
```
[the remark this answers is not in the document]
altendky: that's a todo saying i don't like how this works.  but i didn't get it working how i wanted and this worked sufficiently and i moved on to other bits with a note to revisit.  the appends are consuming the insides of `self` out of it right now.  so `self` is now garbage, but rust doesn't know that because of the function signature.  but i don't actually want to consume it in any form.

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+const NULL_PARENT: TreeIndex = 0xffff_ffffu32;
+
+#[derive(Debug, PartialEq)]
+pub struct NodeMetadata {
+    pub node_type: NodeType,
+    pub dirty: bool,
+}
+
+impl NodeMetadata {
+    pub fn from_bytes(blob: [u8; METADATA_SIZE]) -> Result<Self, String> {
```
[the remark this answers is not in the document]
altendky: that was a thought and i haven't looked at how that would work out.  i stuck with this low level for at least awhile because i wanted to get some familiarity at this layer.  i (think i) do in places actually deserialize just individual fields which i'm guessing streamable doesn't support.  so it may be that i have a (pre-optimization) reason to not use streamable.  but yes, this is a good thing to reconsider prior to considering this code 'done'.

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+impl NodeSpecific {
+    pub fn sibling_index(&self, index: TreeIndex) -> TreeIndex {
+        let NodeSpecific::Internal { right, left } = self else {
+            panic!()
+        };
+
+        match index {
+            x if (x == *right) => *left,
+            x if (x == *left) => *right,
+            _ => panic!(),
```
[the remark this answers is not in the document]
altendky: when traversing up you know your index and sometimes you need your sibling.  you don't know whether your sibling is on the left or right until you check which side of your parent you are on.  hence taking the index instead of a side directly.

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+        }
+    }
+}
+
+const PARENT_RANGE: Range<usize> = 0..4;
+const LEFT_RANGE: Range<usize> = 4..8;
+const RIGHT_RANGE: Range<usize> = 8..12;
+const KEY_VALUE_RANGE: Range<usize> = 4..12;
+// TODO: move the common parts to the beginning of the serialization?
+const HASH_RANGE: Range<usize> = 12..44;
```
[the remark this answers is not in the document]
altendky: yes it is.  poorly...  `:]`  handily both of the types of nodes just so happen to have the same serialized length so i can just pretend that no padding is ever, nor ever will be, needed.  this goes to your streamable comment and considering how serialization should _really_ be handled now that i've gotten some experience at dealing with byte packing and unpacking.

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+                blob.extend(hash);
+            }
+        }
+
+        blob.try_into().unwrap()
+    }
+
+    // TODO: yes i know i'm trying to write this code in a non-rusty way and i need to stop that
+    pub fn key_value(&self) -> KvId {
+        let NodeSpecific::Leaf { key_value } = self.specific else {
```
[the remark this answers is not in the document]
altendky: then you have a pile of silliness (more of these panic'ing methods) to get at the common field like the parent and hash.  i started that way and then decided to shift this way.  there were a few other options considered along the way that might still be good to think about.  in this form i do still pattern match, just against this specific bit rather than the whole node.

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+}
+
+fn get_free_indexes(blob: &[u8]) -> Result<Vec<TreeIndex>, String> {
+    let index_count = blob.len() / BLOCK_SIZE;
+
+    if index_count == 0 {
+        return Ok(vec![]);
+    }
+
+    let mut seen_indexes: Vec<bool> = vec![false; index_count];
```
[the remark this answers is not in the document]
altendky: i asked about a bunch of stuff, but didn't ask this as i wondered about bit-packing tooling, or in this case tooling that happens to bit-pack behind the scenes.  so, what would you reach for here?  or just a manual vec of 1/8 the size and bit masking?

crates/chia-datalayer/src/lib.rs
altendky: i bet you can imagine the python source that i was copying when i wrote this.  `:]`  type narrowing and all.  but yes, the `blech` is saying i don't like this and should revisit to try to do something else.  i was definitely hitting this due to some bugs though.

crates/chia-datalayer/src/lib.rs
altendky: it's called from `.insert()`

```rust
        // TODO: make this a parameter so we have one insert call where you specify the location
        let old_leaf = self.get_random_leaf_node_from_bytes(Vec::from(key_value.to_be_bytes()))?;
```

though, as per the comment, i think i want the outer layer (python datalayer code) to just first call this 'get a leaf' method and then pass it to the insert function.  then we don't need an 'insert to an automatic location' method and a separate (or parameter controlled) 'insert to this specific location' method.  just one method to auto-pick a location and one method to insert to a specific location.

crates/chia-datalayer/src/lib.rs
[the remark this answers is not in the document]
arvidn: if you just want to consume the vectors of `self`, you could use `self.nodes.drain()`.

crates/chia-datalayer/src/lib.rs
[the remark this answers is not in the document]
arvidn: I believe there are crates implementing bitfields that are a bit easier to use than a "manual" byte array, but I don't recall having explored them

crates/chia-datalayer/src/lib.rs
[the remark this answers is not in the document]
Rigidity: Sounds like he's wanting to clone, rather than drain?

crates/chia-datalayer/src/lib.rs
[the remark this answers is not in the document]
Rigidity: Perhaps
https://docs.rs/bitflags/1.0.4/bitflags/

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+        let block =
+            Block::from_bytes(blob[offset..offset + BLOCK_SIZE].try_into().unwrap(), index)?;
+        seen_indexes[index as usize] = true;
+        match block.node.specific {
+            NodeSpecific::Internal { left, right } => {
+                queue.push(left);
+                queue.push(right);
+            }
+            NodeSpecific::Leaf { .. } => (),
+        }
```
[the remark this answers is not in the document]
Rigidity: if let but yes

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+    while let Some(index) = queue.pop() {
+        let offset = index as usize * BLOCK_SIZE;
+        let block =
+            Block::from_bytes(blob[offset..offset + BLOCK_SIZE].try_into().unwrap(), index)?;
+        match block.node.specific {
+            NodeSpecific::Leaf { key_value } => {
+                kv_to_index.insert(key_value, index);
+            }
+            NodeSpecific::Internal { .. } => (),
+        }
```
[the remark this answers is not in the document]
Rigidity: ditto `if let`

crates/chia-datalayer/src/lib.rs
Rigidity: For future reference (sounds maybe not applicable here) I think all test related stuff should go inside a test module, so you don't have to repeat `#[cfg(test)]`

crates/chia-datalayer/src/lib.rs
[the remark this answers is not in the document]
altendky: 85aefbc4d4e827d5ad541d758e6c8df973c1c2b7 slightly less poor description (both in terms of code and comments).  but still quite 'low level' at least for now.

crates/chia-datalayer/src/lib.rs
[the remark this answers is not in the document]
altendky: one of the things that runs through my mind with this is keeping the `match` must-handle-every-case checks.  i haven't thought through the details, but i considered having another node type for the root (which would not have a parent) instead of having an `Option` parent as is now.  using `match` would just remind me of every place i need to review if i were to refactor that.

but regardless, it's a good reminder to me to think about the non-match options and i'll review this and related cases now.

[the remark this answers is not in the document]
altendky: ok, so that refactor sounds silly as i would have to then have two root types.  both internal and leaf. so probably not doing that...  anyways, changed a few in 28548288a3a4894be7d09f99f8ee584ae2508ca4

crates/chia-datalayer/src/lib.rs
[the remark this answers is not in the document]
altendky: 28548288a3a4894be7d09f99f8ee584ae2508ca4 probably

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+pub struct MerkleBlob {
+    blob: Vec<u8>,
+    free_indexes: Vec<TreeIndex>,
+    kv_to_index: HashMap<KvId, TreeIndex>,
+    // TODO: maybe name it next_index_to_allocate
+    last_allocated_index: TreeIndex,
+}
+
+impl MerkleBlob {
+    pub fn new(blob: Vec<u8>) -> Result<Self, String> {
```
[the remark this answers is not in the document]
altendky: and then make it copy inside?  or are you suggesting changing the actual field type as well?  i was going to ask at some point what the options are around this.  i could change size by creating a new array as needed, for example.  or maybe some other type etc.  maybe someday even use a memory mapped file for really large trees on low-ram systems (wild ideas).

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,1259 @@
...
+
+        Ok(Self {
+            blob,
+            free_indexes,
+            kv_to_index,
+            last_allocated_index: block_count as TreeIndex,
+        })
+    }
+
+    pub fn insert(&mut self, key_value: KvId, hash: Hash) -> Result<(), String> {
```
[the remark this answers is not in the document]
altendky: cb90651430f0c2946e018c8e4646070419602de8

crates/chia-datalayer/src/lib.rs
altendky: probably 28548288a3a4894be7d09f99f8ee584ae2508ca4

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1997 @@
...
+
+use clvmr::sha2::Sha256;
+use num_traits::ToBytes;
+use std::cmp::Ordering;
+use std::collections::{HashMap, HashSet, VecDeque};
+use std::iter::zip;
+use std::mem::size_of;
+use std::ops::Range;
+
+type TreeIndex = u32;
```
Rigidity, line 16: I think this should be a newtype:

```rs
#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct TreeIndex(u32);
```

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1997 @@
...
+type Parent = Option<TreeIndex>;
```
Rigidity, line 17: This doesn't seem like a useful type alias to me

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1997 @@
...
+type Hash = [u8; 32];
```
Rigidity, line 18: Presumably you could just use `[u8; 32]` directly or make this a newtype?

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1997 @@
...
+type KvId = i64;
```
Rigidity, line 19: I'm not sure what this is for, maybe a comment explaining it and/or a newtype with a doc comment would be appropriate for this as well.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1997 @@
...
+
+// assumptions
+// - root is at index 0
+// - any case with no keys will have a zero length blob
+
+const fn range_by_length(start: usize, length: usize) -> Range<usize> {
+    start..start + length
+}
+const fn max(left: usize, right: usize) -> usize {
```
Rigidity, line 28: https://doc.rust-lang.org/std/cmp/fn.max.html

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1997 @@
...
+type TreeIndex = u32;
+type Parent = Option<TreeIndex>;
+type Hash = [u8; 32];
+type KvId = i64;
+
+// assumptions
+// - root is at index 0
+// - any case with no keys will have a zero length blob
+
+const fn range_by_length(start: usize, length: usize) -> Range<usize> {
```
Rigidity, line 25: Is this really an improvement?

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1997 @@
...
+pub enum NodeType {
+    Internal = 0,
+    Leaf = 1,
+}
+
+impl NodeType {
+    pub fn from_u8(value: u8) -> Result<Self, String> {
+        match value {
+            // ha!  feel free to laugh at this
+            x if (NodeType::Internal as u8 == x) => Ok(NodeType::Internal),
```
Rigidity, line 70: I really think this should either be `0 => Ok(NodeType::Internal)` and so on, or use a macro to generate both conversions. Same with `to_u8`.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1997 @@
...
+// #[cfg_attr(feature = "py-bindings", pyclass(name = "NodeSpecific"))]
+#[cfg_attr(feature = "py-bindings", pyclass(name = "NodeSpecific", get_all))]
+#[derive(Clone, Debug, PartialEq)]
+pub enum NodeSpecific {
+    Internal { left: TreeIndex, right: TreeIndex },
+    Leaf { key: KvId, value: KvId },
+}
+
+impl NodeSpecific {
+    // TODO: methods that only handle one variant seem kinda smelly to me, am i right?
```
Rigidity, line 181: Yes these enum variants should be factored out into structs if you need to add special methods on one variant.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1997 @@
...
+    pub fn sibling_index(&self, index: TreeIndex) -> TreeIndex {
+        let NodeSpecific::Internal { right, left } = self else {
+            panic!("unable to get sibling index from a leaf")
+        };
+
+        match index {
+            x if (x == *right) => *left,
```
Rigidity, line 188:
```rs
if index == *right {
    *left
} else if index == *left {
    *right
} else {
    ...
}
```

Rigidity (commented):
Instead of using `String` errors or panicking, you should create an enum with `thiserror`:

```rs
#[derive(Debug, Error)]
pub enum Error {
    #[error("This is the first error")]
    First,
    
    #[error("Foo error: {0}")]
    Second(#[from] FooError),
}
```

And if you're going to write your own serialization, I think you should at least create a trait (and possibly derive macro) for it.

crates/chia-datalayer/src/merkle.rs
altendky: The intent is to explicitly remind folks that a parent is not known to be 'something' and that we are _not_ using a magic value like a `TreeIndex` of `-1` or something to represent the no-parent case.

I would expect this to be caught up in a general switch to the newtype pattern so even if I convinced you of the value, let's leave this discussion open.

crates/chia-datalayer/src/merkle.rs
altendky: f5ee0c87599003ffa2cb5f6a6201e62d0472488f

crates/chia-datalayer/src/merkle.rs
Rigidity: oh, interesting

crates/chia-datalayer/src/merkle.rs
altendky: newtype at [`58c79a2` (#673)](https://github.com/Chia-Network/chia_rs/pull/673/commits/58c79a290017a08030a60c427a1f7bd0ce891be5)

crates/chia-datalayer/src/merkle.rs
altendky: macro in [`825f7ac` (#673)](https://github.com/Chia-Network/chia_rs/pull/673/commits/825f7ac3c3ec2af506dbb9559ef9f380c924c053)

crates/chia-datalayer/src/merkle.rs
altendky: removed in [`9ab4599` (#673)](https://github.com/Chia-Network/chia_rs/pull/673/commits/9ab45995f33984bd2f02728b1343bf3c8fac7b0a)

crates/chia-datalayer/src/merkle.rs
altendky: switched to `Bytes32` while changing to streamable serialization.  [`9ab4599` (#673)](https://github.com/Chia-Network/chia_rs/pull/673/commits/9ab45995f33984bd2f02728b1343bf3c8fac7b0a)  i do presently still have `type Hash = Bytes32` but will probably remove that.  didn't want quite so much noise in the streamable pr.

crates/chia-datalayer/src/lib.rs
altendky: I think this is all gone at this point.  Or at least wrapped up a bit.  https://github.com/Chia-Network/chia_rs/pull/673/files#diff-a48504f4b97458ce535cd911838701acf1c6b739aeb9feac35a52688c432ec6aR776-R828

crates/chia-datalayer/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+mod merkle;
+
+pub use merkle::{InsertLocation, InternalNode, LeafNode, MerkleBlob, Side};
```
Rigidity, line 3: Does this need to export more things? Like `TreeIndex`, `KvId`, `Block`, `Node`, etc? Might be simpler just to do `pub use merkle::*;`

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1902 @@
...
+    Internal = 0,
+    Leaf = 1,
+}
+
+impl NodeType {
+    pub fn from_u8(value: u8) -> Result<Self, Error> {
+        streamable_from_bytes_ignore_extra_bytes(&[value])
+    }
+
+    #[allow(clippy::wrong_self_convention, clippy::trivially_copy_pass_by_ref)]
```
Rigidity, line 143: This lint exists for a reason, taking it by reference isn't idiomatic and uses 8 bytes for a pointer instead of 1 for the enum itself. You should take self by value here.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1902 @@
...
+fn internal_hash(left_hash: &Hash, right_hash: &Hash) -> Hash {
+    let mut hasher = Sha256::new();
+    hasher.update(b"\x02");
+    hasher.update(left_hash);
+    hasher.update(right_hash);
+
+    Bytes32::new(hasher.finalize())
+}
+
+#[cfg_attr(feature = "py-bindings", pyclass(name = "Side", eq, eq_int))]
```
Rigidity, line 173: You shouldn't need to set the name here since it's going to default to the struct name.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1902 @@
...
+#[repr(u8)]
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Streamable)]
+pub enum Side {
+    Left = 0,
+    Right = 1,
+}
+
+#[cfg_attr(feature = "py-bindings", pyclass(name = "InsertLocation"))]
```
Rigidity, line 181: This name is also implied. There are several other so I won't put a comment on all of them.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1902 @@
...
+pub struct NodeMetadata {
+    // OPT: could save 1-2% of tree space by packing (and maybe don't do that)
+    pub node_type: NodeType,
+    pub dirty: bool,
+}
+
+#[cfg_attr(feature = "py-bindings", pyclass(name = "InternalNode", get_all))]
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, Streamable)]
+pub struct InternalNode {
+    parent: Parent,
```
Rigidity, line 202: Maybe these fields should be public? Or are there invariants that need to be upheld - in that case, you could add getters on the Rust side instead.

I know this is for a Python binding but we should make sure the Rust API is solid too.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1902 @@
...
+        } else {
+            Err(Error::IndexIsNotAChild(index))
+        }
+    }
+}
+
+#[cfg_attr(feature = "py-bindings", pyclass(name = "LeafNode", get_all))]
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, Streamable)]
+pub struct LeafNode {
+    parent: Parent,
```
Rigidity, line 223: Same comment as `InternalNode` here.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1902 @@
...
+    }
+
+    fn hash(&self) -> Hash {
+        match self {
+            Node::Internal(node) => node.hash,
+            Node::Leaf(node) => node.hash,
+        }
+    }
+
+    fn set_hash(&mut self, hash: &Hash) {
```
Rigidity, line 257: You should take ownership of `hash` here so you don't need to dereference it.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1902 @@
...
+            Node::Internal(node) => node.to_bytes(),
+            Node::Leaf(node) => node.to_bytes(),
+        }
+        .map_err(Error::Streaming)?;
+        assert!(base.len() <= DATA_SIZE);
+        base.resize(DATA_SIZE, 0);
+        Ok(base
+            .as_slice()
+            .try_into()
+            .expect("padding was added above, might be too large"))
```
Rigidity, line 283: If this error should never occur in practice, this is fine. But otherwise, it should be part of the `Error` enum.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1902 @@
...
+    }
+
+    fn expect_leaf(&self, message: &str) -> LeafNode {
```
Rigidity, line 286: What is this?

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1902 @@
...
+        if let Node::Internal(ref mut internal_node, ..) = old_parent_block.node {
+            if old_leaf_index == internal_node.left {
+                internal_node.left = new_internal_node_index;
+            } else if old_leaf_index == internal_node.right {
+                internal_node.right = new_internal_node_index;
+            } else {
+                panic!("child not a child of its parent");
+            }
+        } else {
+            panic!("expected internal node but found leaf");
```
Rigidity, line 601: Are all these panics unreachable? You should use `unreachable!()` instead to indicate this. Otherwise, use `Error`.

crates/chia-datalayer/src/merkle.rs
altendky: It's purpose was to make it easy to express that this should indeed be a leaf variant and to panic if not.  Reviewing the uses there are only two in runtime code.  Those two uses would panic if there was a coding bug resulting in corrupted data or cache.  For example, we get an index from our key -> index mapping and then instead of getting a leaf node at that index (key/value node) we get an internal node.  Or an iterator yielding a block but which only yields leaf blocks.

crates/chia-datalayer/src/merkle.rs
altendky: I think I do have some unreachable code somewhere that I wasn't able to get rid of via typing.  I don't think I would categorize either of these specifically as unreachable though.  Both of these asserts could come from data corruption elsewhere in the code I think.

crates/chia-datalayer/src/merkle.rs
altendky: made public in 0be9f5015b1854477cd3bd3633b09082efc89487

crates/chia-datalayer/src/merkle.rs
altendky: made public in 0be9f5015b1854477cd3bd3633b09082efc89487

crates/chia-datalayer/src/merkle.rs
Rigidity: nope just wanted to make sure, looks good then

crates/chia-datalayer/src/lib.rs
altendky: changed as suggested 6904bcd2cd1e3accc6bc8d2aef9ebb114412d514

Rigidity (approved):

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1897 @@
...
+            free_indexes.insert(TreeIndex(index as u32));
+        }
+    }
+
+    (free_indexes, key_to_index)
+}
+
+#[cfg_attr(feature = "py-bindings", pyclass(get_all))]
+#[derive(Debug)]
+pub struct MerkleBlob {
```
arvidn, line 367: this name isn't super intuitive to me. the fact that so many operations call `clear()` also makes it hard for me to understand what the behaviour is expected to be.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1897 @@
...
+fn get_free_indexes_and_keys_values_indexes(
+    blob: &[u8],
+) -> (HashSet<TreeIndex>, HashMap<KvId, TreeIndex>) {
+    let index_count = blob.len() / BLOCK_SIZE;
+
+    let mut seen_indexes: Vec<bool> = vec![false; index_count];
+    let mut key_to_index: HashMap<KvId, TreeIndex> = HashMap::default();
+
+    for (index, block) in MerkleBlobLeftChildFirstIterator::new(blob) {
+        seen_indexes[index.0 as usize] = true;
```
arvidn, line 348: are we OK with a `panic!` here, if `index.0` is out of range?

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1897 @@
...
+            match block.node {
+                Node::Leaf(..) => return Some(block),
+                Node::Internal(node) => {
+                    self.deque.push_back(node.left);
+                    self.deque.push_back(node.right);
+                }
+            }
+        }
+    }
+}
```
arvidn, line 1430: it doesn't look like these iterators have very thorough tests

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1897 @@
...
+            },
+            node: Node::Leaf(LeafNode {
+                parent: None,
+                key,
+                value,
+                hash: *hash,
+            }),
+        };
+
+        self.clear();
```
arvidn, line 462: it's not obvious to me why `insert_first()` clears everything in `Self`, along with the other functions that also `clear()` self.

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1897 @@
...
+                }
+            }
+        }
+    }
+}
+
+#[cfg(any(test, debug_assertions))]
+impl Drop for MerkleBlob {
+    fn drop(&mut self) {
+        self.check_integrity().unwrap();
```
arvidn, line 1435: I would think this `unwrap()` would be better as an `expect()` to get some helpful message

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1897 @@
...
+        self.insert_entry_to_blob(leaf_index, &block)?;
+
+        if let Some(parent) = block.node.parent() {
+            self.mark_lineage_as_dirty(parent)?;
+        }
+
+        Ok(())
+    }
+
+    pub fn check_integrity(&self) -> Result<(), Error> {
```
arvidn, line 843: is this a test-only function? should it be marked with `#[cfg(test)]`?

crates/chia-datalayer/src/merkle/dot.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        result
+    }
+}
+
+// TODO: better conditional execution than the commenting i'm doing now
+#[allow(unused)]
+pub fn open_dot(lines: &mut DotLines) {
+    let mut url = Url::parse("http://edotor.net").unwrap();
```
arvidn, line 104: this looks a bit scary. I take it you prefer this service over just running `dot`

arvidn (commented):
it's a pretty big patch and it seems it's sparsely tested.
I think it would be really good to fuzz some of these endpoint. Especially the ones parsing merkle tree blobs, but possibly also ones that operate on trees, where the tree could be built using `Arbitrary` and the fuzzer input

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1997 @@
...
+
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
+#[repr(u8)]
+pub enum NodeType {
+    Internal = 0,
+    Leaf = 1,
+}
+
+impl NodeType {
+    pub fn from_u8(value: u8) -> Result<Self, String> {
```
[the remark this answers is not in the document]
altendky: `NodeType` was simplified.

```rust
#[repr(u8)]
#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, Streamable)]
pub enum NodeType {
    Internal = 0,
    Leaf = 1,
}
```

crates/chia-datalayer/src/merkle.rs
```diff
@@ -0,0 +1,1997 @@
...
+const DATA_RANGE: Range<usize> = METADATA_SIZE..METADATA_SIZE + DATA_SIZE;
+
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
+#[repr(u8)]
+pub enum NodeType {
+    Internal = 0,
+    Leaf = 1,
+}
+
+impl NodeType {
```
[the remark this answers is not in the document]
altendky: `NodeType` was simplified.

```rust
#[repr(u8)]
#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, Streamable)]
pub enum NodeType {
    Internal = 0,
    Leaf = 1,
}
```

crates/chia-datalayer/src/merkle.rs
altendky: the call here in `.insert_first()` is theoretically pointless.  it does stand a chance of cleaning up any state left in the caches, but it is likely quite uncommon that folks end up emptying their store anyways so this is probably more confusing than valuable at all.  removed in 7867000996e0824b2de457528e5cac091e54ab0d.

in `.add_second()` the existing code just starts a fresh tree and adds back the data from the single existing block.  when adding a second leaf, the first leaf is presently 'moved', a new internal node created as the root, and an additional leaf added.  basically, nothing is staying the same.  so, we could remove the existing node from the key to index cache etc, or we could just wipe 'everything' (one thing) and put stuff where we want it.

in `.delete()`, the use of `.clear()` occurs only when the last leaf is deleted and the tree is empty.  again, we could carefully clean each cache, or just wipe them all because we know there should be nothing.

i do kind of feel like there's opportunity for simplification of the various logic around inserting and deleting.  that may yield better patterns of code without these corners that use `.clear()` for simplicity.

crates/chia-datalayer/src/merkle.rs
altendky: i added a small doc comment in be407fd4bd5aa3a4cd7403e8d15ff4df724db6b8.  the uses of `.clear()` are addressed directly in another comment.  let me know how much operational detail you feel belongs in the doc comment.

as to the name itself, my distaste for it has grown.  i'm happy to consider another name.

crates/chia-datalayer/src/merkle.rs
arvidn: the all the indices being valid is an invariant that is maintained (unless there's a bug)

crates/chia-datalayer/src/merkle.rs
arvidn: looks good. if it breaks in the future, it doesn't look like CI would tell us though

crates/chia-datalayer/src/merkle.rs
arvidn: I see, so what's being cleared is a kind of cache of parsed tree structure. and when we change the view to look at a different part of the tree, we clear the cache and parse that part. is that right?

arvidn (approved):

arvidn:
test coverage of `merkle.rs` is not excellent. (88%).
https://coveralls.io/builds/71236100/source?filename=crates%2Fchia-datalayer%2Fsrc%2Fmerkle.rs#L24
(note that you need to log in to coverall with your github account to see source code coverage)

altendky:
https://github.com/actions/dependency-review-action/issues/854

crates/chia-datalayer/src/merkle.rs
```diff
@@ -398,8 +398,12 @@ fn get_free_indexes_and_keys_values_indexes(
 #[derive(Debug)]
 pub struct MerkleBlob {
     blob: Vec<u8>,
+    // TODO: would be nice for this to be deterministic ala a fifo set
     free_indexes: HashSet<TreeIndex>,
```
arvidn, line 402: can this be a `Vec<TreeIndex>`? I would expect all you do is push and pop indices, no?

crates/chia-datalayer/src/merkle.rs
```diff
@@ -398,8 +398,12 @@ fn get_free_indexes_and_keys_values_indexes(
...
     key_to_index: HashMap<KvId, TreeIndex>,
+    // TODO: used by fuzzing, some cleaner way?  making it cfg-dependent is annoying with
+    //       the type stubs
+    pub check_integrity_on_drop: bool,
```
arvidn, line 406: would you have the same type-stub issue if you made it depend on `debug_assertions` configuration? i.e. debug builds.

crates/chia-datalayer/src/merkle.rs
altendky: also membership checks.  i guess the _free_ indexes are likely often a small group but i was looking into a more complete solution.

crates/chia-datalayer/src/merkle.rs
altendky: i think that a `maturin develop` was perhaps leading to the compilcation with this present while the wheel build did not.  maybe.  i had it on `#[cfg(any(test, debug_assertions))]` which is the enabler for the drop implementation itself.  the fuzzer docs also talk about a fuzzer cfg though using it triggered errors.  i do want to revisit this to better understand the various cases and make it cleaner.

arvidn (approved):

arvidn (approved):

PR #675, opened 2024-08-21
Add Python bindings to chiavdf and chiapos

Rigidity:


crates/chia-consensus/src/vdf.rs
```diff
@@ -0,0 +1,28 @@
...
+    let proof_buf = [info.output.data.as_slice(), proof.witness.as_slice()].concat();
+
+    verify_n_wesolowski(
+        &discriminant,
+        &input_el.data,
+        &proof_buf,
+        info.number_of_iterations,
+        proof.witness_type as u64,
+    )
+}
```
arvidn, line 28: A few tests would be good here too. At least covering both sides of this condition `if proof.witness_type + 1 > constants.max_vdf_witness_size`

arvidn (commented):
could you copy-paste some python test from chia-blockchain or chiapos and chiavdf?

PR #677, opened 2024-08-26
Improve CI workflows

Rigidity:
Makes the following changes:
* Splits `build-crate-and-npm` into two separate workflows
* Improves Rust tests by covering the whole workspace in both release and debug mode, with all features
* Removes redundant build check for `chia-tools`, since it's part of the workspace
* Adds `cargo-machete` to check unused dependencies (which have also been removed)
* Uses `cargo-lcov` and `pytest-cov` for coverage instead of `grcov`, and uploads two separate reports
* Other small changes to the workflows


arvidn:
these seems like really useful changes. They would be much easier to review if split up into separate PRs

PR #678, opened 2024-08-26
Switch to #[classmethod] for SpendBundle::py_aggregate

matt-o-how:
This will allow derived classes to return themselves rather than the SpendBundle parent class


Rigidity:
Did you change the line endings in these files? The diff seems way too large

crates/chia-protocol/src/spend_bundle.rs
```diff
@@ -94,10 +96,10 @@ impl SpendBundle {
 #[pymethods]
 #[allow(clippy::needless_pass_by_value)]
 impl SpendBundle {
-    #[staticmethod]
+    #[classmethod]
     #[pyo3(name = "aggregate")]
-    fn py_aggregate(spend_bundles: Vec<SpendBundle>) -> SpendBundle {
-        SpendBundle::aggregate(&spend_bundles)
+    fn py_aggregate(_cls: &Bound<'_, PyType>, spend_bundles: Vec<Self>) -> Self {
+        Self::aggregate(&spend_bundles)
```
arvidn, line 102: `Self` is just an alias for `SpendBundle` here

arvidn (commented):
I would have thought you'd need to use the passed-in `cls` type eomshow

wheel/generate_type_stubs.py
```diff
@@ -55,19 +55,19 @@ def __repr__(self) -> str: ...
...
     def __deepcopy__(self) -> {name}: ...
     def __copy__(self) -> {name}: ...
-    @staticmethod
-    def from_bytes(bytes) -> {name}: ...
-    @staticmethod
-    def from_bytes_unchecked(bytes) -> {name}: ...
-    @staticmethod
-    def parse_rust(ReadableBuffer, bool = False) -> Tuple[{name}, int]: ...
+    @classmethod
+    def from_bytes(cls, bytes) -> {name}: ...
```
arvidn, line 59: Does `cls` not need a type? something like `Type[X]` where `X = TypeVar("X", bound="{name}")`.
I would also expect the return value to be `X`
In practice, `X` would need a unique name, probably derived from `{name}` to avoid collisions.

matt-o-how: I don't believe so.
`cls` is a special keyword for class methods that is automatically passed in when it's called, like `self`.

Maybe I'm misunderstanding what you mean though?

arvidn: take this function as an example. the return value is not (necessarily) `{name}`. It's an object of the type `cls`

arvidn: and we know `cls` is a type that is `{name}` or something that derives from it, so we can define it as `bound`

arvidn:
we should try going back to invoking the derived class with all the fields of the class itself, as if it is a `dataclass`

crates/chia-protocol/src/coin_spend.rs
```diff
@@ -2,10 +2,29 @@ use chia_streamable_macro::streamable;
...
     puzzle_reveal: Program,
     solution: Program,
 }
+
+#[cfg(feature = "py-bindings")]
+#[pymethods]
+impl CoinSpend {
+    #[classmethod]
+    #[pyo3(name = "from_parent")]
+    pub fn from_parent(cls: &Bound<'_, PyType>, cs: Self) -> PyResult<PyObject> {
```
arvidn, line 22: actually, this *does* make me a bit nervous about performance impact. parsing coins is in the critical path in some places

crates/chia-protocol/src/spend_bundle.rs
```diff
@@ -94,10 +96,33 @@ impl SpendBundle {
...
+            )?;
+
+            Ok(instance.into_py(py))
+        })
+    }
+
+    #[classmethod]
+    #[pyo3(name = "from_parent")]
+    pub fn from_parent(cls: &Bound<'_, PyType>, spend_bundle: Self) -> PyResult<PyObject> {
+        Python::with_gil(|py| {
```
arvidn, line 117: I think you can just take a `Python` object as a parameter to get access to the GIL, rather than wrapping everything in `Python::with_gil()`. But maybe you've explored that.

crates/chia_py_streamable_macro/src/lib.rs
```diff
@@ -159,10 +159,23 @@ pub fn py_streamable_macro(input: proc_macro::TokenStream) -> proc_macro::TokenS
...
         impl #ident {
-            #[staticmethod]
+            #[classmethod]
             #[pyo3(signature=(json_dict))]
-            pub fn from_json_dict(json_dict: &pyo3::Bound<pyo3::PyAny>) -> pyo3::PyResult<Self> {
-                <Self as #crate_name::from_json_dict::FromJsonDict>::from_json_dict(json_dict)
+            pub fn from_json_dict(cls: &pyo3::Bound<'_, pyo3::types::PyType>, json_dict: &pyo3::Bound<pyo3::PyAny>) -> pyo3::PyResult<pyo3::PyObject> {
+                use pyo3::prelude::PyAnyMethods;
+                use pyo3::IntoPy;
+                let rust_obj = <Self as #crate_name::from_json_dict::FromJsonDict>::from_json_dict(json_dict);
```
arvidn, line 167: I would expect that a `?` at the end of this statement would work. Then you don't need to `match` the `rust_obj` to handle the `Err()` case

crates/chia_py_streamable_macro/src/lib.rs
```diff
@@ -174,34 +187,60 @@ pub fn py_streamable_macro(input: proc_macro::TokenStream) -> proc_macro::TokenS
...
+                use pyo3::prelude::PyAnyMethods;
+                use pyo3::IntoPy;
                 if !blob.is_c_contiguous() {
                     panic!("from_bytes() must be called with a contiguous buffer");
                 }
                 let slice = unsafe {
                     std::slice::from_raw_parts(blob.buf_ptr() as *const u8, blob.len_bytes())
                 };
-                <Self as #crate_name::Streamable>::from_bytes(slice).map_err(|e| <#crate_name::chia_error::Error as Into<pyo3::PyErr>>::into(e))
+                let rust_obj = <Self as #crate_name::Streamable>::from_bytes(slice);
```
arvidn, line 201: you can use `.map_err()?` here, instead of matching below

crates/chia_py_streamable_macro/src/lib.rs
```diff
@@ -174,34 +187,60 @@ pub fn py_streamable_macro(input: proc_macro::TokenStream) -> proc_macro::TokenS
...
+                use pyo3::prelude::PyAnyMethods;
+                use pyo3::IntoPy;
                 if !blob.is_c_contiguous() {
                     panic!("from_bytes_unchecked() must be called with a contiguous buffer");
                 }
                 let slice = unsafe {
                     std::slice::from_raw_parts(blob.buf_ptr() as *const u8, blob.len_bytes())
                 };
-                <Self as #crate_name::Streamable>::from_bytes_unchecked(slice).map_err(|e| <#crate_name::chia_error::Error as Into<pyo3::PyErr>>::into(e))
+                let rust_obj = <Self as #crate_name::Streamable>::from_bytes_unchecked(slice).map_err(|e| <#crate_name::chia_error::Error as Into<pyo3::PyErr>>::into(e));
```
arvidn, line 226: I would expect just `?` would work here as well.

wheel/generate_type_stubs.py
arvidn: I still think this is relevant. Some of these functions *do* return the type `{name}` I suppose, but some of them don't. The type signature is wrong for the ones supporting `cls` (afaict).

wheel/generate_type_stubs.py
```diff
@@ -222,7 +222,7 @@ def parse_rust_source(filename: str, upper_case: bool) -> List[Tuple[str, List[s
         "def uncurry(self) -> Tuple[ChiaProgram, ChiaProgram]: ...",
     ],
     "SpendBundle": [
-        "@staticmethod\n    def aggregate(sbs: List[SpendBundle]) -> SpendBundle: ...",
+        "@classmethod\n    def aggregate(sbs: List[SpendBundle]) -> SpendBundle: ...",
```
arvidn, line 225: the `aggregate()` function also takes `cls`, right? And it doesn't necessarily return a `SpendBundle`

arvidn (commented):
I think there are a few simplification opportunities in the error propagation, and I think the type stubs can be improved

crates/chia_py_streamable_macro/src/lib.rs
matt-o-how: Done

crates/chia_py_streamable_macro/src/lib.rs
matt-o-how: Done

crates/chia_py_streamable_macro/src/lib.rs
matt-o-how: Done

wheel/generate_type_stubs.py
matt-o-how: Done

crates/chia-protocol/src/coin_spend.rs
matt-o-how: Added a check if `cls` is different from constructed Rust object and now this is only called if that's true, meaning it is skipped every time in practice.

wheel/generate_type_stubs.py
matt-o-how: Updated return val to `typing.Self`

crates/chia-protocol/src/spend_bundle.rs
```diff
@@ -94,10 +98,25 @@ impl SpendBundle {
...
+        let aggregated = Self::aggregate(&spend_bundles);
+        Python::with_gil(|py| {
+            // Convert result into potential child class
+            let instance = cls.call(
+                (aggregated.coin_spends, aggregated.aggregated_signature),
+                None,
+            )?;
+
+            Ok(instance.into_py(py))
+        })
```
arvidn, line 113:
```suggestion
    fn py_aggregate(cls: &Bound<'_, PyType>, spend_bundles: Vec<Self>, py: Python<`_>) -> PyResult<PyObject> {
        let aggregated = Self::aggregate(&spend_bundles);
        // Convert result into potential child class
        let instance = cls.call(
            (aggregated.coin_spends, aggregated.aggregated_signature),
            None,
        )?;
        Ok(instance.into_py(py))
```

crates/chia_py_streamable_macro/src/lib.rs
```diff
@@ -174,45 +189,91 @@ pub fn py_streamable_macro(input: proc_macro::TokenStream) -> proc_macro::TokenS
...
+                // Check if python class is different from rust class (in case of child classes)
+                // if so call the python class's conversion code
+
+                let cls_type = cls.borrow().as_ref();
+                if <Self as PyTypeInfo>::is_exact_type_of_bound(cls_type) {
+                    Ok(rust_obj.into_py(py))
+                } else {
+                    let instance = cls.call_method1("from_parent", (rust_obj.into_py(py),))?;
+                    Ok(instance.into_py(py))
+                }
```
arvidn, line 244: is there any chance this snippet could be factored out into a function? or something else to address the apparent copy-pasted code

arvidn (commented):
looking good. I thought you used to have a python test as well, to ensure it works

crates/chia-protocol/src/spend_bundle.rs
matt-o-how: Done

crates/chia-bls/src/signature.rs
```diff
@@ -486,6 +490,15 @@ impl Signature {
         Self::default()
     }
 
+    #[classmethod]
+    #[pyo3(name = "from_parent")]
+    pub fn from_parent(_cls: &Bound<'_, PyType>, instance: Self) -> PyResult<PyObject> {
```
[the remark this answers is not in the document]
matt-o-how: a copy here allows the required move of instance's values into .call()

I've changed the other `Self`s to be `&Self` however

crates/chia-protocol/src/spend_bundle.rs
matt-o-how: Done

crates/chia-protocol/src/coin_spend.rs
```diff
@@ -2,10 +2,29 @@ use chia_streamable_macro::streamable;
...
+        Python::with_gil(|py| {
+            // Convert result into potential child class
+            let instance = cls.call1((cs.coin, cs.puzzle_reveal, cs.solution))?;
+
+            Ok(instance.into_py(py))
+        })
+    }
```
arvidn, line 29:
```suggestion
    pub fn from_parent(
        cls: &Bound<'_, PyType>,
        py: Python<'_>,
        cs: Self) -> PyResult<PyObject> 
    {
        // Convert result into potential child class
        let instance = cls.call1((cs.coin, cs.puzzle_reveal, cs.solution))?;

        Ok(instance.into_py(py))
    }
```

arvidn (commented):
look good, you just missed one place that sill uses `with_gil()`

crates/chia-bls/src/signature.rs
```diff
@@ -494,10 +496,10 @@ impl Signature {
     #[pyo3(name = "from_parent")]
     pub fn from_parent(
         _cls: &Bound<'_, PyType>,
-        py: Python<'_>,
-        instance: Self,
+        _py: Python<'_>,
```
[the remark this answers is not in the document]
matt-o-how: Done

arvidn (approved):

PR #679, opened 2024-08-26
include byte cost in return value from `get_conditions_from_spendbundle()`

arvidn:
the byte cost was not included in the returned cost field, but it's supposed to be


Rigidity:
Ah nice, I was wondering whether this was intentional or not

matt-o-how (approved):

PR #680, opened 2024-08-27
CHIA-1226 Include the cost of the whole spend in get_conditions_from_spendbundle's return value

AmineKhaldi:
The generator doesn't just include the puzzle reveal and the solution, it also includes the coin that's being spent, specifically in the form of `(parent_id, puzzle_hash, amount)`. Each spend will necessarily have to pay that size-cost.

### DIFF
--- a/crates/chia-consensus/src/spendbundle_conditions.rs
+++ b/crates/chia-consensus/src/spendbundle_conditions.rs
@@ -4,6 +4,7 @@ use crate::gen::conditions::{
 };
 use crate::gen::flags::MEMPOOL_MODE;
 use crate::gen::run_block_generator::subtract_cost;
+use crate::gen::solution_generator::solution_generator;
 use crate::gen::validation_error::ValidationErr;
 use crate::spendbundle_validation::get_flags_for_height_and_constants;
 use chia_protocol::SpendBundle;
@@ -30,12 +31,21 @@ pub fn get_conditions_from_spendbundle(
     let mut ret = SpendBundleConditions::default();
     let mut state = ParseState::default();
 
+    let spends_info = spend_bundle.coin_spends.iter().map(|coin_spend| {
+        (
+            coin_spend.coin,
+            &coin_spend.puzzle_reveal,
+            &coin_spend.solution,
+        )
+    });
+    // We don't pay the size cost (nor execution cost) of being wrapped by a
+    // quote (in solution_generator).
+    const QUOTE_BYTES: usize = 2;
+    let generator_length_without_quote = solution_generator(spends_info)?.len() - QUOTE_BYTES;
+    let byte_cost = generator_length_without_quote as u64 * constants.cost_per_byte;
+    subtract_cost(a, &mut cost_left, byte_cost)?;
+
     for coin_spend in &spend_bundle.coin_spends {
-        let byte_cost = (coin_spend.puzzle_reveal.len() + coin_spend.solution.len()) as u64
-            * constants.cost_per_byte;
-
-        subtract_cost(a, &mut cost_left, byte_cost)?;
-
         // process the spend
         let puz = node_from_bytes(a, coin_spend.puzzle_reveal.as_slice())?;
         let sol = node_from_bytes(a, coin_spend.solution.as_slice())?;
@@ -74,6 +84,7 @@ mod tests {
     use super::*;
     use crate::allocator::make_allocator;
     use crate::gen::conditions::{ELIGIBLE_FOR_DEDUP, ELIGIBLE_FOR_FF};
+    use crate::gen::run_block_generator::run_block_generator2;
     use chia_bls::Signature;
     use chia_protocol::CoinSpend;
     use chia_traits::Streamable;
@@ -82,8 +93,8 @@ mod tests {
     use std::fs::read;
 
     #[rstest]
-    #[case("3000253", 8, 2, 46_860_870)]
-    #[case("1000101", 34, 15, 231_687_677)]
+    #[case("3000253", 8, 2, 51_216_870)]
+    #[case("1000101", 34, 15, 250_083_677)]
     fn test_get_conditions_from_spendbundle(
         #[case] filename: &str,
         #[case] spends: usize,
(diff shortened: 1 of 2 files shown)


Rigidity (changes requested):
The benefit of subtracting the cost incrementally as you go is to exit early if the spend exceeds the limit, before processing the actual CLVM and conditions, or other spends. I think doing it afterward could open up a DoS attack vector

Rigidity (approved):

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -30,12 +31,22 @@ pub fn get_conditions_from_spendbundle(
...
+        .coin_spends
+        .iter()
+        .map(|coin_spend| {
+            (
+                coin_spend.coin,
+                &coin_spend.puzzle_reveal,
+                &coin_spend.solution,
+            )
+        })
+        .collect();
```
arvidn, line 44: you don't need to `collect()` here. `solution_generator()` already takes an iterator.
However, since this is time-critical, I think it would be best to compute this length without also building the generator. Would that be unreasonably complex?

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -108,6 +120,32 @@ mod tests {
...
+            &mut a,
+            program.as_slice(),
+            blocks,
+            11_000_000_000,
+            MEMPOOL_MODE,
+            &TEST_CONSTANTS,
+        )
+        .expect("run_block_generator2 failed");
+        const QUOTE_COST: u64 = 20;
+        // run_block_generator2 wraps the program in a quote
```
arvidn, line 147: this comment is not correct. `run_block_generator2` just *runs* the generator. it's `solution_generator()` that does the wrapping.  but executing the quote costs 20. However, you should also deduct the cost of the size of the wrapping, which I believe is 2 bytes

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -108,6 +120,32 @@ mod tests {
...
+        // Generate a block with the same spend bundle and compare its cost
+        let program_spends: Vec<_> = bundle
+            .coin_spends
+            .iter()
+            .map(|coin_spend| {
+                (
+                    coin_spend.coin,
+                    &coin_spend.puzzle_reveal,
+                    &coin_spend.solution,
+                )
```
arvidn, line 132: rather than returning this tuple, you could just return the byte-size of serializing it. and then `sum()` that.

arvidn (changes requested):

crates/chia-consensus/src/spendbundle_conditions.rs
AmineKhaldi: > you don't need to collect() here. solution_generator() already takes an iterator.

Done.

> Would that be unreasonably complex?

I believe so.

crates/chia-consensus/src/spendbundle_conditions.rs
arvidn: right, and the only thing you do with the vector it returns is to call `.len()`. You could just compute the length instead

arvidn: this comment was not meant for the test, but the production code, here: https://github.com/Chia-Network/chia_rs/pull/680/files#diff-9ad3aa7496f5fdc7db0a33393d2c8bd18727a7924d3d98189433b273e3d15439R34-R42

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -30,12 +31,18 @@ pub fn get_conditions_from_spendbundle(
...
-        subtract_cost(a, &mut cost_left, byte_cost)?;
+    let spends_info = spend_bundle.coin_spends.iter().map(|coin_spend| {
+        (
+            coin_spend.coin,
+            &coin_spend.puzzle_reveal,
+            &coin_spend.solution,
+        )
+    });
+    let spend_generator_length = solution_generator(spends_info)?.len();
+    let byte_cost = spend_generator_length as u64 * constants.cost_per_byte;
```
arvidn, line 42: I think the long term goal should be to not actually generate the block-generator, but just compute the byte-cost of it. For now, I think it's fine, to get everything to work first.

You should deduct the 2 bytes for the quote as well

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -108,6 +116,36 @@ mod tests {
...
+            MEMPOOL_MODE,
+            &TEST_CONSTANTS,
+        )
+        .expect("run_block_generator2 failed");
+        // The QUOTE_COST difference here is because get_conditions_from_spendbundle
+        // uses solution_generator which wraps the program in a quote, and
+        // while that quote's byte cost is included in the cost calculation, its
+        // execution cost is not, because get_conditions_from_spendbundle doesn't
+        // run that generator.
+        const QUOTE_COST: u64 = 20;
```
arvidn, line 147: this difference *should* be 24020 because of the bytes-cost of the quote as well

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -108,6 +116,36 @@ mod tests {
...
+            11_000_000_000,
+            MEMPOOL_MODE,
+            &TEST_CONSTANTS,
+        )
+        .expect("run_block_generator2 failed");
+        // The QUOTE_COST difference here is because get_conditions_from_spendbundle
+        // uses solution_generator which wraps the program in a quote, and
+        // while that quote's byte cost is included in the cost calculation, its
+        // execution cost is not, because get_conditions_from_spendbundle doesn't
+        // run that generator.
```
arvidn, line 146: The cost difference comes from the overhead of creating and executing a block generator (i.e. `solution_generator()` + `run_block_generator2()`) versus just the cost of the spend bundles. The spend bundles don't pay the size-cost nor execution cost of being wrapped by a quote.

crates/chia-consensus/src/spendbundle_conditions.rs
AmineKhaldi: Done.

crates/chia-consensus/src/spendbundle_conditions.rs
AmineKhaldi: Done.

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -108,6 +119,34 @@ mod tests {
...
+            program.as_slice(),
+            blocks,
+            11_000_000_000,
+            MEMPOOL_MODE,
+            &TEST_CONSTANTS,
+        )
+        .expect("run_block_generator2 failed");
+        // The cost difference here is because get_conditions_from_spendbundle
+        // uses solution_generator which wraps the program in a quote, and we
+        // don't pay the size cost nor execution cost for that there.
```
arvidn, line 143: this isn't quite right. the quote difference here isn't *caused* by an implementation detail. It's deliberate and, ideally, the comment would explain why it's deliberate. Specifically, `get_conditions_from_spendbundle()` does not include the overhead to make a block.

arvidn (approved):

crates/chia-consensus/src/spendbundle_conditions.rs
AmineKhaldi: Done.

arvidn (approved):

Rigidity (approved):

PR #681, opened 2024-08-27
Update .gitignore using gitignore.io

altendky:


.gitignore
```diff
@@ -1,14 +1,419 @@
...
+pyrightconfig.json
+
+### Rust ###
+# Generated by Cargo
+# will have compiled files and executables
+debug/
+
+# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
+# More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
+Cargo.lock
```
arvidn, line 391: I believe the python wheel would be considered an "executable" in this context.

.gitignore
```diff
@@ -1,14 +1,419 @@
...
+### Python ###
+# Byte-compiled / optimized / DLL files
 __pycache__/
-venv/
+*.py[cod]
+*$py.class
+
+# C extensions
 *.so
-*.pyd
```
arvidn, line 6: this is the file extension for native python modules on windows, right? Do we not need this?

.gitignore
altendky: I see two topics here.  They may be in conflict.

1) Should we include `Cargo.lock` file(s) in the repo?
2) Should we edit the standard output of the gitignore template?

For 1), there is presently a single `Cargo.lock` in the repo and it is revisioned.  For that file, this entry in the configuration is irrelevant unless we remove the file from git.  In that case git would not remind us to add it back.  This doesn't seem like a critical backstop.

For 2), I guess we can edit it.  When someone does an update to include either updated changes from the templates for the groups of ignores selected here, or to add another group such as for another editor, then they will overwrite the manual modification.  Odds are the rust template won't change drastically so the manual modification will show in the diff and be noticed.  If we were to make the ignore files managed like other general repo files then this exception would probably get more annoying.  Especially since the comment suggests it is a conditional thing that would vary and require both additional configuration per rust repo and code to manage the modification.

Given the near zero upside of modifying this and the everpresent issues around modifying generated content, I'm obviously not in favor of it.  But, if someone clicks commit below, then so it goes.

```suggestion
# Cargo.lock
```

.gitignore
altendky: It is covered by `*.py[cod]`.

PR #689, opened 2024-08-28
Delete wheel/py.typed (leaving wheel/python/chia_rs/py.typed)

altendky:
I think this outer one is unneeded?

https://github.com/Chia-Network/chia_rs/actions/runs/10597276846?pr=689
image


arvidn:
that sounds right. We don't have a test to ensure mypy detects type errors, so would we notice if it breaks type checking?

altendky:
Probably not here presently.  The stubtest PR would be a check I think, once cleaned up and approved etc.  Do you want a manual check now?  To wait for a CI check?

altendky:
The stubtest check https://github.com/Chia-Network/chia_rs/pull/685 is merged so this should be checking the in-wheel hinting.

Rigidity (approved):

PR #692, opened 2024-08-28
Check the whole workspace in one place

Rigidity:


.github/workflows/build-wheels.yml
```diff
@@ -216,23 +216,6 @@ jobs:
...
-
-      - name: Checkout code
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-
-      - name: cargo build
-        run: |
-          cd crates/chia-tools
-          cargo build
```
arvidn, line 234: I don't think `cargo test` will build these targets.

Rigidity: Added `cargo build --workspace --all-features --release`, which will cover it

arvidn (approved):

PR #699, opened 2024-09-03
Remove references to `NPCResult`

matt-o-how:
A straightforward variable rename


arvidn:
It's not obvious to me that this name is an improvement.

arvidn (approved):

PR #703, opened 2024-09-06
CHIA-1344 Release the GIL when calling validate_clvm_and_signature in the python binding

AmineKhaldi:


wheel/src/api.rs
```diff
@@ -422,16 +422,19 @@ fn fast_forward_singleton<'p>(
 #[pyo3(name = "validate_clvm_and_signature")]
 #[allow(clippy::type_complexity)]
 pub fn py_validate_clvm_and_signature(
+    py: Python<'_>,
     new_spend: &SpendBundle,
```
arvidn, line 426: can you double check that it's OK to pass a reference to the `SpendBundle` into `allow_threads()` without introducing races? If the reference pyo3 passes in there points directly to the object held by python, and it's mutable, it can change behind this thread's back. (and would introduce a race)

arvidn (approved):

wheel/src/api.rs
AmineKhaldi: > can you double check that it's OK to pass a reference to the `SpendBundle` into `allow_threads()` without introducing races? If the reference pyo3 passes in there points directly to the object held by python, and it's mutable, it can change behind this thread's back. (and would introduce a race)

From the python side, we get `SpendBundle` from Rust, and if you try to alter such an object you'd get "object is not writable" error, so we shouldn't be able to mutate it behind a thread's back as a result.

arvidn (approved):

PR #710, opened 2024-09-12
Change BLSCache to take and return deserialized GTElements

matt-o-how:
Changes affect `.update()` and `.items()` in the pybindings


arvidn (approved):
looks good. We're not using `items()` anymore though

arvidn:
did you try this in `chia-blockchain` btw? Is it possible to pass `GTElement` over a process boundary? I suppose we can always fix it by explicitly converting to `bytes` for now

PR #714, opened 2024-09-17
Optimize `get_conditions_from_spendbundle()` by calculating the generator's size without creating the generator

matt-o-how:
Adds a new function called `calculate_generator_length()` which deterministically calculates the generators byte length given the spends. This function can act as a drop in replacement when we only care about the size of the generator, as the input is the same.

### DIFF
--- a/crates/chia-consensus/fuzz/fuzz_targets/solution-generator.rs
+++ b/crates/chia-consensus/fuzz/fuzz_targets/solution-generator.rs
@@ -22,7 +22,7 @@ fuzz_target!(|data: &[u8]| {
             spend.puzzle_reveal.to_vec(),
             spend.solution.to_vec(),
         ));
-        // Check if puzzle or solution are atoms which can represented in a smaller form
+        // Check for atoms which can represented in a smaller form
         let node = node_from_bytes_backrefs(&mut a, spend.puzzle_reveal.as_ref()).expect("node");
         let puz = node_to_bytes(&a, node).expect("bytes");
         discrepancy += spend.puzzle_reveal.as_ref().len() as i64 - puz.len() as i64;
--- a/crates/chia-consensus/fuzz/fuzz_targets/solution-generator.rs
+++ b/crates/chia-consensus/fuzz/fuzz_targets/solution-generator.rs
@@ -3,7 +3,7 @@ use chia_consensus::gen::solution_generator::{calculate_generator_length, soluti
 use chia_protocol::{Coin, CoinSpend};
 use chia_traits::Streamable;
 use clvmr::{
-    serde::{node_from_bytes, node_to_bytes},
+    serde::{node_from_bytes_backrefs, node_to_bytes},
     Allocator,
 };
 use libfuzzer_sys::fuzz_target;
@@ -13,7 +13,7 @@ fuzz_target!(|data: &[u8]| {
     let mut spends = Vec::<CoinSpend>::new();
     let mut generator_input = Vec::<(Coin, Vec<u8>, Vec<u8>)>::new();
     let mut data = Cursor::new(data);
-    let mut discrepancy: usize = 0;
+    let mut discrepancy: i64 = 0;
     let mut a = Allocator::new();
     while let Ok(spend) = CoinSpend::parse::<false>(&mut data) {
         spends.push(spend.clone());
@@ -23,30 +23,23 @@ fuzz_target!(|data: &[u8]| {
             spend.solution.to_vec(),
         ));
         // Check if puzzle or solution are atoms which can represented in a smaller form
-        let node = node_from_bytes(&mut a, spend.puzzle_reveal.as_ref()).expect("node");
-        if node.is_atom() {
-            let puz = node_to_bytes(&a, node).expect("bytes");
-            discrepancy += spend.puzzle_reveal.as_ref().len() - puz.len();
-        }
-        let node = node_from_bytes(&mut a, spend.solution.as_ref()).expect("node");
-        if node.is_atom() {
-            let sol = node_to_bytes(&a, node).expect("bytes");
-            discrepancy += spend.solution.as_ref().len() - sol.len();
-        }
+        let node = node_from_bytes_backrefs(&mut a, spend.puzzle_reveal.as_ref()).expect("node");
+        let puz = node_to_bytes(&a, node).expect("bytes");
+        discrepancy += spend.puzzle_reveal.as_ref().len() as i64 - puz.len() as i64;
+        let node = node_from_bytes_backrefs(&mut a, spend.solution.as_ref()).expect("node");
+        let sol = node_to_bytes(&a, node).expect("bytes");
+        discrepancy += spend.solution.as_ref().len() as i64 - sol.len() as i64;
     }
     if spends.is_empty() {
         return;
     }
-    let Ok(result) = solution_generator(generator_input) else {
-        return;
-    };
-
-    if result.len() != calculate_generator_length(spends.clone()) - discrepancy {
+    let result = solution_generator(generator_input).expect("solution_generator");
+    if result.len() as i64 != calculate_generator_length(spends.clone()) as i64 - discrepancy {
         panic!("Debug spends: {:?}", spends);
     }
 
     assert_eq!(
-        result.len(),
-        calculate_generator_length(spends) - discrepancy
+        result.len() as i64,
+        calculate_generator_length(spends) as i64 - discrepancy
     );
 });
--- a/crates/chia-consensus/fuzz/fuzz_targets/solution-generator.rs
+++ b/crates/chia-consensus/fuzz/fuzz_targets/solution-generator.rs
@@ -1,0 +1,23 @@
+#![no_main]
+use chia_protocol::{Coin, CoinSpend};
+use chia_traits::Streamable;
+use libfuzzer_sys::fuzz_target;
+use std::io::Cursor;
+use chia_consensus::gen::solution_generator::{calculate_generator_length, solution_generator};
+
+fuzz_target!(|data: &[u8]| {
+    let mut spends = Vec::<CoinSpend>::new();
+    let mut generator_input = Vec::<(Coin, Vec<u8>, Vec<u8>)>::new();
+    for _i in 1..1000 {
+        let Ok(spend) = CoinSpend::parse::<false>(&mut Cursor::new(data)) else {
+            return;
+        };
+        spends.push(spend.clone());
+        generator_input.push((spend.coin, spend.puzzle_reveal.to_vec(), spend.solution.to_vec()));
+    }
+
+    let result = solution_generator(generator_input).expect("solution_generator");
+
+    assert_eq!(result.len(), calculate_generator_length(spends));
+
+});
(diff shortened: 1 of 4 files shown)


arvidn (commented):
it might make sense to add a fuzzer for this, that ensures the results always match

crates/chia-consensus/src/gen/solution_generator.rs
```diff
@@ -39,6 +39,37 @@ where
...
+        let solution = s.2.as_ref();
+
+        let amount_bytes = s.0.amount.to_be_bytes();
+        // chialisp represents integers as small as possible so remove leading 0s
+        let leading_zeroes = amount_bytes.iter().take_while(|&&b| b == 0).count();
+        let mut amount_size = amount_bytes[leading_zeroes..].len();
+        if amount_bytes[leading_zeroes] >= 0x80 {
+            // add 0x00 for two's compliment as amount is always positive
+            amount_size += 1;
+        }
```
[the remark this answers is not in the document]
arvidn: or more like this, but for `u64`: https://github.com/Chia-Network/clvm_rs/blob/main/src/allocator.rs#L202

crates/chia-consensus/src/gen/solution_generator.rs
```diff
@@ -39,6 +39,37 @@ where
     Ok(quote)
 }
 
+// calculate the size in bytes of a generator with no backref optimisations
+pub fn calculate_generator_length<BufRef, I>(spends: I) -> usize
+where
+    BufRef: AsRef<[u8]>,
+    I: IntoIterator<Item = (Coin, BufRef, BufRef)>,
```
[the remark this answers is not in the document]
matt-o-how: I was doing that so it was a drop in for build_generator, but it's swapped to CoinSpend now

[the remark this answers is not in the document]
matt-o-how: I was doing that so it was a drop in for build_generator, but it's swapped to CoinSpend now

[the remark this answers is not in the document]
Rigidity: CoinSpend requires the items to be owned copies of the `Program` whereas with `BufRef` you can pass in slices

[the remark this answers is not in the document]
Rigidity: But I guess with `&[CoinSpend]` for example like you have it, it's more ergonomic and you still don't need to clone anything

crates/chia-consensus/src/gen/solution_generator.rs
```diff
@@ -39,36 +40,41 @@ where
     Ok(quote)
 }
 
+// do not use generally
+fn len_for_generator(val: u64) -> usize {
```
arvidn, line 44: I don't quite understand what this function is doing exactly. The return values aren't intuitive to me.
It would be easy to think that this function ought to return the same as:

```
let mut a = Allocator::new();
let atom = a.new_number(val);
a.atom_len(atom)
```

But you're primarily concerned about the *serialized* length, right? Which is slightly different, a value < `0x80` is serialized as a single byte, but as soon as you need two bytes, we also have a length prefix for the atom. So I think it's more accurate to say that this function should return the same as:

```
let mut a = Allocator::new();
let atom = a.new_number(val);
let bytes = node_to_bytes(&a, atom);
bytes.len()
```

arvidn (commented):
This looks good. Just to be extra sure this is producing the same result as `solution_generator()` I think it would be great to have a fuzzer that run both on arbitrary coin spends.
You can use https://github.com/Chia-Network/chia_rs/blob/main/crates/chia-consensus/fuzz/fuzz_targets/fast-forward.rs#L19-L22 as inspiration (really just how to obtain arbitrart `CoinSpend` objects).

You could also extend this tool to generate the corpus for the new fuzzer:

https://github.com/Chia-Network/chia_rs/blob/main/crates/chia-tools/src/bin/gen-corpus.rs#L132-L138

crates/chia-consensus/src/gen/solution_generator.rs
matt-o-how: This is generally accurate, however I observed that the chialisp itself gets optimised when the size of the amount is less than 128, saving 1 byte. So while `len(bytes(Program.to(97)))` is actually 1, with the optimisation the net size change is 0.

arvidn (commented):
I will make a PR against this to add support in `gen-corpus`, and I'll try out the fuzzer

crates/chia-consensus/src/gen/solution_generator.rs
arvidn: I think this is a misunderstanding. the length-prefix for atoms is similar to UTF-8, so small values (< 0x80) are encoded as a single byte (the leading 0 bit indicates a small value). I suspect your calculation compensates for this 0-bytes by adding it back later

arvidn: The full details of the serialization can be found here: https://chialisp.com/clvm/#large-atoms

In the case of a `u64`, we know that it will never be more than 9 bytes (including a leading zero if the value is large enough). That means the *length prefix* never has to be more than 1 byte.
If the value is < `0x80`, it's encoded as 1 byte. Otherwise it's encoded as 1-byte length-prefix (`0b10xxxxxx`) + the actual bytes themselves. The bytes being encoded as variable length, signed, big endian. I think you end up with something like this:

| value | serialized length |
| --- | --- |
| 0 | 1 byte |
| 0x7f | 1 byte |
| 0x80 | 1 byte prefix + 2 bytes value |
| 0x7fff | 1 byte prefix + 2 bytes value |
| 0x8000 | 1 byte prefix + 3 bytes value |
| 0x800000 | 1 byte prefix + 4 bytes value |

and so on. I think it would make it simpler to have this function be well defined as the serialized length of the `u64` value. You could then also have a test that covers the most relevant values (i.e. the boundary values where it the length changes)

crates/chia-consensus/src/gen/solution_generator.rs
```diff
@@ -39,6 +40,46 @@ where
...
+where
+    I: AsRef<[CoinSpend]>,
+{
+    let mut size: usize = 5; // (q . (())) => ff01ff8080 => 5 bytes
+
+    for s in spends.as_ref() {
+        let puzzle = s.puzzle_reveal.as_ref();
+        let solution = s.solution.as_ref();
+        // parent-id puzzle-reveal amount solution + bytes for list extension and atom prep bytes
+        size += 32 + puzzle.len() + len_for_generator(s.coin.amount) + solution.len() + 8;
```
arvidn, line 77: the 32 is the parent coin ID, right? the serialized length is 33 bytes (because of the length prefix). I suspect that the length prefix (as well as the missing byte from `len_for_generator()` is lumped into the `8` at the end. I think it would be easier to understand if they were more explicitly acknowledged in the code.

arvidn (approved):

arvidn (approved):

crates/chia-consensus/fuzz/fuzz_targets/solution-generator.rs
```diff
@@ -23,12 +23,16 @@ fuzz_target!(|data: &[u8]| {
             spend.solution.to_vec(),
         ));
         // Check if puzzle or solution are atoms which can represented in a smaller form
-        let node = node_from_bytes_backrefs(&mut a, spend.puzzle_reveal.as_ref()).expect("atom");
+        let Ok(node) = node_from_bytes(&mut a, spend.puzzle_reveal.as_ref()) else {
```
arvidn, line 26: I think it's fine to use `.expect()` here still

arvidn: the `Program` class is supposed to guarantee that the buffer is fine

crates/chia-consensus/src/gen/solution_generator.rs
```diff
@@ -89,7 +127,7 @@ where
     I: IntoIterator<Item = (Coin, BufRef, BufRef)>,
 {
     let mut a = Allocator::new();
-    let generator = build_generator(&mut a, spends)?;
+    let generator = build_generator_no_backrefs(&mut a, spends)?;
```
arvidn, line 130: oh, I meant I think you can change `build_generator()` to not use backrefs

crates/chia-consensus/src/gen/solution_generator.rs
```diff
@@ -39,14 +42,92 @@ where
     Ok(quote)
 }
 
+// the tuple has the Coin, puzzle-reveal and solution
+fn build_generator_no_backrefs<BufRef, I>(a: &mut Allocator, spends: I) -> io::Result<NodePtr>
```
arvidn, line 46: the existing `build_generator()` function (above) is private in this file, and only used in one place. I don't think there's any need to preserve it. i.e. I think it would be simpler to just change it to not parse `solution` and `puzzle` supporting backrefs

crates/chia-consensus/fuzz/fuzz_targets/solution-generator.rs
```diff
@@ -0,0 +1,42 @@
...
+    let mut discrepancy: i64 = 0;
+    let mut a = Allocator::new();
+    while let Ok(spend) = CoinSpend::parse::<false>(&mut data) {
+        spends.push(spend.clone());
+        generator_input.push((
+            spend.coin,
+            spend.puzzle_reveal.to_vec(),
+            spend.solution.to_vec(),
+        ));
+        // Check if puzzle or solution are atoms which can represented in a smaller form
```
arvidn, line 25: this comment is technically outdated now. It doesn't matter whether the atom is at the top level or buried deep down a tree. it still introduces the same discrepancy

arvidn (dismissed):

crates/chia-consensus/src/gen/solution_generator.rs
```diff
@@ -39,6 +40,48 @@ where
...
+    } else if val < 0x8000_0000_0000 {
+        7
+    } else if val < 0x0080_0000_0000_0000 {
+        8
+    } else if val < 0x8000_0000_0000_0000 {
+        9
+    } else {
+        10
+    }
+}
```
AmineKhaldi, line 63: Perhaps a comment explaining the behavior and a test that further documents it?

crates/chia-consensus/src/gen/solution_generator.rs
```diff
@@ -39,6 +40,48 @@ where
...
+        let solution = s.solution.as_ref();
+        // parent-id puzzle-reveal amount solution
+        // parent-id is always 32 bytes + prepend 1 byte = 33
+        // + 6 bytes for list extension
+        // coin amount is already prepended correctly in clvm_bytes_len()
+        size += 39 + puzzle.len() + clvm_bytes_len(s.coin.amount) + solution.len();
+    }
+
+    size
+}
```
AmineKhaldi, line 83: Perhaps a comment similar to https://github.com/Chia-Network/chia_rs/blob/38579c836ea560ae74b4821ce397c0d88630bdeb/crates/chia-consensus/src/gen/solution_generator.rs#L12-L14 would make all the other comments below it clearer?

arvidn (approved):

arvidn:
I'll address Amine's comment in a follow up PR

PR #717, opened 2024-09-20
Bump clvm to 0.9.0

arvidn:
`ENABLE_BLS_OPS_OUTSIDE_GUARD` and `ENABLE_FIXED_DIV` have been removed in the new version of `clvm_rs`. Those features are now enabled unconditionally.

Some of the python tests imported `chia-blockchain`, which in turn depends on `chia_rs`. With this change, `chia_rs` no longer implements the exact same API, and so will fail those tests. Since we've had the rust-python fidelity tests running for a while now, we probably don't need them anymore. If we do, we should probably copy the old `chia-blockchain` code into the tests.

The python stub file was still mentioning `ENABLE_BLS_OPS`, `ENABLE_SECP_OPS` even though they were removed some time ago.


Rigidity (approved):

wjblanke (commented):
aok

PR #735, opened 2024-09-27
CHIA-1527: Add chia-sha2 crate to prevent cyclic dependency with clvmr

Rigidity:


arvidn (approved):
I take it most of this was copied verbatim from `clvm_rs`, right?

Rigidity:
> I take it most of this was copied verbatim from `clvm_rs`, right?

Yeah the sha2.rs file just copied over

PR #744, opened 2024-10-03
Change BLSCache to use the augmented message for its lookup

matt-o-how:
This reduces the allocations required inside the bls_cache and inside `spendbundle_validation.rs`.

### DIFF
--- a/crates/chia-bls/src/bls_cache.rs
+++ b/crates/chia-bls/src/bls_cache.rs
@@ -53,7 +53,9 @@ impl BlsCache {
             // Hash pubkey + message
             let mut hasher = Sha256::new();
             hasher.update(pk.borrow().to_bytes());
-            hasher.update(msg.as_ref());
+            let mut aug_msg = pk.borrow().to_bytes().to_vec();
+            aug_msg.extend_from_slice(msg.as_ref());
+            hasher.update(&aug_msg);
             let hash: [u8; 32] = hasher.finalize();
 
             // If the pairing is in the cache, we don't need to recalculate it.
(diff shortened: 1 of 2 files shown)


crates/chia-bls/src/bls_cache.rs
```diff
@@ -53,7 +53,9 @@ impl BlsCache {
             // Hash pubkey + message
             let mut hasher = Sha256::new();
             hasher.update(pk.borrow().to_bytes());
-            hasher.update(msg.as_ref());
+            let mut aug_msg = pk.borrow().to_bytes().to_vec();
+            aug_msg.extend_from_slice(msg.as_ref());
+            hasher.update(&aug_msg);
```
arvidn, line 58: this is prepending the public key twice. I don't think we need to do that. actually, maybe I was a bit confused when I wrote up that ticket

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -54,14 +53,12 @@ pub fn validate_clvm_and_signature(
...
-                final_msg.extend_from_slice(msg.as_slice());
                 aug_msg.extend_from_slice(&pk.to_bytes());
-                make_aggsig_final_message(condition, &mut final_msg, spend, constants);
-                aug_msg.extend(&final_msg);
+                aug_msg.extend_from_slice(msg.as_slice());
+                make_aggsig_final_message(condition, &mut aug_msg, spend, constants);
                 let aug_hash = hash_to_g2(&aug_msg);
                 let pairing = aug_hash.pair(pk);
-                pairs.push((hash_pk_and_msg(&pk.to_bytes(), &final_msg), pairing));
+                pairs.push((hash_pk_and_msg(&pk.to_bytes(), &aug_msg), pairing));
```
arvidn, line 61: perhaps it's possible to *just* simplify this part, by just hashing `aug_msg` here, rather than using `hash_pk_and_msg()`, without any other changes

arvidn (approved):

PR #745, opened 2024-10-04
[CHIA-1562] validate block signature as part of `run_block_generator()`

arvidn:
There is an opt-out for tests and maybe we'll need it in production too.

There are 3 main intentions:

1. Allow releasing the GIL for longer periods at a time, by combining executing CLVM, parsing conditions and validating the signature
2. Make the logic around validating signatures simpler, by always validating them when running a block (there's an opt-out for tests right now, and we may need it while transitioning).
3. (in the future) avoid copying all the public keys and messages back into python.

~~In this first version, the `BLSCache` is not thread safe. This would likely be a problem since we use `run_block_generator()` from threads in python. It should probably contain a `Mutex` to serialize access to it.~~

The new flag `DONT_VALIDATE_SIGNATURE` can be passed in to skip the validation.
The new field `validated_signature` on `SpendBundleConditions` can be used to ensure it was in fact validated.

The main changes are in `conditions.rs`. The `pkm_pairs` are collected, as they are parsed, into a single list in the `ParseState` object. Here: https://github.com/Chia-Network/chia_rs/compare/run_block_validate_signature?expand=1#diff-c93029cf40a6e979ca6c9b02528cfc71cad2e7099818a2deeb92cb9cf5cf344bR788

At the end, these key-message pairs are validated against the signature in this new function `validate_signature()`: https://github.com/Chia-Network/chia_rs/compare/run_block_validate_signature?expand=1#diff-c93029cf40a6e979ca6c9b02528cfc71cad2e7099818a2deeb92cb9cf5cf344bR1502


arvidn:
note that the meat of this patch is in `conditions.rs`, which github collapses by default in the diff view. Almost everything else are changes to tests to compensate.

matt-o-how (dismissed):

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -4461,6 +4686,90 @@ fn test_eligible_for_ff_invalid_agg_sig_me(
...
+        add_signature(&mut signature, &mut puzzle, *opcode);
+    }
+    puzzle.push_str("))))");
+    if !expect_pass {
+        signature.aggregate(&sign(
+            &SecretKey::from_bytes(SECRET_KEY).unwrap(),
+            b"foobar",
+        ));
+    }
+    match cond_test_sig(puzzle.as_str(), &signature, cache, 0) {
```
Rigidity, line 4763:
```rs
assert_eq!(expect_pass, cond_test_sig(...).is_ok());
```

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -775,6 +779,13 @@ pub struct ParseState {
     // ASSERT_MY_BIRTH_HEIGHT
     // each item is the index into the SpendBundleConditions::spends vector
     assert_not_ephemeral: HashSet<usize>,
+
+    // All public keys and messages emitted by the generator. We'll validate
+    // these against the aggregate signature at the end, unless the
+    // DONT_VALIDATE_SIGNATURE flag is set
+    // TODO: We would probably save heap allocations by turning this into a
+    // blst_pairing object.
+    pub pkm_pairs: Vec<(PublicKey, Vec<u8>)>,
```
Rigidity, line 788: Maybe using the `Bytes` type here would be ideal for consistency and to make `Debug` printing cleaner?

Rigidity (dismissed):

Rigidity (approved):

PR #746, opened 2024-10-05
clean up test-block-generators to be more useful for future work

arvidn:
This tool is useful when validating changes to `run_block_generator()` against all blocks on mainnet or testnet11.
For example, validating that signature validation passes on all blocks.

The old flag `test-backrefs` no longer worked and was removed.

The assert `args.validate && !args.mempool` was incorrect and was removed.

The option `--validate` did not work past the hard fork height, which was fixed. It was also renamed to `--original-generator`.

A new option `--hard-fork-height` was added to support testnet databases.

### DIFF
--- a/crates/chia-tools/src/bin/test-block-generators.rs
+++ b/crates/chia-tools/src/bin/test-block-generators.rs
@@ -30,14 +30,17 @@ struct Args {
     #[arg(long, default_value_t = false)]
     mempool: bool,
 
-    /// re-serialize each generator using backrefs and ensure it still produces
-    /// the same output
-    #[arg(long, default_value_t = false)]
-    test_backrefs: bool,
-
-    /// Compare the output from the default ROM running in consensus mode.
+    /// Compare the output from the default ROM running in consensus mode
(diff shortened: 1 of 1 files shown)


crates/chia-tools/src/bin/test-block-generators.rs
```diff
@@ -158,17 +177,25 @@ fn main() {
 
                 // after the hard fork, we run blocks without paying for the
                 // CLVM generator ROM
-                let block_runner = if height >= 5_496_000 {
+                let block_runner = if args.original_generator || height >= args.hard_fork_height {
```
AmineKhaldi, line 180:
```suggestion
                let block_runner = if !args.original_generator || height >= args.hard_fork_height {
```

arvidn: I don't think this is correct. the point of `--original-generator` is to compare the original one to the hard-fork one. The original one is called on line 211, so we need to use the hard-fork one here

AmineKhaldi (approved):

PR #748, opened 2024-10-07
[CHIA-1563] add function to run a *trusted* block and return additions and removals.

arvidn:
This adds a function called `additions_and_removals()`, which is similar to `run_block_generator()` and `run_block_generator2()` but with some important differences.

`additions_and_removals()`:
* assumes that the block is valid. It's only meant to be run on blocks we've already validated (and perhaps pulled out of the blockchain database).
* only return the created coins ("additions") and spent coins ("removals"). The created coins also include the optional hint.

# motivation

The two main reasons to introduce this function are:
* It's faster than `run_block_generator()`
* It completes the two use cases `chia-blockchain` has for running blocks. Letting `run_block_generator()` be more specialized to always fully validate blocks (e.g. the signature).

## running trusted blocks

In `chia-blockchain` we sometimes validate blocks completely (1). Ensuring their conditions hold, their signatures are correct, the proof of space is correct etc. This is done as part of syncing a node as well as keeping up with the chain and maintaining the mempool. When we receive a transaction for the mempool, we fully validate it before accepting it. When we receive an `UnfinishedBlock`, we fully validated it.

The other use case for running block generators is to just get additions and removals (2) to respond to an RPC or to rebuild the coin store during a reorg (or when adding an orphaned block). This, case 2, could be replaced by the cheaper call to `additions_and_removals().

The non-test call sites of `get_name_puzzle_conditions()` are:

* `CostLogger.add_cost()` - part of the spend-sim, to compute the cost of a block in the simulator. This requires the full `run_block_generator() to compute the cost. Case (1).
* `Blockchain.run_single_block()` - this is to catch up a `ForkInfo` object with all the additions and removals of a fork of the chain. We need to do this when adding orphaned blocks, or reorging. Case (2).
* `batch_pre_validate_blocks()` - this is run when validating full blocks, either because we're syncing or just received a new block extending the blockchain. Case (1).
* `_run_generator()` - this is called when validating an `UnfinishedBlock`. Case (1).
`FullNodeAPI.request_block_header()` - this is an RPC called on existing, validated, blocks in the chain. Case (2).
* `FullBlockApi.request_block_header()` - To construct a block header from a full block, we need the additions and removals, so we recompute them from a trusted block. Case (2).

### git grep output, for non-tests and non-imports:

```
git grep -B 1 -A 1 get_name_puzzle_conditions

chia/clvm/spend_sim.py-        program: BlockGenerator = simple_solution_generator(spend_bundle)
chia/clvm/spend_sim.py:        npc_result: NPCResult = get_name_puzzle_conditions(
chia/clvm/spend_sim.py-            program,
--
chia/consensus/block_creation.py-def compute_block_cost(generator: BlockGenerator, constants: ConsensusConstants, height: uint32) -> uint64:
chia/consensus/block_creation.py:    result: NPCResult = get_name_puzzle_conditions(
chia/consensus/block_creation.py-        generator, constants.MAX_BLOCK_COST_CLVM, mempool_mode=True, height=height, constants=constants
--
chia/consensus/blockchain.py-            assert block.foliage_transaction_block is not None
chia/consensus/blockchain.py:            npc = get_name_puzzle_conditions(
chia/consensus/blockchain.py-                block_generator,
--
chia/consensus/multiprocess_validation.py-                assert block_generator.program == block.transactions_generator
chia/consensus/multiprocess_validation.py:                npc_result = get_name_puzzle_conditions(
chia/consensus/multiprocess_validation.py-                    block_generator,
--
chia/consensus/multiprocess_validation.py-        assert block_generator.program == unfinished_block.transactions_generator
chia/consensus/multiprocess_validation.py:        npc_result: NPCResult = get_name_puzzle_conditions(
chia/consensus/multiprocess_validation.py-            block_generator,
--
chia/full_node/full_node_api.py-                functools.partial(
chia/full_node/full_node_api.py:                    get_name_puzzle_conditions,
chia/full_node/full_node_api.py-                    block_generator,
```

## Performance

One of the main reasons to justify this function, rather than just calling `run_block_generator()` is the time you can save. Here's the benchmark for block 4671894 on mainnet. The plot below compares `run_block_generator()` (slowest), `run_block_generator2()` (middle) and `additions_and_removals()` (fastest).

image

### DIFF
--- a/crates/chia-consensus/src/gen/additions_and_removals.rs
+++ b/crates/chia-consensus/src/gen/additions_and_removals.rs
@@ -1,0 +1,234 @@
+use crate::gen::run_block_generator::setup_generator_args;
+use crate::gen::run_block_generator::subtract_cost;
+use chia_protocol::Coin;
+
+use crate::allocator::make_allocator;
+use crate::consensus_constants::ConsensusConstants;
+use crate::gen::validation_error::{atom, first, next, rest, ErrorCode, ValidationErr};
+use chia_protocol::{Bytes, Bytes32};
+use clvm_traits::FromClvm;
+use clvm_utils::{tree_hash_cached, TreeHash};
+use clvmr::allocator::NodePtr;
+use clvmr::chia_dialect::ChiaDialect;
+use clvmr::reduction::Reduction;
+use clvmr::run_program::run_program;
+use clvmr::serde::node_from_bytes_backrefs_record;
+use std::collections::HashMap;
+
+/// Run a *trusted* block generator and return its additions and removals. This
+/// function does not validate the block, it is assumed to be valid.
+/// The returned vectors are additions (with hints) and removals.
+#[allow(clippy::type_complexity)]
+pub fn additions_and_removals<GenBuf: AsRef<[u8]>, I: IntoIterator<Item = GenBuf>>(
+    program: &[u8],
+    block_refs: I,
+    flags: u32,
+    constants: &ConsensusConstants,
+) -> Result<(Vec<(Coin, Option<Bytes>)>, Vec<Coin>), ValidationErr>
+where
+    <I as IntoIterator>::IntoIter: DoubleEndedIterator,
+{
+    let mut a = make_allocator(flags);
+    let mut additions = Vec::<(Coin, Option<Bytes>)>::new();
+    let mut removals = Vec::<Coin>::new();
+
+    let mut cost_left = constants.max_block_cost_clvm;
+
+    let (program, backrefs) = node_from_bytes_backrefs_record(&mut a, program)?;
+
+    let args = setup_generator_args(&mut a, block_refs)?;
+    let dialect = ChiaDialect::new(flags);
+
+    let Reduction(clvm_cost, mut all_spends) =
+        run_program(&mut a, &dialect, program, args, cost_left)?;
+
+    subtract_cost(&a, &mut cost_left, clvm_cost)?;
+    all_spends = first(&a, all_spends)?;
+
+    let mut cache = HashMap::<NodePtr, TreeHash>::new();
+    // at this point all_spends is a list of:
+    // (parent-coin-id puzzle-reveal amount solution . extra)
+    // where extra may be nil, or additional extension data
+
+    while let Some((spend, tail)) = a.next(all_spends) {
+        all_spends = tail;
+        // process the spend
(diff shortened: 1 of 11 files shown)


matt-o-how (approved):
I like this direction

crates/chia-consensus/src/gen/additions_and_removals.rs
```diff
@@ -0,0 +1,234 @@
...
+                amount,
+            };
+
+            // there was another item in the list
+            // the item was a cons-box, and params is the left-hand
+            // side, the list element
+
+            let hint =
+                if let Ok(((hint, _), _)) = <((Bytes, NodePtr), NodePtr)>::from_clvm(&a, hint) {
+                    if hint.len() <= 32 {
```
Rigidity, line 105: A hint isn't valid unless it's exactly 32 bytes, so you could just use `Bytes32` and skip this check here right?

crates/chia-consensus/src/gen/additions_and_removals.rs
```diff
@@ -0,0 +1,234 @@
...
+            };
+            let coin_id = removal.coin_id();
+            expect_removals.insert(removal);
+            for add in &spend.create_coin {
+                let addition = Coin {
+                    parent_coin_info: coin_id,
+                    puzzle_hash: add.puzzle_hash,
+                    amount: add.amount,
+                };
+                let hint = if add.hint != NodePtr::NIL && a.atom_len(add.hint) <= 32 {
```
Rigidity, line 207: I think the hint length should be `== 32`

Rigidity (approved):

crates/chia-consensus/src/gen/additions_and_removals.rs
arvidn: the first incarnation of this patch required it to be exactly 32 bytes, but the existing rules (in `run_block_generator()`) is that as long as it's <= 32 bytes, we still pass it on to python. So, this is required to stay consistent with that behavior. I discovered this in the unit test where there is such a coin created.

crates/chia-consensus/src/gen/additions_and_removals.rs
arvidn: it's to maintain consistency with `run_block_generator()`. This isn't a consensus rule, it's just a rule full nodes enforce to make the hint table reasonable.

PR #755, opened 2024-10-09
Add needed items for sdist

emlowe:
Add in some missing bits in wheel/Cargo.toml and some hacks around missing parts in the tar archive

Main items are:
needing to add in `chia-client` and `chia-ssl` to `wheel/Cargo.toml`
needing to remove the `../LICENSE` from `wheel/Cargo.toml` - this resulted in the tar not having the LICENSE file but also pip expecting it, so I just removed it for now.
I noticed the tar was missing the `src` directory, so I had to manually add in the src directory to the tar archive. 

This does make a` tar.gz` that does build on my riscv machine and also my mac.

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -258,6 +258,14 @@ jobs:
         run: |
           pip install maturin
           maturin sdist -m wheel/Cargo.toml
+          cd target/wheels
+          dirname=`basename chia_rs*.tar.gz .tar.gz`
+          echo $dirname
+          mkdir $dirname
+          cp -r ./src/ $dirname/src/
+          gunzip chia_rs*.tar.gz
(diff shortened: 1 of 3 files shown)


arvidn (commented):
It's not obvious why `chia-client` and `chia-ssl` are needed to build the wheel. Is maturin failing to bundle transitive dependencies? I wouldn't expect these to be part of the dependency tree at all though

emlowe:
When maturin builds the sdist, it puts in the tar the `Cargo.toml` from the root workspace and not from the wheel directory. They are listed in the root Cargo as deps and default features.

arvidn (approved):

arvidn:
I see. thanks for the explanation.

Rigidity (approved):
lgtm for now

PR #758, opened 2024-10-17
[CHIA-1608] Use a cache eviction logic that matches our BlsCache usage pattern

AmineKhaldi:


arvidn (commented):
did you run the benchmark to compare against `main`? I think that would be an interesting comparison.

Could you also make sure that it's OK to evict on a cache hit? i.e. that we never rely on two or more cache hits?

If there's a reorg we might get more cache hits, but we should probably not optimize for that case. If there are competing unfinished blocks is another example where we'll have multiple cache hits.

It might make sense to make the eviction explicit. When a block is fully validated and added to the chain it should be safe to evict its keys from the cache.

crates/chia-bls/src/bls_cache.rs
```diff
@@ -88,6 +113,25 @@ impl BlsCache {
...
+        for (pk, msg) in pks_msgs {
+            let mut hasher = Sha256::new();
+            let mut aug_msg = pk.borrow().to_bytes().to_vec();
+            aug_msg.extend_from_slice(msg.as_ref());
+            hasher.update(&aug_msg);
+            let hash: [u8; 32] = hasher.finalize();
+            hashes_to_remove.insert(hash);
+        }
+        let mut c = self.cache.lock().expect("cache");
+        c.items.retain(|h, _| !hashes_to_remove.contains(h));
```
arvidn, line 132: this is a hash map. surely it's faster to remove items individually, using its own index, rather than iterating over all items. Did you benchmark that?
This looks a bit unintuitive I think. If you benchmarked and this is faster, please leave a comment saying so

crates/chia-bls/src/bls_cache.rs
```diff
@@ -88,6 +113,25 @@ impl BlsCache {
...
+        c.insertions_order.retain(|h| !hashes_to_remove.contains(h));
```
arvidn, line 133: it might make sense to explain why we need a linear scan here too. at least until we use a multi-index container instead

wheel/generate_type_stubs.py
```diff
@@ -368,6 +368,7 @@ def len(self) -> int: ...
     def aggregate_verify(self, pks: list[G1Element], msgs: list[bytes], sig: G2Element) -> bool: ...
     def items(self) -> list[tuple[bytes, GTElement]]: ...
     def update(self, other: Sequence[tuple[bytes, GTElement]]) -> None: ...
+    def evict(self, pks: list[G1Element], msgs: list[bytes]) -> None: ...
```
arvidn, line 371: both `update()` and `items()` use a list of tuples rather than two separate lists for the keys and messages. Is there a good reason to diverge from that for this function?

AmineKhaldi: Yes, this follows the expected use case, which is more similar to `aggregate_verify` than `update` and `items` in terms of input data.

arvidn (dismissed):

crates/chia-bls/src/bls_cache.rs
```diff
@@ -17,16 +17,35 @@ use crate::{GTElement, PublicKey, Signature};
...
+}
+
+impl BlsCacheData {
+    pub fn put(&mut self, hash: [u8; 32], pairing: GTElement) {
+        // If the cache is full, remove the oldest item.
+        if self.items.len() == self.capacity.get() {
+            if let Some((oldest_key, _)) = self.items.pop_front() {
+                self.items.remove(&oldest_key);
+            }
+        }
```
arvidn, line 35:
```suggestion
        while self.items.len() > self.capacity.get() {
            let Some((oldest_key, _)) = self.items.pop_front() else { continue; };
            self.items.remove(&oldest_key);
        }
```
I'm not sure if this makes sense. It might be a bit more defensive

crates/chia-bls/src/bls_cache.rs
```diff
@@ -167,6 +205,20 @@ impl BlsCache {
         }
         Ok(())
     }
+
+    #[pyo3(name = "evict")]
+    pub fn py_evict(&mut self, pks: &Bound<'_, PyList>, msgs: &Bound<'_, PyList>) -> PyResult<()> {
```
arvidn, line 210:
```suggestion
    pub fn py_evict(&self, pks: &Bound<'_, PyList>, msgs: &Bound<'_, PyList>) -> PyResult<()> {
```

crates/chia-bls/src/bls_cache.rs
```diff
@@ -88,6 +110,22 @@ impl BlsCache {
         let hash: [u8; 32] = hasher.finalize();
         self.cache.lock().expect("cache").put(hash, gt);
     }
+
+    pub fn evict<Pk, Msg>(&mut self, pks_msgs: impl IntoIterator<Item = (Pk, Msg)>)
```
arvidn, line 114:
```suggestion
    pub fn evict<Pk, Msg>(&self, pks_msgs: impl IntoIterator<Item = (Pk, Msg)>)
```

arvidn (changes requested):

arvidn (approved):

PR #761, opened 2024-10-21
add `bytes*.fill()` helper

altendky:


wheel/python/chia_rs/sized_byte_class.py
```diff
@@ -78,6 +79,27 @@ def random(
     def secret(cls: Type[_T_SizedBytes]) -> _T_SizedBytes:
         return cls.random(r=system_random)
 
+    @classmethod
+    def fill(cls: Type[_T_SizedBytes], blob: bytes, fill: bytes, align: Literal["<", ">"] = ">") -> _T_SizedBytes:
+        if len(blob) == cls._size:
```
arvidn, line 84: if `len(blob) > cls._size`, do we want to truncate and construct `cls` then too?

altendky: I am tending towards no.  It seems like with these classes that truncation would be surprising.  I think most use cases for this are tests where you are trying to create unique values.  If your blob is too long and is truncated you hazard unexpectedly having multiple of the same value generated.

This isn't meant to be a 'format' level tool, but I will note that format just creates a larger output in this case.

PR #799, opened 2024-11-25
clippy clean up

altendky:


arvidn:
I incorporated these changes in https://github.com/Chia-Network/chia_rs/pull/798
Would they conflict or merge cleanly you think?

altendky:
I would guess if they are in their own diff hunks and exactly the same change they would merge cleany.  But, `main` is checking fine now so whatever.  `:]`

PR #800, opened 2024-11-25
Add negative tests  and pytests for `validate_clvm_and_signature()`

matt-o-how:
Extended the current rust tests to check against an incorrect message and an incorrect key, and added a python test for one of the cases

### DIFF
--- a/crates/chia-consensus/src/spendbundle_validation.rs
+++ b/crates/chia-consensus/src/spendbundle_validation.rs
@@ -177,6 +177,38 @@ ff01\
             236,
         )
         .expect("SpendBundle should be valid for this test");
+        
+        // test wrong message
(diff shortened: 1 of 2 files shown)


crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -251,14 +291,69 @@ ff01\
...
+            coin_spends,
+            aggregated_signature: sig,
+        };
+        let result = validate_clvm_and_signature(
+            &spend_bundle,
+            TEST_CONSTANTS.max_block_cost_clvm,
+            &TEST_CONSTANTS,
+            246,
+        );
+        assert!(matches!(result, Err(ErrorCode::BadAggregateSignature)));
```
arvidn, line 349: this looks like a verbatim copy of the block above labeled "test sk message"

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -300,6 +395,65 @@ ff01\
...
```
arvidn, line 427: the only difference between this and the other "test wrong message" block is the condition code. It seems like a perfect opportunity for parameterizing the test with `rstest`

arvidn (approved):

PR #806, opened 2024-11-29
[CHIA-1927] add tool to roll-back a blockchain database file to a previous height

arvidn:
if `--rollback` is not specified, prints the peak height and hash.

If `--rollback` is specified with a height lower than the current peak, the database is updated like this:
* the peak is updated to point to the block at the specified height
* all blocks with a height greater than the new peak are marked as `in_main_chain=0`
* all coins *created* after the new peak are deleted from the `coin_records` table
* all coins *spent* after the new peak are marked as unspent

Example use:

```
cargo run --release --bin rollback-blockchain-db -- --rollback 6255740 ~/.chia/mainnet/db/blockchain_v2_mainnet.sqlite
```

Example output:

```
opening blockchain database file: /home/arvid/.chia/mainnet/db/blockchain_v2_mainnet.sqlite
peak height: 6267502
peak hash: 9fb57e29c398ffa69895d31a81af033e7a10cbf37625dcecd52307dfbefaa467
Rolling back to height 6255740
updating new peak to 9fb57e29c398ffa69895d31a81af033e7a10cbf37625dcecd52307dfbefaa467
clearing in_main_chain for all blocks with height above 6255740
deleting all coins created after height 6255740
unspending all coins spent after height 6255740
```

### DIFF
--- a/crates/chia-tools/Cargo.toml
+++ b/crates/chia-tools/Cargo.toml
@@ -75,3 +75,8 @@ bench = false
 name = "validate-blockchain-db"
 test = false
 bench = false
+
+[[bin]]
+name = "rollback-blockchain-db"
+test = false
+bench = false
--- a/crates/chia-tools/src/bin/rollback-blockchain-db.rs
+++ b/crates/chia-tools/src/bin/rollback-blockchain-db.rs
@@ -1,0 +1,111 @@
+use clap::Parser;
+use rusqlite::Connection;
+
+/// Prints information about a blockchain peak and can roll it back to a
+/// specified height
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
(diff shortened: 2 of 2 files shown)


AmineKhaldi (commented):
Is there anything preventing us from having these queries as part of a `chia db` command (`chia db rollback` for example) so that it can conveniently benefit the community right from the `chia-blockchain` repo?

Rigidity (approved):

arvidn:
> Is there anything preventing us from having these queries as part of a chia db command (chia db rollback for example) so that it can conveniently benefit the community right from the chia-blockchain repo?

It would be more work and I don't think it's a priority. the community can use this tool too

AmineKhaldi (approved):

PR #825, opened 2024-12-16
catchup fuzzer versions to 0.17.0

altendky:


arvidn:
I'm not sure if this matters. I don't believe we publish the fuzzers anyway. But I don't see any downside either.

PR #836, opened 2025-01-01
match python exception message casing for now

altendky:


Rigidity:
Can we at least be consistent within this enum?

altendky:
handling via [Chia-Network/chia-blockchain@`1d0df78` (#18716)](https://github.com/Chia-Network/chia-blockchain/pull/18716/commits/1d0df78bdd642337c10d4d66e27658addc47113a) instead.  either way it's mostly pointing out a mess to clean up.

PR #869, opened 2025-01-13
add class to build a compressed block incrementally

arvidn:
# purpose

This PR instroduces a type called `BlockBuilder` which lets you add spend bundles, one at a time, until the block cost reaches the limit. As spend bundles are added, they are serialized into the final block generator with back-references (deduplication). This makes the final *byte-cost* difficult to predict, hence the iterative approach.

This patch reproduces some logic from the current mempool block building function, such as the logic of when to give up adding more spend bundles to the block.

# cost

When adding a spend bundle you also have to specify the cost of it. This cost is computed when the mempool item is first validated. In existing block creation logic, we use the *total* cost of the spend bundle, execution cost, condition cost and byte-cost. However, with the `BlockBuilder`, the accompanied cost is supposed to only be execution cost + condition cost. This is because the byte-cost is only known after the spend bundle has been added.

# test

The test picks random permutations of the spend bundles in the `test-bundles` directory and creates blocks by feeding the bundles one-by-one. The test tracks how long it takes to fill the block, and how long the most expensive call to `add_spend_bundle()` took.

We need to compute the cost of each spend bundle, which is done on start-up, when all bundles are being loaded.

The test then parses the resulting block and runs it through `run_block_generator2()` (just like consensus). This ensures the signature is correct, and that the cost is correct.

See the example output below, run on Ubuntu/Threadripper:

```
loading spend bundles from disk
loaded 90 spend bundles
idx:   0 built block in 11.75 seconds, cost: 10998575664 skipped:  3 longest-call:  9.99s TX: 37
idx:   1 built block in 10.05 seconds, cost: 10983357358 skipped:  7 longest-call:  8.93s TX: 32
idx:   2 built block in 10.88 seconds, cost: 10847563861 skipped:  7 longest-call:  9.50s TX: 62
idx:   3 built block in  2.85 seconds, cost: 10965747249 skipped:  7 longest-call:  0.76s TX: 54
idx:   4 built block in 11.45 seconds, cost: 10754122135 skipped:  7 longest-call:  9.69s TX: 40
idx:   5 built block in  3.31 seconds, cost: 10583555007 skipped:  7 longest-call:  0.88s TX: 46
idx:   6 built block in  9.54 seconds, cost: 10993197686 skipped:  7 longest-call:  8.30s TX: 55
idx:   7 built block in 19.23 seconds, cost: 10977464946 skipped:  7 longest-call: 11.16s TX: 40
idx:   8 built block in 10.08 seconds, cost: 10758545088 skipped:  7 longest-call:  5.90s TX: 39
idx:   9 built block in 10.68 seconds, cost: 10889601079 skipped:  7 longest-call:  8.43s TX: 36
idx:  10 built block in 11.08 seconds, cost: 10996639192 skipped:  3 longest-call: 10.00s TX: 29
idx:  11 built block in 11.44 seconds, cost: 10970637745 skipped:  7 longest-call:  9.22s TX: 34
idx:  12 built block in 11.48 seconds, cost: 10996973255 skipped:  6 longest-call:  8.99s TX: 37
idx:  13 built block in  8.75 seconds, cost: 10999105794 skipped:  6 longest-call:  7.30s TX: 31
idx:  14 built block in 19.63 seconds, cost: 10994129896 skipped:  6 longest-call:  9.26s TX: 40
idx:  15 built block in 11.53 seconds, cost: 10664917802 skipped:  7 longest-call:  8.81s TX: 55
idx:  16 built block in 12.19 seconds, cost: 10994836257 skipped:  6 longest-call: 10.36s TX: 56
idx:  17 built block in  1.93 seconds, cost: 10913252607 skipped:  7 longest-call:  0.34s TX: 42
idx:  18 built block in  2.22 seconds, cost: 10990565458 skipped:  7 longest-call:  0.56s TX: 51
idx:  19 built block in  9.36 seconds, cost: 10997538040 skipped:  3 longest-call:  8.60s TX: 28
idx:  20 built block in 10.04 seconds, cost: 10995570352 skipped:  0 longest-call:  9.38s TX: 23
idx:  21 built block in  9.24 seconds, cost: 10793651464 skipped:  7 longest-call:  6.58s TX: 61
idx:  22 built block in 19.62 seconds, cost: 10996436556 skipped:  4 longest-call: 10.26s TX: 27
idx:  23 built block in 11.23 seconds, cost: 10741182550 skipped:  7 longest-call:  9.00s TX: 47
idx:  24 built block in 18.68 seconds, cost: 10996512566 skipped:  4 longest-call: 11.17s TX: 35
idx:  25 built block in  7.65 seconds, cost: 10511794612 skipped:  7 longest-call:  5.78s TX: 31
idx:  26 built block in  2.80 seconds, cost: 10880611048 skipped:  7 longest-call:  0.61s TX: 55
idx:  27 built block in 10.76 seconds, cost: 10761481781 skipped:  7 longest-call:  9.60s TX: 50
idx:  28 built block in 10.80 seconds, cost: 10700285540 skipped:  7 longest-call:  9.81s TX: 33
idx:  29 built block in  9.47 seconds, cost: 10999810076 skipped:  4 longest-call:  6.80s TX: 32
test gen::build_compressed_block::tests::test_build_block ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2250 filtered out; finished in 20.55s
```

Some of these timings are not great. There's separate work in progress to optimize the CLVM compression.

# test data

This PR adds some test spend bundles to the `test-bundles` directory. My intention is to, in a future update, make CI pull down additional test bundles ( https://github.com/Chia-Network/test-cache/pull/16 ) to run more comprehensive tests.

```
$ ls -la test-bundles/ | wc -l
96
$ du -sh test-bundles/
1.5M    test-bundles/
```


crates/chia-consensus/src/gen/build_compressed_block.rs
```diff
@@ -0,0 +1,417 @@
...
+            let item = a.new_pair(puzzle, item)?;
+            // parent-id
+            let parent_id = a.new_atom(&spend.coin.parent_coin_info)?;
+            let item = a.new_pair(parent_id, item)?;
+
+            spend_list = a.new_pair(item, spend_list)?;
+        }
+
+        let (done, state) = self.ser.add(a, spend_list)?;
+        assert!(!done);
```
AmineKhaldi, line 142: This will run in release builds as well, is this intentional (vs for example `debug_assert!`)?

crates/chia-consensus/src/gen/build_compressed_block.rs
```diff
@@ -0,0 +1,417 @@
...
+
+        self.byte_cost = (self.ser.size() + 2) * constants.cost_per_byte;
+        // closing the lists at the end needs 2 extra bytes
```
AmineKhaldi, line 145: This comment is meant for the line above it, not below it, right?

crates/chia-consensus/src/gen/build_compressed_block.rs
```diff
@@ -0,0 +1,417 @@
...
+            self.num_skipped += 1;
+            return Ok((false, result(self.num_skipped)));
+        }
+        self.block_cost += cost;
+        self.signature.aggregate(&bundle.aggregated_signature);
+
+        // if we're very close to a full block, we're done. It's very unlikely
+        // any transaction will be smallar than MIN_COST_THRESHOLD
+        let result = if self.byte_cost + self.block_cost + MIN_COST_THRESHOLD
+            >= constants.max_block_cost_clvm
```
AmineKhaldi, line 159: Same question as above about whether we're intentionally changing behavior to disallow values equal to the maximum allowed.

crates/chia-consensus/src/gen/build_compressed_block.rs
```diff
@@ -0,0 +1,417 @@
...
+        mut self,
+        constants: &ConsensusConstants,
+    ) -> io::Result<(Vec<u8>, Signature, u64)> {
+        let (done, _) = self.ser.add(&self.allocator, self.allocator.nil())?;
+        assert!(done);
+
+        // add the size cost before returning it
+        self.block_cost += self.ser.size() * constants.cost_per_byte;
+
+        assert!(self.block_cost <= constants.max_block_cost_clvm);
```
AmineKhaldi, line 178: Same question about whether it's intentional to want to run these asserts even in release builds.

crates/chia-consensus/src/gen/build_compressed_block.rs
```diff
@@ -0,0 +1,417 @@
...
+        for entry in &all_bundles {
+            let (cond, cost, _) = entry.as_ref();
+            if *cost != last_cost {
+                println!("\n== {cost}");
+                last_cost = *cost;
+            }
+            print!("{}.bundle ", cond.name());
+        }
+        println!("\n");
+        */
```
AmineKhaldi, line 340: This looks like debugging leftovers, right?

crates/chia-consensus/src/gen/build_compressed_block.rs
```diff
@@ -0,0 +1,417 @@
...
+            // this field
+            for s in &mut conds.spends {
+                s.flags = 0;
+                // when parsing conditions, create coin conditions are stored in
+                // a hash set to cheaply check for double spends. This means the
+                // order of this collection is not deterministic. In order to
+                // compare to the generator output later, we need to sort both.
+                s.create_coin.sort();
+            }
+            all_bundles.push(Box::new((bundle, cost, conds)));
```
AmineKhaldi, line 326: Out of curiosity, why did you need a box for these vector elements?

crates/chia-consensus/src/gen/build_compressed_block.rs
```diff
@@ -0,0 +1,417 @@
...
+                let (generator, signature, cost) =
+                    builder.finalize(&TEST_CONSTANTS).expect("finalize()");
+
+                println!(
+                    "idx: {seed:3} built block in {:>5.2} seconds, cost: {cost:11} skipped: {skipped:2} longest-call: {max_call_time:>5.2}s TX: {num_tx}",
+                    start.elapsed().as_secs_f32()
+                );
+
+                //fs::write(format!("../../{seed}.generator"), generator.as_slice())
+                //    .expect("write generator");
```
AmineKhaldi, line 387: Leftover?

tests/test_block_builder.py
```diff
@@ -0,0 +1,75 @@
...
+        if len(name) != 64 + 7:
+            continue
+        with open(join("test-bundles", name), "rb") as f:
+            sb = SpendBundle.from_bytes(f.read())
+            conds, bls_cache, duration = validate_clvm_and_signature(
+                sb, 11000000000, DEFAULT_CONSTANTS, 5000000
+            )
+            cost = uint64(conds.execution_cost + conds.condition_cost)
+
+            # print(f"{name} {conds.execution_cost} {conds.condition_cost}")
```
AmineKhaldi, line 33: Left intentionally?

crates/chia-consensus/src/gen/build_compressed_block.rs
```diff
@@ -0,0 +1,417 @@
...
+use pyo3::prelude::*;
+
+/// Maximum number of mempool items that can be skipped (not considered) during
+/// the creation of a block bundle. An item is skipped if it won't fit in the
+/// block we're trying to create.
+const MAX_SKIPPED_ITEMS: u32 = 6;
+
+/// Typical cost of a standard XCH spend. It's used as a heuristic to help
+/// determine how close to the block size limit we're willing to go.
+const MIN_COST_THRESHOLD: u64 = 6_000_000;
```
AmineKhaldi, line 18: Would it be reasonable to allow us to build the class with these parameters, and default to these values if they're not provided?

crates/chia-consensus/src/gen/build_compressed_block.rs
```diff
@@ -0,0 +1,417 @@
...
+    /// add a spend bundle to the generator. The cost must be *only* the CLVM
+    /// execution cost + the cost of the conditions. It must not include the byte cost
+    /// of the bundle. The byte cost is unpredictible as the generator is being
+    /// compressed. The true byte cost is computed by this function.
+    /// returns true if this bundle could be added to the generator, false otherwise
+    pub fn add_spend_bundle(
+        &mut self,
+        bundle: &SpendBundle,
+        cost: u64,
+        constants: &ConsensusConstants,
```
AmineKhaldi, line 107: Because these are constants, I would pass them to the class and simplify all the methods where they're needed.

crates/chia-consensus/src/gen/build_compressed_block.rs
```diff
@@ -0,0 +1,417 @@
...
+            BuildBlockResult::Done
+        } else {
+            BuildBlockResult::KeepGoing
+        };
+        Ok((true, result))
+    }
+
+    pub fn finalize(
+        mut self,
+        constants: &ConsensusConstants,
```
AmineKhaldi, line 170: Same suggestion as above about constants.

crates/chia-consensus/src/gen/build_compressed_block.rs
```diff
@@ -0,0 +1,417 @@
...
+
+        let mut ser = Serializer::new(Some(sentinel));
+        ser.add(&a, quoted_list)?;
+
+        Ok(Self {
+            allocator: a,
+            signature: Signature::default(),
+            sentinel,
+            // TODO: where does this cost overhead come from?
+            block_cost: 20,
```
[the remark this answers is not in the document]
arvidn: the cost 20 is intentional. Without it we don't get the correct cost. It would be good to understand where the cost is coming from though. It's probably the execution cost of the quotes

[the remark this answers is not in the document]
Rigidity: I believe 20 is the cost to allocate a byte in CLVM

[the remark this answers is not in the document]
arvidn: the byte cost is 12000 per byte

crates/chia-consensus/src/gen/build_compressed_block.rs
arvidn: I like this idea, however, that would require a lifetime parameter on `BlockBuilder`, which isn't supported by pyo3 (kind of understandably).

crates/chia-consensus/src/gen/build_compressed_block.rs
arvidn: yes, this is intentional. I don't think this is expensive, and it would be bad if the assumption doesn't hold

crates/chia-consensus/src/gen/build_compressed_block.rs
arvidn: yes. is there a good reason not to?

crates/chia-consensus/src/gen/build_compressed_block.rs
arvidn: the only way I can think of to avoid a life-time parameter is to *copy* the constants into the class. That seems a bit excessive

crates/chia-consensus/src/gen/build_compressed_block.rs
arvidn: they are shared between the threads below

crates/chia-consensus/src/gen/build_compressed_block.rs
arvidn: it's not really for debugging, it's really handy to learn something about the spend bundles that were loaded. I expect to make the CI use a larger collection later

crates/chia-consensus/src/gen/build_compressed_block.rs
arvidn: also very useful when turning an interesting example into a benchmark

tests/test_block_builder.py
arvidn: yes

crates/chia-consensus/src/gen/build_compressed_block.rs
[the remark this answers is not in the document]
arvidn: I've updated the comment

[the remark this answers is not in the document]
Rigidity: It's 12,000 to store a byte in a block, and 20 to allocate it at runtime, I believe.

crates/chia-consensus/src/gen/build_compressed_block.rs
arvidn: yes, and ideally add whatever test you need to this repository as well, in rust. given that so much of the consensus logic lives in rust now this cannot be avoided. And I don't think it's an argument against moving more logic (and constants) into rust.

crates/chia-consensus/src/gen/build_compressed_block.rs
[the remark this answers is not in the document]
arvidn: it's this cost: https://github.com/Chia-Network/clvm_rs/blob/main/src/run_program.rs#L10

crates/chia-consensus/src/gen/build_compressed_block.rs
AmineKhaldi: I was thinking of two scenarios when I mentioned this:
1. Running a node to experiment with tweaking these values, where right now you're one local edit and a node restart away from achieving that.
2. End-to-end tests, AFAIK it's not currently possible to set those up in this repository, in rust.

crates/chia-consensus/src/gen/build_compressed_block.rs
AmineKhaldi: Hmm I see.

crates/chia-consensus/src/gen/build_compressed_block.rs
AmineKhaldi: Agreed, it's fine as-is then.

AmineKhaldi (approved):

PR #872, opened 2025-01-13
Somewhat opinionated cleanup of exception macro

Rigidity:


altendky:
while the underlying macro has been changed to a derive macro and moved, the same ideas here may still apply.  that said, the new macro's output formatting is wildly bad for reasons i don't know yet and that should be addressed before concerns like here.  but...  this may still be applicable so leaving open with this context.

PR #887, opened 2025-01-17
Remove chia-blockchain dependency for most of chia_rs

matt-o-how:
This PR removes imports from `test_blscache.py` and deletes `test_program_fidelity.py`.

It also removes the python running from `block_record.rs` by implementing `pot_iterations.py` inside the class, and porting the code to a new file called `pot_iterations.rs`. 

There is a corresponding PR to import this code from here in chia-blockchain: https://github.com/Chia-Network/chia-blockchain/pull/19197


crates/chia-protocol/src/pos_quality.rs
```diff
@@ -4,7 +4,9 @@
 pub const UI_ACTUAL_SPACE_CONSTANT_FACTOR: f32 = 0.78;
 
 // TODO: Update this when new plot format releases
-pub fn expected_plot_size(k: u32) -> u64 {
+#[cfg(feature = "py-bindings")]
+#[pyo3::pyfunction]
```
Rigidity, line 8: I think you have to do `#[cfg_attr(feature = "py-bindings", pyo3::pyfunction)]` instead?

wheel/generate_type_stubs.py
```diff
@@ -336,6 +336,35 @@ def get_flags_for_height_and_constants(
     constants: ConsensusConstants
 ) -> int: ...
 
+def calculate_ip_iters(
+    num_sps_sub_slot: int,
+    num_sp_intervals_extra: int,
+    sub_slot_iters: int,
+    signage_point_index: int,
+    required_iters: int,
+) -> int: ...
```
arvidn, line 345: are all of these plain `int` in the python implementation? it seems like it might make sense to make them sized ints

tests/test_pot_iterations.py
```diff
@@ -0,0 +1,216 @@
...
+            == (
+                sp_iters
+                + test_constants.NUM_SP_INTERVALS_EXTRA * sp_interval_iters
+                + required_iters
+            )
+            % ssi
+        )
+        assert sp_iters > ip_iters
+
+    # def test_win_percentage(self):
```
arvidn, line 180: I think this commented-out code can be removed, right?

tests/test_block_record_fidelity.py
```diff
@@ -129,3 +129,28 @@ def wrap_call(expr: str, br: Any) -> str:
...
+    ri = sp_interval_iters - 1
+    rng = Random()
+    rng.seed(1337)
+    br = get_block_record(rng, ssi=ssi, spi=31, ri=ri)
+    with raises(ValueError):
+        res = br.ip_iters_impl(DEFAULT_CONSTANTS)
+
+    br = get_block_record(rng, ssi=ssi, spi=13, ri=1)
+    res = br.ip_iters_impl(DEFAULT_CONSTANTS)
+    assert res is not None
```
arvidn, line 156: not `None`, it seems like the wrong value would also pass here

matt-o-how: Made this a specific value

tests/test_pot_iterations.py
matt-o-how: Done

wheel/generate_type_stubs.py
matt-o-how: Done

crates/chia-protocol/src/block_record.rs
```diff
@@ -121,48 +154,38 @@ impl BlockRecord {
...
-            None,
-            Some(&ctx),
-            )?;
-        ctx.get_item("ret").unwrap().unwrap().extract::<u64>()
-    }
-
-    fn sp_total_iters_impl(&self, py: Python<'_>, constants: &Bound<'_, PyAny>) -> PyResult<u128> {
-        self.sp_sub_slot_total_iters_impl(py, constants)?
-            .checked_add(self.sp_iters_impl(py, constants)? as u128)
+    fn sp_iters_impl(&self, constants: &Bound<'_, PyAny>) -> PyResult<u64> {
```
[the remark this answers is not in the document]
matt-o-how: Done

crates/chia-protocol/src/pot_iterations.rs
```diff
@@ -0,0 +1,292 @@
...
+}
+
+#[cfg(feature = "py-bindings")]
+#[pyo3::pyfunction]
+#[pyo3(name = "calculate_sp_interval_iters")]
+pub fn py_calculate_sp_interval_iters(
+    num_sps_sub_slot: u32,
+    sub_slot_iters: u64,
+) -> pyo3::PyResult<u64> {
+    calculate_sp_interval_iters(num_sps_sub_slot, sub_slot_iters).map_err(Into::into)
```
arvidn, line 83: I think you can use `?` instead of explicitly mapping the error with `Into::into`.

i.e.
```
Ok(calculate_sp_interval_iters(num_sps_sub_slot, sub_slot_iters)?)
```

crates/chia-protocol/src/pot_iterations.rs
```diff
@@ -0,0 +1,292 @@
...
+    sub_slot_iters: u64,
+    signage_point_index: u32,
+    required_iters: u64,
+) -> Result<u64> {
+    let sp_interval_iters = calculate_sp_interval_iters(num_sps_sub_slot, sub_slot_iters)?;
+    let sp_iters = calculate_sp_iters(num_sps_sub_slot, sub_slot_iters, signage_point_index)?;
+    if sp_iters % sp_interval_iters != 0 || sp_iters > sub_slot_iters {
+        return Err(Error::Custom(format!(
+            "Invalid sp iters {sp_iters} for this ssi {sub_slot_iters}"
+        )));
```
arvidn, line 48: I really think we should have error codes instead of free-form strings. Handling errors like this, programatically, is very brittle.

arvidn (commented):
I would suggest you have tests that attempt to overflow integers, and especially look at all `as` casts and see if there's a way to truncate integers, in tests

crates/chia-protocol/src/pot_iterations.rs
matt-o-how: done

crates/chia-protocol/src/pos_quality.rs
```diff
@@ -0,0 +1,34 @@
...
+pub fn py_expected_plot_size(k: u32) -> pyo3::PyResult<u64> {
+    // """
+    // Given the plot size parameter k (which is between 32 and 59), computes the
+    // expected size of the plot in bytes (times a constant factor). This is based on efficient encoding
+    // of the plot, and aims to be scale agnostic, so larger plots don't
+    // necessarily get more rewards per byte. The +1 is added to give half a bit more space per entry, which
+    // is necessary to store the entries in the plot.
+    // """
+
+    expected_plot_size(k).map_err(Into::into)
```
arvidn, line 33: I thought clippy suggested `Ok(call()?)` over this. I think they do the same thing, as the `?`-operator also implicitly converts the error type

arvidn (dismissed):

arvidn:
we should squash this when landing it

crates/chia-protocol/src/pos_quality.rs
arvidn: @matt-o-how what happened with this? Did you try it and it didn't work?

crates/chia-protocol/src/block_record.rs
```diff
@@ -171,38 +164,38 @@ impl BlockRecord {
...
+        ChiaToPython::to_python(&self.py_ip_iters_impl(constants)?, py)
     }
 
     fn sp_total_iters<'a>(
         &self,
         py: Python<'a>,
         constants: &Bound<'_, PyAny>,
     ) -> PyResult<Bound<'a, PyAny>> {
-        ChiaToPython::to_python(&self.sp_total_iters_impl(py, constants)?, py)
+        ChiaToPython::to_python(&self.sp_total_iters_impl(constants)?, py)
```
arvidn, line 199: it's not clear to me why we need these extra wrapper functions. I would expect we could have pure rust member functions and (when enabled) python wrappers that call the rust functions. This appears to add 2 layers of python wrapper around the rust functions

tests/test_pot_iterations.py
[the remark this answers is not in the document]
matt-o-how: yes

tests/test_block_record_fidelity.py
```diff
@@ -129,3 +129,28 @@ def wrap_call(expr: str, br: Any) -> str:
         return f"V:{ret}"
     except Exception as e:
         return f"E:{e}"
+
+
+# TODO: more thoroughly check these new functions which use self
```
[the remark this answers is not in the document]
matt-o-how: yes, the block record uses its own data in those function

tests/test_pot_iterations.py
```diff
@@ -0,0 +1,195 @@
+from __future__ import annotations
+
+from pytest import raises
+
+from run_gen import DEFAULT_CONSTANTS
+
+# from chia.consensus.pos_quality import _expected_plot_size
```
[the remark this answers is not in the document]
matt-o-how: Removed

crates/chia-protocol/src/pos_quality.rs
matt-o-how: Correct - I tried this change and reverted it

arvidn (approved):

PR #942, opened 2025-02-24
Disable FF with relative timelock conditions and disable amount changing

matt-o-how:


Rigidity (dismissed):
Looks fine to me

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -99,14 +111,15 @@ impl SpendVisitor for MempoolVisitor {
             Condition::AggSigMe(_, _)
             | Condition::AggSigParent(_, _)
             | Condition::AggSigParentAmount(_, _)
-            | Condition::AggSigParentPuzzle(_, _) => {
+            | Condition::AggSigParentPuzzle(_, _)
+            | Condition::AggSigAmount(_, _)
+            | Condition::AggSigPuzzleAmount(_, _)
+            | Condition::AssertHeightRelative(_)
+            | Condition::AssertSecondsRelative(_) => {
```
arvidn, line 118: These two should not clear the `ELIGIBLE_FOR_DEDUP` flag, should they?

I believe you should also check `AssertEphemeral` here. It's considered a relative-timelock.

Also `ASSERT_BEFORE_SECONDS_RELATIVE`, `ASSERT_BEFORE_HEIGHT_RELATIVE`, `ASSERT_MY_BIRTH_HEIGHT` , `ASSERT_MY_BIRTH_SECONDS`. I think that's it

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -1860,7 +1873,7 @@ fn test_invalid_condition_args_terminator() {
     let spend = &conds.spends[0];
     assert_eq!(*spend.coin_id, test_coin_id(H1, H2, 123));
     assert_eq!(a.atom(spend.puzzle_hash).as_ref(), H2);
-    assert_eq!(spend.flags, ELIGIBLE_FOR_DEDUP | HAS_RELATIVE_CONDITION);
```
arvidn, line 1863: I don't think your change is supposed to affect the DEDUP flag, right?

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -2116,9 +2129,15 @@ fn test_extra_arg(
     ]
     .contains(&condition);
 
+    let banned_ff = [ASSERT_SECONDS_RELATIVE, ASSERT_HEIGHT_RELATIVE].contains(&condition);
```
arvidn, line 2132: I think this list needs to be longer

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -2168,14 +2187,15 @@ fn test_single_condition(
...
     assert_eq!(conds.cost, 0);
     assert_eq!(conds.spends.len(), 1);
     let spend = &conds.spends[0];
     assert_eq!(*spend.coin_id, test_coin_id(H1, H2, 123));
     assert_eq!(a.atom(spend.puzzle_hash).as_ref(), H2);
-    assert!((spend.flags & ELIGIBLE_FOR_DEDUP) != 0);
-
+    if !banned_ff {
+        assert!((spend.flags & ELIGIBLE_FOR_DEDUP) != 0);
```
arvidn, line 2197: I don't think your patch *should* affect the `DEDUP` flag

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -4070,7 +4091,8 @@ fn test_impossible_constraints_single_spend(
     } else {
         // we don't expect any error
         let (a, conds) = cond_test(test).unwrap();
-
+        let banned_list = [ASSERT_SECONDS_RELATIVE, ASSERT_HEIGHT_RELATIVE];
```
arvidn, line 4094: I think this list needs to be longer

arvidn (changes requested):

crates/chia-consensus/src/gen/conditions.rs
matt-o-how: Ephemeral and Before conditions are not affected by the fastforward - good catch with the birth height and birth seconds though, I've added those now

crates/chia-consensus/src/gen/conditions.rs
matt-o-how: Correct, I've reverted this

crates/chia-consensus/src/gen/conditions.rs
matt-o-how: I've reverted this

crates/chia-consensus/src/gen/conditions.rs
matt-o-how: I've reverted this

arvidn (commented):
do we already update the solution with the new coin amount when we fast-forward?

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -99,14 +119,14 @@ impl SpendVisitor for MempoolVisitor {
             Condition::AggSigMe(_, _)
             | Condition::AggSigParent(_, _)
             | Condition::AggSigParentAmount(_, _)
-            | Condition::AggSigParentPuzzle(_, _) => {
+            | Condition::AggSigParentPuzzle(_, _)
```
arvidn, line 122: Can you think of why we restrict committing to the parent puzzle? it's supposed to be unchanged anyway, right?

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -99,14 +119,14 @@ impl SpendVisitor for MempoolVisitor {
...
                 spend.flags &= !ELIGIBLE_FOR_FF;
             }
-            Condition::AggSigPuzzle(_, _)
-            | Condition::AggSigAmount(_, _)
-            | Condition::AggSigPuzzleAmount(_, _)
-            | Condition::AggSigUnsafe(_, _) => {
+            Condition::AggSigPuzzle(_, _) | Condition::AggSigUnsafe(_, _) => {
                 spend.flags &= !ELIGIBLE_FOR_DEDUP;
             }
             Condition::SendMessage(src_mode, _dst, _msg) => {
```
arvidn, line 132: I think you need to restrict sent and received messages that commit to coin amount as well

matt-o-how: Done

crates/chia-consensus/src/gen/conditions.rs
matt-o-how: That's our Parent ID and our Puzzle, not our parent's puzzle. Ambiguous name.

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -48,6 +48,7 @@ pub const HAS_RELATIVE_CONDITION: u32 = 2;
 // 3. No ASSERT_MY_COIN_ID condition, no more than one ASSERT_MY_PARENT_ID condition
 //    (as the second condition)
 // 4. it has an output coin with the same puzzle hash as the spend itself
+// 5. there are no ASSERT_MY_COIN_AMOUNT or AGGSIG_*_AMOUNT conditions
```
arvidn, line 51: good idea to amend this comment. you should add the other new restrictions too though. with sending/receiving messages and `ASSERT_BIRTH_*`, `ASSERT_EPHEMERAL` etc.

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -82,9 +83,31 @@ impl SpendVisitor for MempoolVisitor {
...
         match c {
-            Condition::AssertMyCoinId(_) => {
+            Condition::AssertMyCoinId(_)
+            | Condition::AssertHeightRelative(_)
+            | Condition::AssertSecondsRelative(_)
+            | Condition::AssertBeforeHeightRelative(_)
+            | Condition::AssertBeforeSecondsRelative(_)
+            | Condition::AssertMyBirthHeight(_)
+            | Condition::AssertMyBirthSeconds(_)
+            | Condition::AssertEphemeral => {
```
arvidn, line 93: `ASSERT_EPHEMERAL`, `ASSERT_BEFORE_SECONDS_RELATIVE` and `ASSERT_BEFORE_HEIGHT_RELATIVE` should have tests too. you could try to comment out one at a time and make sure the tests fail

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -4636,6 +4660,11 @@ fn test_eligible_for_ff_output_coin(#[case] amount: u64, #[case] ph: &str, #[cas
 #[rstest]
 #[case(ASSERT_MY_PARENT_ID, "{h1}")]
 #[case(ASSERT_MY_COIN_ID, "{coin12}")]
+#[case(ASSERT_MY_AMOUNT, "123")]
```
arvidn, line 4663: it's not obvious why this fails. we allow this condition as the first one, right?
I have a feeling that maybe this test isn't testing what we think it is.

crates/chia-consensus/src/gen/conditions.rs
matt-o-how: changed this comment to reflect timelocks

crates/chia-consensus/src/gen/conditions.rs
matt-o-how: this test was not what we thought it was - it was actually testing conditions order

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -4760,6 +4792,44 @@ fn test_eligible_for_ff_invalid_assert_parent(
...
 
+#[cfg(test)]
+#[rstest]
+#[case(ASSERT_HEIGHT_RELATIVE, "0", false)]
+#[case(ASSERT_SECONDS_RELATIVE, "0", false)]
+#[case(ASSERT_BEFORE_HEIGHT_RELATIVE, "0", false)]
+#[case(ASSERT_BEFORE_SECONDS_RELATIVE, "0", false)]
+#[case(ASSERT_MY_BIRTH_HEIGHT, "0", false)]
+#[case(ASSERT_MY_BIRTH_SECONDS, "0", false)]
+#[case(ASSERT_MY_AMOUNT, "123", true)]
```
arvidn, line 4803: I think it would be good to test other conditions here too

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -4737,6 +4768,7 @@ fn test_eligible_for_ff_output_coin(#[case] amount: u64, #[case] ph: &str, #[cas
 #[rstest]
 #[case(ASSERT_MY_PARENT_ID, "{h1}")]
 #[case(ASSERT_MY_COIN_ID, "{coin12}")]
+
```
arvidn, line 4771: not a big deal, but these added and removed empty lines add noise to the review

crates/chia-consensus/src/gen/conditions.rs
arvidn: especially `ASSERT_EPHEMERAL` is missing

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -4768,8 +4838,6 @@ fn test_eligible_for_ff_invalid_assert_parent(
 #[case(AGG_SIG_PARENT_PUZZLE, false)]
 #[case(AGG_SIG_UNSAFE, true)]
 #[case(AGG_SIG_PUZZLE, true)]
-#[case(AGG_SIG_AMOUNT, true)]
-#[case(AGG_SIG_PUZZLE_AMOUNT, true)]
```
arvidn, line 4772: why were these removed?

matt-o-how: re-added

arvidn (approved):

Rigidity (approved):

PR #951, opened 2025-02-26
Handle RESERVE_FEE differently with identical spend aggregation

Rigidity:


crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -1540,6 +1542,33 @@ pub fn validate_signature(
     Ok(())
 }
 
+pub fn check_deduplication(ret: &mut SpendBundleConditions) {
+    let mut non_dedup_removal_amount = 0;
+    let mut non_dedup_addition_amount = 0;
```
arvidn, line 1547: it would make me a bit less nervous to specify these types as `u64`

arvidn (commented):
I think this core logic is correct. We would need tests and I think it would be nice to avoid having to do this in consensus mode (where the `SpendVisitor` is `EmptyVisitor`)

arvidn:
this can be closed now, right?

PR #959, opened 2025-02-27
Revoke FF eligibility for coins referenced by ASSERT_CONCURRENT_SPEND

Rigidity:


crates/chia-consensus/src/gen/run_block_generator.rs
```diff
@@ -227,7 +227,7 @@ where
         return Err(ValidationErr(all_spends, ErrorCode::GeneratorRuntimeError));
     }
 
-    validate_conditions(a, &ret, &state, a.nil(), flags)?;
+    validate_conditions(a, &ret, &mut state, a.nil(), flags)?;
```
arvidn, line 230: we don't need this anymore

arvidn (commented):
looks good. but needs a few tests

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -5636,3 +5646,26 @@ fn test_message_eligible_for_ff() {
...
+           ((64 ({} ) \
+           ((86 (1000 ) \
+           ))\
+       ))",
+        if is_dedup_id { "{coin12}" } else { "{coin21}" }
+    );
+
+    let (_a, cond) = cond_test_flag(&test, 0).expect("cond_test");
+    assert!(cond.spends.len() == 2);
+    assert_ne!((cond.spends[0].flags & ELIGIBLE_FOR_FF), is_dedup_id as u32);
```
arvidn, line 5669: this test looks suspicious. it's implying that `ELIGIBIBLE_FOR_FF` is `1` and that casting a `bool` to `u32` is `0` or `1`. But that's a lot of assumptions that aren't obvious by just reading this line. If we ever change around the flags to make `ELIGIBLE_FOR_FF` be, say, `2`, this test would always succeed, even if the flag isn't set.
```suggestion
    assert_ne!((cond.spends[0].flags & ELIGIBLE_FOR_FF) as bool, is_dedup_id);
```

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -5636,3 +5646,26 @@ fn test_message_eligible_for_ff() {
         assert_eq!((cond.spends[1].flags & ELIGIBLE_FOR_FF), 0);
     }
 }
+
+#[cfg(test)]
+#[rstest]
+fn test_assert_concurrent_spend_ff(#[values(true, false)] is_dedup_id: bool) {
+    let test = format!(
```
arvidn, line 5653: I think it would be nice to add a comment translating the condition codes used in the test. 51, 73, 64 and 86

arvidn (commented):
looks good apart from a few comments

arvidn (approved):

PR #960, opened 2025-02-27
Prevent dedup with excess amount

Rigidity:
For a coin to be eligible for dedup it must reserve more value than its amount. `CREATE_COIN` and `RESERVE_FEE` are ways of reserving value.

### DIFF
--- a/crates/chia-consensus/src/gen/conditions.rs
+++ b/crates/chia-consensus/src/gen/conditions.rs
@@ -34,8 +34,11 @@ use std::sync::Arc;
 
 // spend flags
 
-// a spend is eligible for deduplication if it does not have any AGG_SIG_ME
-// nor AGG_SIG_UNSAFE
+// If the CoinSpend is eligible for deduplication, this flag is set. A spend is
+// eligible if:
+// 1. There are no AGG_SIG_* conditions
+// 2. There are no message conditions
+// 3. The coin amount is less than or equal to its own outputs (to prevent paying for something multiple times)
 pub const ELIGIBLE_FOR_DEDUP: u32 = 1;
 
 // If the spend bundle contained *any* relative seconds or height condition, this flag is set
@@ -167,6 +170,21 @@ impl SpendVisitor for MempoolVisitor {
             })
         {
             spend.flags &= !ELIGIBLE_FOR_FF;
+        }
+
+        // A spend cannot be eligible for dedup if it has an excess amount
+        if (spend.flags & ELIGIBLE_FOR_DEDUP) != 0 {
+            let spend_additions = spend
+                .create_coin
+                .iter()
+                .map(|c| c.amount as u128)
+                .sum::<u128>();
+
+            let spend_commitments = spend_additions + spend.reserve_fee as u128;
+
+            if spend.coin_amount as u128 > spend_commitments {
+                spend.flags &= !ELIGIBLE_FOR_DEDUP;
+            }
         }
     }
 
@@ -726,6 +744,7 @@ pub struct SpendConditions {
     pub agg_sig_puzzle_amount: Vec<(PublicKey, NodePtr)>,
     pub agg_sig_parent_amount: Vec<(PublicKey, NodePtr)>,
     pub agg_sig_parent_puzzle: Vec<(PublicKey, NodePtr)>,
+    pub reserve_fee: u64,
     // Flags describing properties of this spend. See flags above
     pub flags: u32,
 }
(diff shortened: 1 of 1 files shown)


crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -2150,8 +2160,11 @@ fn test_extra_arg(
 
     let (a, conds) = cond_test_sig(
         &format!(
-            "((({{h1}} ({{h2}} (123 ((({} ({} ( 1337 ) {} ))))",
-            condition as u8, arg, extra_cond
```
arvidn, line 2154: this test isn't really about testing DEDUP. I think it would be fine to just change the expectation of the `ELIGIBLE_FOR_DEDUP` flag being set.

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -2362,11 +2377,12 @@ fn test_multiple_conditions(
 ) {
     let val = condition as u8;
     let (a, conds) = cond_test(&format!(
-        "((({{h1}} ({{h2}} (1234 ((({val} (100 ) (({val} (503 ) (({val} (90 )))))"
+        "((({{h1}} ({{h2}} (1234 (((51 ({{h2}} ({} ) (({val} (100 ) (({val} (503 ) (({val} (90 )))))",
+        if condition == RESERVE_FEE { 541 } else { 1234 }
```
arvidn, line 2381: This passed previously because we didn't create a coin, right? It's not obvious that making this change is simpler than to just not expect `ELIGIBLE_FOR_DEDUP` being set

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -5636,3 +5655,59 @@ fn test_message_eligible_for_ff() {
...
+    match output {
+        DedupCoinOutput::Absorb | DedupCoinOutput::Equal => {
+            assert_eq!(cond.spends[0].flags & ELIGIBLE_FOR_DEDUP, 1);
+        }
+        DedupCoinOutput::Excess => {
+            assert_eq!(cond.spends[0].flags & ELIGIBLE_FOR_DEDUP, 0);
+        }
+    }
+
+    assert_eq!(cond.spends[1].flags & ELIGIBLE_FOR_DEDUP, 0);
```
arvidn, line 5712: I think it would be good to comment why spend 1 is never dedup and why spend 0 is sometimes and sometimes not

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -5636,3 +5655,59 @@ fn test_message_eligible_for_ff() {
...
+           ))\
+       ))"
+    );
+
+    let (_a, cond) = cond_test_flag(&test, DONT_VALIDATE_SIGNATURE).expect("cond_test");
+    assert!(cond.spends.len() == 2);
+
+    match output {
+        DedupCoinOutput::Absorb | DedupCoinOutput::Equal => {
+            assert_eq!(cond.spends[0].flags & ELIGIBLE_FOR_DEDUP, 1);
```
arvidn, line 5705: I'm not excited about making this check depend on the actual flag value of `ELIGIBLE_FOR_DEDUP`

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -5636,3 +5655,59 @@ fn test_message_eligible_for_ff() {
...
+        DedupCoinOutput::Excess => 122,
+        DedupCoinOutput::Equal => 123,
+        DedupCoinOutput::Absorb => 124,
+    };
+
+    let coin2_output = match output {
+        DedupCoinOutput::Excess => 124,
+        DedupCoinOutput::Equal => 123,
+        DedupCoinOutput::Absorb => 122,
+    };
```
arvidn, line 5686: It's not obvious to me why you can't just hard-code the coin created by this spend, or just not have it create any coins at all. It's spend 0 that's being tested, right? And the only reason to have a second coin is to avoid the `MINTING_COIN` error, as I understand it.

arvidn (commented):
I think the tests could be simplified a little bit, but the core logic looks solid.

arvidn (approved):
nice! Although, in my mind, the concrete integer value for the flag `ELIGIBLE_FOR_DEDUP` should be an implementation detail. your tests pin it down to be 1. Ideally the flags type and the flags should be a custom type that doesn't define conversion to integer.

arvidn (approved):

PR #968, opened 2025-03-03
Unpin rust-toolchain.toml temporarily to resolve CI issue

Rigidity:
Until we can figure out why `1.85.0` isn't considered the same as `stable` by Rustup


arvidn (approved):
looks reasonable. it worries me a bit that this defeats some of the purpose of having a toolchain file

arvidn:
I think search-and-replacing `stable` for the concrete version we use, in the workflow files

PR #985, opened 2025-03-07
update ring to 0.17.13

altendky:


Rigidity:
This is resolved

altendky:
i did check before making this...  i think...  but, yeah, would rather take the dependabot pr given it was present.  thanks.

PR #1039, opened 2025-05-02
CHIA-2826 validate puzzle hash in run_spendbundle()

arvidn:
to ensure the puzzle reveal matches the puzzle hash of the coin being spent


crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -75,6 +76,9 @@ pub fn run_spendbundle(
         subtract_cost(a, &mut cost_left, clvm_cost)?;
 
         let buf = tree_hash(a, puz);
+        if coin_spend.coin.puzzle_hash != buf.into() {
+            return Err(ValidationErr(puz, ErrorCode::WrongPuzzleHash));
```
matt-o-how, line 80: Perhaps we should have a test that explicitly fails this check?

arvidn: added. I refactored the tests some as well, to make them shorter (and easier to read, I think). They are best reviewed one commit at a time

matt-o-how (approved):

PR #1051, opened 2025-05-08
debug failures on beast

altendky:


.github/workflows/benchmark.yml
```diff
@@ -113,6 +113,11 @@ jobs:
         python-version: ["3.10", "3.11"]
 
     steps:
+      - run: |
+          ls -la .cargo/bin/ || true
+          echo CARGO_HOME=${CARGO_HOME} || true
+          ls -la "${CARGO_HOME}"/bin/ || true
```
arvidn, line 119: do you think we should keep these in?

PR #1052, opened 2025-05-09
add fields to ConsensusConstants to hang new hard fork features on

arvidn:
The idea of renaming `number_zero_bits_plot_filter` to have the `_old` suffix is to make sure that we review all places in chia-blockchain where this constant is used, to make sure we support the new plot format


matt-o-how (approved):
I like the idea of forcing a review of relevant code by breaking compatibility

Rigidity (approved):

PR #1053, opened 2025-05-09
Add a cost to all conditions and a new "free conditions" counter

matt-o-how:
This PR adds a new flag called COST_CONDITIONS which is activated after hard fork 2 which adds a cost to all conditions after the hard fork. There are a number of "free conditions" for every spend, which is stored as constant in opcodes set to 100.


arvidn (changes requested):
you need tests for this as well. positive tests, edge cases and negative tests

crates/chia-consensus/src/gen/opcodes.rs
```diff
@@ -64,6 +64,10 @@ pub const SOFTFORK: ConditionOpcode = 90;
 
 pub const CREATE_COIN_COST: Cost = 1_800_000;
 pub const AGG_SIG_COST: Cost = 1_200_000;
+
+pub const GENERIC_CONDITION_COST: Cost = 1_200_000;
```
arvidn, line 68: I think this should be in the range 100-500.

crates/chia-consensus/src/gen/test_generators.rs
```diff
@@ -2,7 +2,7 @@ use super::conditions::{NewCoin, SpendBundleConditions, SpendConditions};
 use super::run_block_generator::{run_block_generator, run_block_generator2};
 use crate::allocator::make_allocator;
 use crate::consensus_constants::TEST_CONSTANTS;
-use crate::gen::flags::{DONT_VALIDATE_SIGNATURE, MEMPOOL_MODE};
+use crate::r#gen::flags::{DONT_VALIDATE_SIGNATURE, MEMPOOL_MODE};
```
arvidn, line 5: what does this do?

crates/chia-consensus/src/spendbundle_validation.rs
```diff
@@ -2,6 +2,7 @@ use crate::allocator::make_allocator;
 use crate::consensus_constants::ConsensusConstants;
 use crate::gen::owned_conditions::OwnedSpendBundleConditions;
 use crate::gen::validation_error::ErrorCode;
+use crate::r#gen::flags::COST_CONDITIONS;
```
arvidn, line 5: if there's some benefit to `r#`, I imagine we should add it to every `use`, but probably not in this PR

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -4668,50 +4744,104 @@ fn test_softfork_condition_failures(#[case] conditions: &str, #[case] expected_e
...
+    ASSERT_CONCURRENT_PUZZLE,
+    1025,
+    COST_CONDITIONS,
     Some(ErrorCode::AssertConcurrentPuzzleFailed)
 )]
-#[case(ASSERT_CONCURRENT_PUZZLE, 1025, Some(ErrorCode::TooManyAnnouncements))]
 fn test_limit_announcements(
     #[case] cond: ConditionOpcode,
     #[case] count: i32,
+    #[case] flag: u32,
```
arvidn, line 4839: in `rstest` you can also add an independent parameter by saying:
```
#[values(0, COST_CONDITIONS)]
```
Then every test is run in both configuration. You could check the result programatically.

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -1211,27 +1227,39 @@ pub fn parse_conditions<V: SpendVisitor>(
...
             Condition::CreateCoinAnnouncement(msg) => {
-                decrement(&mut announce_countdown, msg)?;
+                if (flags & COST_CONDITIONS) == 0 {
+                    decrement(&mut announce_countdown, msg)?;
+                };
                 state.announce_coin.insert((spend.coin_id.clone(), msg));
             }
             Condition::CreatePuzzleAnnouncement(msg) => {
-                decrement(&mut announce_countdown, msg)?;
+                if flags & COST_CONDITIONS != COST_CONDITIONS {
```
arvidn, line 1236:
```suggestion
                if flags & COST_CONDITIONS == 0 {
```
Perhaps this is a matter of taste. I prefer this one, and right now it's not consistent, which looks a bit odd.

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -1039,7 +1043,19 @@ pub fn parse_conditions<V: SpendVisitor>(
                 *max_cost -= AGG_SIG_COST;
                 ret.condition_cost += AGG_SIG_COST;
             }
-            _ => (),
+            _ => {
+                if (flags & COST_CONDITIONS) != 0 {
```
arvidn, line 1047: maybe it was more reasonable the way you had it before, where it always applied to all conditions. I suppose we'll have to update a lot of tests that way. But the `SOFTFORK` condition causes complications with this

crates/chia-consensus/src/gen/opcodes.rs
matt-o-how: Set to 500

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -1493,7 +1524,7 @@ pub fn validate_conditions(
     }
 
     if !state.assert_concurrent_puzzle.is_empty() {
-        let mut spent_phs = HashSet::<Bytes32>::with_capacity(state.spent_puzzles.len());
```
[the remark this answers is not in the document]
matt-o-how: reverted

crates/chia-consensus/src/gen/conditions.rs
matt-o-how: standardised

crates/chia-consensus/src/gen/conditions.rs
matt-o-how: implemented

crates/chia-consensus/src/spendbundle_validation.rs
arvidn: I think you should remove these

crates/chia-consensus/src/gen/conditions.rs
```diff
@@ -4668,50 +4844,104 @@ fn test_softfork_condition_failures(#[case] conditions: &str, #[case] expected_e
...
+    ASSERT_CONCURRENT_PUZZLE,
+    1025,
+    COST_CONDITIONS,
     Some(ErrorCode::AssertConcurrentPuzzleFailed)
 )]
-#[case(ASSERT_CONCURRENT_PUZZLE, 1025, Some(ErrorCode::TooManyAnnouncements))]
 fn test_limit_announcements(
     #[case] cond: ConditionOpcode,
     #[case] count: i32,
+    #[case] flag: u32,
```
arvidn, line 4939:
```suggestion
    #[values(0, COST_CONDITIONS)] flag: u32,
```

arvidn (approved):

PR #1055, opened 2025-05-13
update ProofOfSpace to support the new plot format

arvidn:
repurpose the most significant bit in the `size` field of `ProofOfSpace` to indicate version 1 or 2 of the plot format.

### DIFF
--- a/crates/chia-protocol/src/proof_of_space.rs
+++ b/crates/chia-protocol/src/proof_of_space.rs
@@ -8,6 +8,74 @@ pub struct ProofOfSpace {
     pool_public_key: Option<G1Element>,
     pool_contract_puzzle_hash: Option<Bytes32>,
     plot_public_key: G1Element,
-    size: u8,
+    version_and_size: u8,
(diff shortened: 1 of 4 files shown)


arvidn:
We could create two classes, called `PlotSizeV1` and `PlotSizeV2` and make the function return `Union[PlotSizeV1, PlotSizeV2]`. In python 3.10 we could then use `match` to switch on those types in a quite convenient way. However, I believe those types would have to be python classes, which would add quite a lot of complexity to the rust bindings,

almogdepaz (approved):

almogdepaz:
> We could create two classes, called `PlotSizeV1` and `PlotSizeV2` and make the function return `Union[PlotSizeV1, PlotSizeV2]`. In python 3.10 we could then use `match` to switch on those types in a quite convenient way. However, I believe those types would have to be python classes, which would add quite a lot of complexity to the rust bindings,

lets punt on this until we figure out the best way, refactoring later would not be a lot of work

PR #1061, opened 2025-05-15
Rename `chia_consensus::gen` to `chia_consensus::general`

matt-o-how:
`gen` is a future reserved keyword in Rust as of 2024
https://doc.rust-lang.org/reference/keywords.html#r-lex.keywords.reserved.edition2018

This causes some issues in editors already, and we should move away from using it to avoid problems in the future


arvidn (dismissed):
it looks like a few `r#gen` had snuck in already :)

arvidn (dismissed):

arvidn (approved):

Rigidity:
I suppose we should rename the PR :)

PR #1067, opened 2025-05-19
Reduce potential for panics in chia-bls

Rigidity:
Adds a fuzzer for `SecretKey::from_seed`, `SecretKey::from_bytes`, `PublicKey::from_bytes`, and `Signature::from_bytes`, to make sure they return errors instead of panics if provided invalid inputs.

`SecretKey::from_seed` now returns a `Result`, and therefore hardened key derivation does as well.

Added `anyhow` to tests to simplify `Result` handling, and enabled the `unwrap_used` lint in `chia-bls` specifically. Replaced all of the `.unwrap()` calls with either `?` or `expect` as appropriate, and disabled the lint for `unwrap_err` for now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -330,6 +330,7 @@ dependencies = [
 name = "chia-bls"
 version = "0.23.0"
 dependencies = [
+ "anyhow",
  "arbitrary",
  "blst",
  "chia-serde",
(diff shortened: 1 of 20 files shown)


Rigidity:
It's unclear this is the right path. Leaving open for discussion

PR #1069, opened 2025-05-20
[CHIA-2884] Pot iterations

arvidn:
* enable mypy on some test functions
* turn `signage_point_index` into a `u8`, like python.
* update python binding signatures to match the python implementation

### DIFF
--- a/crates/chia-protocol/src/pot_iterations.rs
+++ b/crates/chia-protocol/src/pot_iterations.rs
@@ -23,15 +23,13 @@ fn div_catch_error(a: u64, b: u64) -> Result<u64> {
 pub fn is_overflow_block(
     num_sps_sub_slot: u32,
     num_sp_intervals_extra: u8,
-    signage_point_index: u32,
+    signage_point_index: u8,
 ) -> Result<bool> {
-    if signage_point_index >= num_sps_sub_slot {
+    if u32::from(signage_point_index) >= num_sps_sub_slot {
         return Err(Error::InvalidPotIteration);
     }
-    Ok(
-        signage_point_index
-            >= sub_catch_underflow(num_sps_sub_slot, num_sp_intervals_extra as u32)?,
-    )
+    Ok(u32::from(signage_point_index)
+        >= sub_catch_underflow(num_sps_sub_slot, num_sp_intervals_extra as u32)?)
 }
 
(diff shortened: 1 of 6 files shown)


crates/chia-protocol/src/pot_iterations.rs
```diff
@@ -194,42 +130,45 @@ mod tests {
...
-            sp_interval_iters.try_into().unwrap(),
-            sp_interval_iters * 12
-        )
-        .is_err());
+        assert_eq!(
+            calculate_ip_iters(
+                NUM_SPS_SUB_SLOT,
+                NUM_SP_INTERVALS_EXTRA,
+                ssi,
+                13,
```
arvidn, line 160: `sp_interval_iters.try_into().unwrap(),` -> `13`

tests/test_pot_iterations.py
```diff
@@ -99,37 +92,33 @@ def test_calculate_ip_iters(self):
         with raises(ValueError):
             # required_iters too high
             calculate_ip_iters(
-                test_constants.NUM_SPS_SUB_SLOT,
-                test_constants.NUM_SP_INTERVALS_EXTRA,
+                test_constants,
                 ssi,
-                sp_interval_iters,
```
arvidn, line 105: these were failing. I think it was a mistake to pass in `sp_interval_iters` as `signage_point_index`. This value doesn't fit in a `u8`

crates/chia-protocol/src/pot_iterations.rs
```diff
@@ -23,15 +23,13 @@ fn div_catch_error(a: u64, b: u64) -> Result<u64> {
 pub fn is_overflow_block(
     num_sps_sub_slot: u32,
     num_sp_intervals_extra: u8,
-    signage_point_index: u32,
+    signage_point_index: u8,
```
almogdepaz, line 26: why is num_sps_sub_slot u32 ?

tests/test_pot_iterations.py
almogdepaz: it also doesn't make much sense, these are unrelated

almogdepaz (approved):

crates/chia-protocol/src/pot_iterations.rs
arvidn: I believe this is mis-typed in `ConsensusConstants`.
https://github.com/Chia-Network/chia-blockchain/blob/main/chia/consensus/default_constants.py#L17

PR #1071, opened 2025-05-21
Remove chia-blockchain from install instructions and fix pytest.ini

matt-o-how:
This removes `chia-blockchain` from the install instructions in the `README` and GitHub actions.
In its stead we install `blspy` which was the actual dependency that we were getting via chia-blockchain.

The largest change is removing the functions in Program which take or return chia-blockchain's Program.
I feel justified in this change as it is still trivial to convert between Program and Program, via `to_bytes()` and `from_bytes()` and the `chia-blockchain` Program will have `run` and `uncurry` - the `chia_rs` Program has no need for these functions, and removing them allows us to entirely remove the `chia-blockchain` dependency.

This PR also changes `-n auto` from `pytest.ini` which was causing problems with running.


matt-o-how:
If we no longer need `test_blspy_fidelity` then we don't need `blspy` as a dependency either

crates/chia-protocol/src/program.rs
```diff
@@ -407,15 +415,7 @@ impl Program {
...
-        let mut a = Allocator::new_limited(500_000_000);
-        let prg = node_from_bytes_backrefs(&mut a, self.0.as_ref())?;
-        let prg = LazyNode::new(Rc::new(a), prg);
-        to_program(py, prg)
-    }
-
-    fn uncurry<'a>(&self, py: Python<'a>) -> PyResult<(Bound<'a, PyAny>, Bound<'a, PyAny>)> {
+    fn uncurry_to_lazynode(&self) -> PyResult<(LazyNode, LazyNode)> {
         use clvm_utils::CurriedProgram;
         use std::rc::Rc;
```
arvidn, line 420: If you're hoisting the `use` statements out to the file-scope, I think you should remove this one as well

tests/test_program.py
```diff
@@ -14,18 +14,6 @@ def test_raise() -> None:
         assert f"{e}" == "('clvm raise', '86666f6f626172')"
 
 
-def test_raise_program() -> None:
```
arvidn, line 17: I would expect this test to still work and be relevant. we're not looking at the return value even

matt-o-how: re-added now with the new run func

crates/chia-protocol/src/program.rs
matt-o-how: Done

pytest.ini
```diff
@@ -1,2 +1,2 @@
 [pytest]
-addopts = --verbose --tb=short -n auto
```
arvidn, line 2: would it be possible to instead add a python dev-dependency on `pytest-xdist`? These tests can take a while to run, not running them in parallel would make it a lot slower

arvidn (commented):
looks good, apart from pytest-xdist I think is really helpful

pytest.ini
matt-o-how: reverted the `-n auto` change

arvidn (approved):

PR #1082, opened 2025-05-25
Make offer memos more flexible

Rigidity:


arvidn (commented):
looks good. just two questions

PR #1083, opened 2025-05-25
Relax the default encoding of list in clvm-derive

Rigidity:


crates/clvm-derive/src/from_clvm.rs
```diff
@@ -151,8 +151,8 @@ fn check_rest_value(crate_name: &Ident, repr: Repr) -> TokenStream {
     match repr {
         Repr::Atom | Repr::Transparent => unreachable!(),
         // We don't need to check the terminator of a solution.
```
arvidn, line 153: I think it would be good to update this comment, to explain that consensus rules don't require lists to be nil terminated, and that we just ignore any additional elements we don't expect.

crates/clvm-traits/docs/derive_macros.md
```diff
@@ -12,9 +12,9 @@ Pick whichever representation fits your use-case the best.
 Note that the syntax `(A . B)` represents a cons-pair with two values, `A` and `B`.
 This is how non-atomic values are structured in CLVM.
 
-### List
+### Proper List
 
-This represents values in a nil terminated series of nested cons-pairs, also known as a proper list.
+This represents values in a nil terminated series of nested cons-pairs.
```
arvidn, line 17: we also don't allow additional elements in a proper list, right?

crates/clvm-traits/docs/derive_macros.md
```diff
@@ -50,11 +50,11 @@ let ptr = value.to_clvm(a).unwrap();
...
 
-### Solution
+### List
 
-The solution representation is the same as list, except it does not check the nil terminator when parsing.
-This allows it to be lenient to additional parameters that are in the CLVM object, since they don't affect anything.
-If you want your solution to be parsed strictly, you can use list instead.
+The list representation is the same as proper list, except it does not check the nil terminator when parsing.
+This allows it to be lenient to additional parameters that are in the CLVM object, which is especially useful
+for preventing errors when parsing on-chain solutions and output conditions, where you wouldn't want strictness.
```
arvidn, line 57: it might be good to explain that consensus rules are relaxed and the reason is to leave the door open for future soft-forks

arvidn (commented):
excellent! I just had a few minor suggestions

PR #1091, opened 2025-05-28
Add new stress tests for the generator based on the new costing

matt-o-how:


tests/test-generators.py
```diff
@@ -58,21 +59,27 @@ def validate_except_cost(output1: str, output2: str):
 
 
 print(f"{'test name':43s}   consensus | mempool")
-for g in sorted(glob.glob("../generator-tests/*.txt")):
+test_list = sorted(glob.glob("../generator-tests/*.txt"))
```
arvidn, line 62: an alternative would be to make this path relative to the current python file

tests/test-generators.py
```diff
@@ -130,9 +147,17 @@ def validate_except_cost(output1: str, output2: str):
         elif "recursion-pairs.txt" in g:
             limit = 4
             strict_limit = 4
-
-        compare_output(consensus.output, expected, "")
-        compare_output(mempool.output, expected_mempool, "STRICT")
+        if (
+            "aa-million-messages.txt" not in g
+            and "aa-four-million-messages.txt" not in g
```
arvidn, line 152: it seems like this would be simpler if it matched the same pattern of the other exceptions

PR #1097, opened 2025-05-30
fix mypy issues and enable more checks on CI

arvidn:


tests/test_spend_bundle.py
```diff
@@ -241,4 +241,4 @@ def test_derive_class():
     assert test.test_bool
     assert test.new_function()
 
-    test = NewAndImprovedSpendBundle.parse_rust(obj_bytes)
+    test = NewAndImprovedSpendBundle.parse_rust(obj_bytes)[0]
```
matt-o-how, line 244: perhaps this line should be `assert NewAndImprovedSpendBundle.parse_rust(obj_bytes)[0] is not None` ?

arvidn: `parse_rust()` doesn't return an `Optional[]` though, so that would be a type error

.github/workflows/benchmark.yml
```diff
@@ -71,7 +71,7 @@ jobs:
 
       - name: python mypy
         run: |
-          mypy tests
+          mypy tests --check-untyped-defs
```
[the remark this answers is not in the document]
arvidn: that's a good idea

AmineKhaldi (approved):

AmineKhaldi (approved):

PR #1103, opened 2025-06-03
Remove hard coded secret keys

Rigidity:


arvidn:
It's not obvious what the upside is. Is it just to mitigate bogus hacker-one reports?

arvidn:
I think it would be better and simpler to add a comment that the keys are only meant for tests

PR #1115, opened 2025-06-12
add missing test return hint

altendky:


arvidn (approved):
I would have expected there to be a corresponding removal of mypy exemption, to "lock in" this fix

fchirica (approved):

altendky:
> oh, this is chia_rs, isn't this checked by mypy already?

I haven't caught up with the stricter checking that has recently been applied in main.  Was (maybe) waiting for blockchain to catch up with 0.26.0.

PR #1121, opened 2025-06-20
Instead of collecting nodes with multiple references in deserialization

arvidn:
run an extra pass over puzzle reveals count references. Use these counters for memoizing tree hash instead.

### DIFF
--- a/crates/chia-consensus/src/additions_and_removals.rs
+++ b/crates/chia-consensus/src/additions_and_removals.rs
@@ -7,13 +7,12 @@ use crate::consensus_constants::ConsensusConstants;
 use crate::validation_error::{atom, first, next, rest, ErrorCode, ValidationErr};
 use chia_protocol::{Bytes, Bytes32};
 use clvm_traits::FromClvm;
-use clvm_utils::{tree_hash_cached, TreeHash};
+use clvm_utils::{tree_hash_cached, TreeCache};
 use clvmr::allocator::NodePtr;
 use clvmr::chia_dialect::ChiaDialect;
 use clvmr::reduction::Reduction;
 use clvmr::run_program::run_program;
-use clvmr::serde::node_from_bytes_backrefs_record;
-use std::collections::HashMap;
+use clvmr::serde::node_from_bytes_backrefs;
 
 /// Run a *trusted* block generator and return its additions and removals. This
 /// function does not validate the block, it is assumed to be valid.
@@ -34,7 +33,7 @@ where
 
     let mut cost_left = constants.max_block_cost_clvm;
 
-    let (program, backrefs) = node_from_bytes_backrefs_record(&mut a, program)?;
+    let program = node_from_bytes_backrefs(&mut a, program)?;
 
     let args = setup_generator_args(&mut a, block_refs)?;
     let dialect = ChiaDialect::new(flags);
@@ -45,7 +44,7 @@ where
     subtract_cost(&a, &mut cost_left, clvm_cost)?;
     all_spends = first(&a, all_spends)?;
 
-    let mut cache = HashMap::<NodePtr, TreeHash>::new();
+    let mut cache = TreeCache::default();
     // at this point all_spends is a list of:
     // (parent-coin-id puzzle-reveal amount solution . extra)
     // where extra may be nil, or additional extension data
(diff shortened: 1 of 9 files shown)


crates/clvm-utils/src/tree_hash.rs
```diff
@@ -61,6 +60,106 @@ impl Deref for TreeHash {
...
+    /// mark the node as being visited. Returns true if this is node has already
+    /// been visited. i.e. returns whether this node is a good candidate for
+    /// memoizing tree hash results.
+    fn visit(&mut self, n: NodePtr) -> bool {
+        if !matches!(n.object_type(), ObjectType::Pair) {
+            return false;
+        }
+        let idx = n.index() as usize;
+        if idx >= self.pairs.len() {
+            self.pairs.resize(idx + 1, NOT_VISITED);
```
Rigidity, line 124: Should this return `false` if `self.hashes.len() == SEEN_MULTIPLE as usize` like in `insert`?

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -61,6 +60,106 @@ impl Deref for TreeHash {
...
+
+    pub fn should_memoize(&mut self, n: NodePtr) -> bool {
+        if !matches!(n.object_type(), ObjectType::Pair) {
+            return false;
+        }
+        let idx = n.index() as usize;
+        if idx >= self.pairs.len() {
+            false
+        } else {
+            self.pairs[idx] <= SEEN_MULTIPLE
```
Rigidity, line 140: If we've already memoized the value, do we really need to do it again? Maybe this should be `==`

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -61,6 +60,106 @@ impl Deref for TreeHash {
...
+    pub fn visit_tree(&mut self, a: &Allocator, node: NodePtr) {
+        if self.visit(node) {
+            return;
+        }
+        let mut nodes = vec![node];
+        while let Some(n) = nodes.pop() {
+            let SExp::Pair(left, right) = a.sexp(n) else {
+                continue;
+            };
+            if matches!(a.sexp(left), SExp::Pair(..)) && !self.visit(left) {
```
Rigidity, line 153: The `matches!` is redundant, since `self.visit` will do this check as well

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -61,6 +60,106 @@ impl Deref for TreeHash {
...
+pub struct TreeCache {
+    hashes: Vec<TreeHash>,
+    // each entry is an index into hashes, or u16::MAX if the pair has not been
+    // cached
+    pairs: Vec<u16>,
+}
+
+const NOT_VISITED: u16 = u16::MAX;
+#[allow(dead_code)]
+const SEEN_ONCE: u16 = u16::MAX - 1;
```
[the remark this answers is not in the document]
arvidn: well, yeah, it was used in an earlier revision. I wanted to keep it just to make it clear what this value means. But maybe it's a net negative

crates/clvm-utils/src/tree_hash.rs
arvidn: good point, probably

arvidn: actually, it's the `hashes` Vec whose size is limited. `pairs` is capped by the `Allocator` limits.

arvidn: the way the cache is used is to first visit the tree, *then* to compute the tree hash. `insert()` is called during tree-hash computation. So, if we've  run the tree-hash, and the cache is full, you're saying there's no point in visiting any more nodes because nothing more will be cached anywy?

crates/chia-consensus/src/additions_and_removals.rs
```diff
@@ -38,19 +38,31 @@ where
...
-    while let Some((spend, tail)) = a.next(all_spends) {
-        all_spends = tail;
+    // first iterate over all puzzle reveals to find duplicate nodes, to know
+    // what to memoize during tree hash computations. This is managed by
+    // TreeCache
+    let mut iter = all_spends;
+    while let Some((spend, rest)) = a.next(iter) {
+        iter = rest;
+        let (_parent_id, (puzzle, _rest)) =
+            <(Bytes32, (NodePtr, NodePtr))>::from_clvm(&a, spend)
```
Rigidity, line 58: If you're going to throw away the `Bytes32` anyway, does it make sense to spend time validating it?

Rigidity (dismissed):

PR #1132, opened 2025-06-25
Add get_spends functions to API

matt-o-how:
By implementing these in Rust we can increase security and also allow for refactoring in `chia-blockchain` which removes  dependencies on the `clvm` repository.

The downside is that parsing clvm in Rust is still pretty gnarly looking - especially in combination with using pyo3 to return a Dict.
I expect (hope) that one day this code will be able to be replaced with something more beautiful, but to my knowledge this will have to do for now.


crates/chia-protocol/src/program.rs
```diff
@@ -68,6 +68,12 @@ impl Program {
         self.0.into_inner()
     }
 
+    pub fn get_tree_hash(&self) -> crate::Bytes32 {
```
arvidn, line 71: why did you move this?

tests/test_get_spends_for_block.py
```diff
@@ -0,0 +1,57 @@
+from chia_rs import get_spends_for_block_with_conditions
+from chia_rs.chia_rs import DONT_VALIDATE_SIGNATURE, get_spends_for_block
```
arvidn:
```suggestion
from chia_rs import DONT_VALIDATE_SIGNATURE, get_spends_for_block
```

wheel/generate_type_stubs.py
```diff
@@ -334,6 +334,20 @@ def get_conditions_from_spendbundle(
     height: int,
 ) -> SpendBundleConditions: ...
 
+def get_spends_for_block(
+    constants: ConsensusConstants,
+    generator: Program,
+    args: Program,
+    flags: uint32,
```
arvidn: we use `int` in other places, for flags

crates/chia-protocol/src/program.rs
matt-o-how: I made it public so I could use it in the new api function

wheel/src/api.rs
```diff
@@ -500,6 +501,123 @@ pub fn py_calculate_ip_iters(
...
+    py: Python<'_>,
+    constants: &ConsensusConstants,
+    generator: Program,
+    args: Program,
+    flags: u32,
+) -> pyo3::PyResult<PyObject> {
+    let mut a = make_allocator(LIMIT_HEAP);
+    let mut output = Vec::<CoinSpend>::new();
+    let (_, res) = generator
+        .run(&mut a, flags, constants.max_block_cost_clvm, &args)
```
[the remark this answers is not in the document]
arvidn: oh, maybe this function is expecting to get the args already prepared. I think it would be nice to change this to just pass in the array of block references (like `run_block_generator2())

wheel/src/api.rs
```diff
@@ -500,6 +501,123 @@ pub fn py_calculate_ip_iters(
...
+        // 117 = GeneratorRuntimeErrors
+        PyErr::new::<PyTypeError, _>(117)
+    })?;
+    let spends_list = Vec::<NodePtr>::from_clvm(&a, first).map_err(|_| {
+        // 117 = GeneratorRuntimeErrors
+        PyErr::new::<PyTypeError, _>(117)
+    })?;
+    for spend in spends_list {
+        let Ok(destructure_list!(parent_coin_info, puzzle, amount, solution)) =
+            <match_list!(BytesImpl<32>, Program, u64, Program)>::from_clvm(&a, spend)
```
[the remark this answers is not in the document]
matt-o-how: Done

wheel/src/api.rs
```diff
@@ -500,6 +501,123 @@ pub fn py_calculate_ip_iters(
...
+    args: Program,
+    flags: u32,
+) -> pyo3::PyResult<PyObject> {
+    let mut a = make_allocator(LIMIT_HEAP);
+    let mut output = Vec::<CoinSpend>::new();
+    let (_, res) = generator
+        .run(&mut a, flags, constants.max_block_cost_clvm, &args)
+        .map_err(|_| {
+            // 117 = GeneratorRuntimeErrors
+            PyErr::new::<PyTypeError, _>(117)
```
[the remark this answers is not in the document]
arvidn: This error type might be more appropriate, as it also has conversion to `PyErr`. `crates/chia-consensus/src/error.rs`

wheel/src/api.rs
```diff
@@ -539,52 +556,64 @@ pub fn get_spends_for_block(
...
         let Ok(destructure_list!(parent_coin_info, puzzle, amount, solution)) =
             <match_list!(BytesImpl<32>, Program, u64, Program)>::from_clvm(&a, spend)
         else {
-            continue; // if we fail at this step then maybe the puzzle was malicious - try other spends
+            continue; // if we fail at this step then maybe the generator was malicious - try other spends
         };
-        let puzhash = puzzle.get_tree_hash();
-        let coin = Coin::new(parent_coin_info, puzhash, amount);
+        let puz_ptr = node_from_bytes(&mut a, puzzle.as_slice())?;
+        let puzhash = tree_hash_cached(&a, puz_ptr, &mut cache);
```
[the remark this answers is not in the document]
matt-o-how: Done

wheel/src/api.rs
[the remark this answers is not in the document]
matt-o-how: Ah, done now

wheel/src/api.rs
```diff
@@ -539,52 +556,64 @@ pub fn get_spends_for_block(
...
+                .extract::<PyBuffer<u8>>()
+                .expect("block_refs must be list of buffers");
+            py_to_slice::<'a>(buf)
+        })
+        .collect::<Vec<&'a [u8]>>();
+    let args = setup_generator_args(&mut a, refs)?;
+    let dialect = ChiaDialect::new(flags);
+
+    let Reduction(clvm_cost, res) = run_program(&mut a, &dialect, program, args, cost_left)
+        .map_err(|_| ValidationErr(program, ErrorCode::GeneratorRuntimeError))?;
```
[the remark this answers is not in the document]
matt-o-how: Most of these have now changed into `continue` or `break`

wheel/src/api.rs
```diff
@@ -539,52 +556,64 @@ pub fn get_spends_for_block(
 }
 
 #[pyo3::pyfunction]
-pub fn get_spends_for_block_with_conditions(
-    py: Python<'_>,
+pub fn get_spends_for_block_with_conditions<'a>(
```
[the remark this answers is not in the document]
matt-o-how: the other function has moved to Rust but the interaction with PyBytes here means it's worth not separating out

Cargo.toml
arvidn, line 1: what happened here? this doesn't look right

crates/chia-consensus/src/run_block_generator.rs
```diff
@@ -244,3 +249,62 @@ where
     ret.cost = max_cost - cost_left;
     Ok(ret)
 }
+
+pub fn get_coinspends_for_trusted_block(
```
arvidn, line 253: do you think we could use the same test cases as we use for `run_block_generator2()` to unit test this function?

crates/chia-consensus/src/run_block_generator.rs
```diff
@@ -244,3 +249,62 @@ where
...
+    )?;
+
+    let (first, _rest) = a
+        .next(res)
+        .ok_or(ValidationErr(res, ErrorCode::GeneratorRuntimeError))?;
+    let mut cache = TreeCache::default();
+    let mut iter = first;
+    while let Some((spend, rest)) = a.next(iter) {
+        iter = rest;
+        let [_, puzzle, _] = extract_n::<3>(&a, spend, ErrorCode::InvalidCondition)?;
```
arvidn, line 281: we should probably be consistent with how we treat this error with line 287. If this fails, we'll abort. Which isn't necessarily unreasonable, it means the precondition was violated, because the block was invalid. But right now line 299 will silently skip the spend if it's invalid, so whatever we do, it would make sense to be consistent.

tests/test_get_spends_for_block.py
```diff
@@ -0,0 +1,69 @@
...
...fff04ffff02ff06ffff04ff02ffff04ff17ff80808080ff80808080ffff02ff17ff2f808080ff0180ffff04ffff01ff32ff02ffff03ffff07ff0580ffff01ff0bffff0102ffff02ff06ffff04ff02ffff04ff09ff80808080ffff02ff06ffff04ff02ffff04ff0dff8080808080ffff01ff0bffff0101ff058080ff0180ff018080ffff04ffff01b0a137bfa49e84c734e0ca689dbf186b98380c366ecd74011eee6840ff88b04476b327a2707852b83103e4faac53fbeb22ff018080), solution: Program(ff80ffff01ffff33ffa0d767c90667a593a8c585cd98ddf766a9d500168d84c01ad9b0dcdc349b4bbc0dff8500c556032980ffff3cffa06fb22bd61d8746104fd0a7168664b8a35c39d461f90ea4da354ba9519c43bd3f8080ff8080) }, 'conditions': [(50, [b'\\xa17\\xbf\\xa4\\x9e\\x84\\xc74\\xe0\\xcah\\x9d\\xbf\\x18k\\x988\\x0c6n\\xcdt\\x01\\x1e\\xeeh@\\xff\\x88\\xb0Dv\\xb3\\'\\xa2pxR\\xb81\\x03\\xe4\\xfa\\xacS\\xfb\\xeb\"', b'=\\x82{g\\xfco\\x90\\xb9\\xc0M\\xe7\\xcc\\xedc)\\x07T\\x8dY\\xec\\x05V\\n3\\xf3\\xcd\\x81\\xbf\\xe0\\xd8\\xbb\\xde']), (51, [b'\\xd7g\\xc9\\x06g\\xa5\\x93\\xa8\\xc5\\x85\\xcd\\x98\\xdd\\xf7f\\xa9\\xd5\\x00\\x16\\x8d\\x84\\xc0\\x1a\\xd9\\xb0\\xdc\\xdc4\\x9bK\\xbc\\r', b'\\x00\\xc5V\\x03)']), (60, [b'o\\xb2+\\xd6\\x1d\\x87F\\x10O\\xd0\\xa7\\x16\\x86d\\xb8\\xa3\\\\9\\xd4a\\xf9\\x0e\\xa4\\xda5K\\xa9Q\\x9cC\\xbd?'])]}]"
```
arvidn, line 67: this is a regression test, to fail if the output changes. It doesn't really prove that this output is *correct*, or consistent with `run_block_generator2()` (which I think it's fair to treat as the aithority on what's expected).

I think it would be good to parameterize a test on the test-generators and actually use the expected output embedded in those test cases to ensure we get something out that matches. We have tests in `crates/chia-consensus/src/test_generators.rs` which parameterize on these test cases, but it's rust and would only be feasible for the rust function.

wheel/src/api.rs
```diff
@@ -495,6 +498,164 @@ pub fn py_calculate_ip_iters(
...
+
+        // condition is each grouped list
+        // (51 0xcafef00d 100)
+        let mut num = 0;
+        let mut iter_two = res;
+        while let Some((condition, rest_two)) = a.next(iter_two) {
+            iter_two = rest_two;
+            let mut iter_three = condition;
+            let mut count: usize = 0;
+            let mut bytes_vec = Vec::<Py<PyBytes>>::new();
```
arvidn, line 612: not a big deal, but I think you could remove `count` and just use `bytes_vec.len()` instead. setting `count = 0` would have to be `bytes_vec.clear()` instead.

matt-o-how: ended up using `num` as the flag instead because of `REMARK` as a condition, but I have removed `count` now

tests/test_get_spends_for_block.py
arvidn, line 67: in this test, do you think it would be possible to generate this expected JSON string from the condition output that's part of this test case file? i.e. after the first new-line. You could parse that and turn it into JSON, probably. That would have two important improvements:
1. The reader of this test can get some confidence in the function being correct
2. You can run the test on multiple test cases

arvidn: oh, I see. this isn't JSON. I think the idea would hold still though.

matt-o-how: No, as the expected output does not include all conditions.
For example the file called `just-puzzle-anounce.txt` makes a puzzle announcement but has this info:
```
SPENDS:
- coin id: b0fd02f220f59cde5e968a7a0ab095cd2f7bc436e92a6693ce808a54ea0b6a7f ph: dd48779e97bee4e932705d33bf1adb8a7f3d5d5c73b1ab449ae32355fa62e445
cost: 13051378
removal_amount: 123
addition_amount: 0
```

crates/chia-consensus/src/run_block_generator.rs
matt-o-how: Done

Cargo.toml
```diff
@@ -124,11 +124,6 @@ clvm-traits = { path = "./crates/clvm-traits", version = "0.26.0" }
 clvm-utils = { path = "./crates/clvm-utils", version = "0.26.0" }
 clvm-derive = { path = "./crates/clvm-derive", version = "0.26.0" }
 chia-fuzz = { path = "./crates/chia-consensus/fuzz", version = "0.26.0" }
-chia-bls-fuzz = { path = "./crates/chia-bls/fuzz", version = "0.26.0" }
-chia-protocol-fuzz = { path = "./crates/chia-protocol/fuzz", version = "0.26.0" }
-chia-puzzle-types-fuzz = { path = "./crates/chia-puzzle-types/fuzz", version = "0.26.0" }
-clvm-traits-fuzz = { path = "./crates/clvm-traits/fuzz", version = "0.26.0" }
-clvm-utils-fuzz = { path = "./crates/clvm-utils/fuzz", version = "0.26.0" }
```
arvidn, line 131: I don't understand why these disappeared

crates/chia-consensus/fuzz/fuzz_targets/generator_python.rs
arvidn, line 1: why is this file called `_python`?

crates/chia-consensus/fuzz/fuzz_targets/generator_python.rs
```diff
@@ -0,0 +1,162 @@
...
+use chia_protocol::{Bytes, Coin, CoinSpend, Program, SpendBundle};
+use chia_traits::Streamable;
+use clvm_traits::ToClvm;
+use clvmr::{serde::node_to_bytes, Allocator, NodePtr};
+use libfuzzer_sys::fuzz_target;
+use std::io::{Cursor, Read};
+
+fuzz_target!(|data: &[u8]| {
+    let mut spends = Vec::<CoinSpend>::new();
+    let mut data = Cursor::new(data);
```
arvidn, line 16: instead of using `Cursor`, I think it would work much better to use `Unstructured` from the `Arbitrary` crate. See https://docs.rs/arbitrary/latest/arbitrary/struct.Unstructured.html

```suggestion
    let mut data = Unstructured::new(data);
```

crates/chia-consensus/src/test_generators.rs
```diff
@@ -291,18 +292,26 @@ fn run_generator(#[case] name: &str) {
...
-        for (i, spend) in conds.spends.into_iter().enumerate() {
-            let parent_id = a.atom(spend.parent_id);
-            assert_eq!(
-                parent_id.as_ref(),
-                coinspends[i].coin.parent_coin_info.as_slice()
-            );
-            let puzhash = a.atom(spend.puzzle_hash);
-            assert_eq!(puzhash.as_ref(), coinspends[i].coin.puzzle_hash.as_slice());
-            assert_eq!(spend.coin_amount, coinspends[i].coin.amount);
+        let result = get_coinspends_for_trusted_block(
```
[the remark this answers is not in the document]
arvidn: as coverall highlights, there's no unit test coverage for this function right now

crates/chia-consensus/src/run_block_generator.rs
```diff
@@ -310,3 +310,116 @@ pub fn get_coinspends_for_trusted_block(
...
     Ok(output)
 }
+
+#[allow(clippy::type_complexity)]
+pub fn get_coinspends_with_conditions_for_trusted_block(
+    constants: &ConsensusConstants,
+    generator: &Program,
+    refs: Vec<&[u8]>,
+    flags: u32,
+) -> Result<Vec<(CoinSpend, Vec<(u32, Vec<Vec<u8>>)>)>, ValidationErr> {
```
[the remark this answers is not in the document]
matt-o-how: Done

crates/chia-consensus/src/run_block_generator.rs
```diff
@@ -244,3 +249,179 @@ where
     ret.cost = max_cost - cost_left;
     Ok(ret)
 }
+
+pub fn get_coinspends_for_trusted_block(
+    constants: &ConsensusConstants,
+    generator: &Program,
+    refs: Vec<&[u8]>,
```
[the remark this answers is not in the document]
matt-o-how: Done

crates/chia-consensus/src/run_block_generator.rs
```diff
@@ -310,3 +310,116 @@ pub fn get_coinspends_for_trusted_block(
...
+            let mut bytes_vec = Vec::<Vec<u8>>::new();
+            while let Some((condition_values, rest_three)) = a.next(iter_three) {
+                iter_three = rest_three;
+                if num == 0 {
+                    // convert the first value to a small number which is the condition opcode
+                    // In our above examples: 51
+                    let Some(n) = a.small_number(condition_values) else {
+                        break;
+                    };
+                    num = n;
```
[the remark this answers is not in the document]
matt-o-how: Done

crates/chia-consensus/src/run_block_generator.rs
```diff
@@ -310,3 +310,116 @@ pub fn get_coinspends_for_trusted_block(
...
+            continue;
+        };
+        let Ok(solution_program) = Program::from_clvm(&a, solution) else {
+            continue;
+        };
+        let coinspend = CoinSpend::new(coin, puzzle_program.clone(), solution_program.clone());
+        let Ok((_, res)) =
+            puzzle_program.run(&mut a, flags, constants.max_block_cost_clvm, &solution)
+        else {
+            continue; // Skip this spend on error
```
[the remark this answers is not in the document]
matt-o-how: This now returns the coinspend without any conditions

PR #1135, opened 2025-07-03
add compute_puzzle_fingerprint() function

arvidn:
to identify identical solutions to dedup spends (without comparing exact bytes)


crates/chia-consensus/fuzz/fuzz_targets/puzzle-fingerprint.rs
```diff
@@ -0,0 +1,77 @@
...
+    ) else {
+        return;
+    };
+
+    assert_eq!(block_conds.spends.len(), 1);
+
+    // if the spend is not eligible for dedup, compute_puzzle_fingerprint()
+    // may or may not fail. It doesn't perform full validation
+    if (block_conds.spends[0].flags & ELIGIBLE_FOR_DEDUP) != 0 {
+        let Ok((cost, _fingerprint)) = fingerprint_result else {
```
Rigidity, line 67: We could use `expect()` here so we know what the error is

crates/chia-consensus/src/puzzle_fingerprint.rs
```diff
@@ -0,0 +1,412 @@
...
+            | RECEIVE_MESSAGE => {
+                return Err(ValidationErr(NodePtr::NIL, ErrorCode::InvalidCondition));
+            }
+            CREATE_COIN => {
+                cost += CREATE_COIN_COST;
+
+                // CREATE_COIN, puzzle_hash, amount
+                let rest = hash_atom_list(&mut fingerprint, &a, c, 3)?;
+
+                // make sure to include the hint if present. If it's not present
```
Rigidity, line 112: We should probably include all memos in the hash, since memos can be important to applications, right? Or is this a DoS vector?

crates/chia-consensus/src/puzzle_fingerprint.rs
```diff
@@ -0,0 +1,412 @@
...
+                // we insert an empty atom instead, to ensure CREATE_COIN always
+                // adds 4 atoms to the fingerprint
+                if let Ok(memos) = first(&a, rest) {
+                    if let Ok(hint) = first(&a, memos) {
+                        if let SExp::Atom = a.sexp(hint) {
```
Rigidity, line 117: `if let` chains just got released in Rust, but perhaps it's too soon to use them.

crates/chia-consensus/src/puzzle_fingerprint.rs
arvidn: how many memos are "all" of them? If it's infinite, then I think it's a DoS vector.
how deep is a memo tree allowed to be? If it's infinite, then I think it's also a DoS vector

crates/chia-consensus/src/puzzle_fingerprint.rs
arvidn: yeah, I can't wait..

crates/chia-consensus/src/puzzle_fingerprint.rs
arvidn: I think that sounds reasonable, at least as a first version

Rigidity (approved):

PR #1146, opened 2025-07-09
add more fuzzing for datalayer

altendky:


crates/chia-datalayer/src/merkle/iterators.rs
```diff
@@ -61,13 +63,39 @@ impl Iterator for LeftChildFirstIterator<'_> {
...
+            match block.node.parent().0 {
+                // TODO: maybe also check the parent considers this the child?
+                Some(index) => {
+                    if item.index != self.from_index && !self.already_queued.contains(&index) {
+                        return Some(Err(Error::ReferenceToUnknownParent()));
+                    }
+                }
+                None => {
+                    if item.index.0 != 0 {
+                        return Some(Err(Error::UnexpectedParentlessNode()));
```
arvidn, line 75: I don't really know what's going on, but I would expect this to happen when you get to the root. Is that not expected?

crates/chia-datalayer/src/merkle/iterators.rs
```diff
@@ -61,13 +63,39 @@ impl Iterator for LeftChildFirstIterator<'_> {
...
+                    }
+                }
+            }
+
             match block.node {
-                Node::Leaf(..) => return Some(Ok((item.index, block))),
+                Node::Leaf(..) => {
+                    if block.metadata.dirty {
```
arvidn, line 82: I got the impression that this would be calculated on-demand. It might be tricky to do while iterating over the tree immutably. There are `RefCell`s though.

crates/chia-datalayer/src/merkle/iterators.rs
altendky: presently the root is at index 0.  we have considered not requiring that, but have not made such a change.  so this checks if the parent is none and the item is not the root (at index 0).

crates/chia-datalayer/src/merkle/blob.rs
```diff
@@ -805,7 +816,14 @@ impl MerkleBlob {
         Ok(())
     }
 
-    pub fn check_integrity(&self) -> Result<(), Error> {
+    pub fn check_integrity(&mut self) -> Result<(), Error> {
+        self.check_just_integrity()?;
+        self.calculate_lazy_hashes()?;
```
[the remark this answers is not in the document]
altendky: https://github.com/Chia-Network/chia_rs/commit/287b694c3200eb00788eac2124e416f38281b14a

arvidn (approved):

PR #1150, opened 2025-07-10
enable allocator-debug in unit tests on CI

arvidn:
This will prevent `Allocator` related bugs going forward, at least if they're covered by a test.


Rigidity (approved):

arvidn:
I'm a bit reluctant to land this still. One of the stress tests use 5 GiB of RAM (down from 6.8 GiB before the optimization). I would like to land this and bump `clvm_rs` first probably: https://github.com/Chia-Network/clvm_rs/pull/600

PR #1156, opened 2025-07-16
Update to clvmrs 0.16.0 with new error codes

jack60612:
seems to be done

### DIFF
--- a/crates/chia-consensus/src/build_compressed_block.rs
+++ b/crates/chia-consensus/src/build_compressed_block.rs
@@ -1,10 +1,10 @@
 use crate::consensus_constants::ConsensusConstants;
+use crate::error::Result;
 use chia_bls::Signature;
 use chia_protocol::SpendBundle;
 use clvmr::allocator::{Allocator, NodePtr};
 use clvmr::serde::{node_from_bytes_backrefs, Serializer};
 use std::borrow::Borrow;
-use std::io;
 
 #[cfg(feature = "py-bindings")]
 use pyo3::prelude::*;
@@ -68,7 +68,7 @@ fn result(num_skipped: u32) -> BuildBlockResult {
 }
 
 impl BlockBuilder {
-    pub fn new() -> io::Result<Self> {
+    pub fn new() -> Result<Self> {
         let mut a = Allocator::new();
 
         // the sentinel just needs to be a unique NodePtr. Since atoms may be
@@ -112,7 +112,7 @@ impl BlockBuilder {
         bundles: T,
         cost: u64,
         constants: &ConsensusConstants,
-    ) -> io::Result<(bool, BuildBlockResult)>
+    ) -> Result<(bool, BuildBlockResult)>
     where
         T: IntoIterator<Item = S>,
         S: Borrow<SpendBundle>,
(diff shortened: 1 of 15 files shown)


crates/chia-consensus/src/build_compressed_block.rs
```diff
@@ -189,10 +189,7 @@ impl BlockBuilder {
     }
 
     // returns generator, sig, cost
-    pub fn finalize(
-        mut self,
-        constants: &ConsensusConstants,
-    ) -> io::Result<(Vec<u8>, Signature, u64)> {
+    pub fn finalize(mut self, constants: &ConsensusConstants) -> Result<(Vec<u8>, Signature, u64)> {
```
arvidn, line 192: I don't see any changes here, except formatting. did `cargo fmt` do this?

crates/chia-protocol/src/program.rs
```diff
@@ -169,6 +169,14 @@ use pyo3::types::{PyList, PyTuple};
 #[cfg(feature = "py-bindings")]
 use pyo3::exceptions::*;
 
+#[cfg(feature = "py-bindings")]
+#[allow(clippy::needless_pass_by_value)]
+// We use this for the map function, so using a reference here is not ideal.
+fn map_pyerr(err: EvalErr) -> PyErr {
```
arvidn, line 175:
```suggestion
fn map_pyerr(err: &EvalErr) -> PyErr {
```

I would think this would also silence clippy, and you could remove `#[allow(clippy::needless_pass_by_value)]`

wheel/src/api.rs
```diff
@@ -146,8 +149,10 @@ pub fn get_puzzle_and_solution_for_coin<'a>(
     let program = py_to_slice::<'a>(program);
     let args = py_to_slice::<'a>(args);
 
-    let program = node_from_bytes_backrefs(&mut allocator, program)?;
-    let args = node_from_bytes_backrefs(&mut allocator, args)?;
```
arvidn, line 150: these used to return `std::io::Result<>`, and that's why they could implicitly be converted to `PyErr`, is that right?

wheel/src/api.rs
```diff
@@ -159,20 +164,25 @@ pub fn get_puzzle_and_solution_for_coin<'a>(
...
     Ok((
-        PyBytes::new(py, &serialize(&allocator, puzzle)?),
-        PyBytes::new(py, &serialize(&allocator, solution)?),
+        PyBytes::new(
+            py,
+            &serialize(&allocator, puzzle).map_err(|e| map_pyerr_w_ptr(&e, &allocator))?,
+        ),
+        PyBytes::new(
+            py,
+            &serialize(&allocator, solution).map_err(|e| map_pyerr_w_ptr(&e, &allocator))?,
```
arvidn, line 184: all of these explicit conversions to `PyErr` is making it look pretty good to have a `python-binding` build feature on `clvm_rs` that implements `From<EvalErr> for PyErr`

wheel/src/error.rs
```diff
@@ -0,0 +1,13 @@
...
+use clvmr::Allocator;
+use pyo3::exceptions::PyValueError;
+use pyo3::PyErr;
+
+pub fn map_pyerr(err: EvalErr) -> PyErr {
+    PyValueError::new_err(err.to_string())
+}
+pub fn map_pyerr_w_ptr(err: &EvalErr, alloc: &Allocator) -> PyErr {
+    let blob = node_to_bytes(alloc, err.node_ptr()).ok().map(hex::encode);
+    PyValueError::new_err((err.to_string(), blob))
```
arvidn, line 12: using this function exposes more information than we used to, right?

wheel/src/api.rs
jack60612: yep

crates/chia-consensus/src/build_compressed_block.rs
jack60612: i think clippy

wheel/src/api.rs
jack60612: Because I need to make it a toggleable feature to avoid breaking wasm

PR #1158, opened 2025-07-17
Port `check_time_locks()` to Rust and pyo3

matt-o-how:


wjblanke (approved):
aok

crates/chia-consensus/src/check_time_locks.rs
```diff
@@ -0,0 +1,348 @@
...
+            agg_sig_parent_amount: vec![],
+            agg_sig_parent_puzzle: vec![],
+            flags: 0,
+            execution_cost: 0,
+            condition_cost: 0,
+        }
+    }
+
+    fn dummy_bundle(spends: Vec<OwnedSpendConditions>) -> OwnedSpendBundleConditions {
+        OwnedSpendBundleConditions {
```
arvidn, line 162: I think this is the same as `OwnedSpendBundleConditions { spends, ..Default::default()}`

crates/chia-consensus/src/check_time_locks.rs
```diff
@@ -0,0 +1,348 @@
...
+            removal_amount: 0,
+            addition_amount: 0,
+            validated_signature: true,
+            execution_cost: 0,
+            condition_cost: 0,
+        }
+    }
+
+    #[rstest]
+    #[case(0, 0, 10, 1000, None, None, None)]
```
arvidn, line 180: these test cases aren't so easy to read. and it's also not so easy to see that all cases are covered. For example, it looks like you don't cover any cases where there are more than one spend.

Can you think of ways to either make these test cases more concise, or maybe make the `rstest` generate all combinations, and define the outcome programmatically? Maybe some test cases could be separate tests, if that makes it simpler

crates/chia-protocol/src/coin_record.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    pub fn coin_state(&self) -> CoinState {
+        let spent_h = if self.spent_block_index > 0 {
+            Some(self.spent_block_index)
+        } else {
+            None
+        };
+
+        let confirmed_height = if self.confirmed_block_index == 0_u32 && self.timestamp == 0_u64 {
```
Rigidity, line 41: I don't see what the timestamp has to do with the height - shouldn't this be the same as the check above?

crates/chia-consensus/src/check_time_locks.rs
Rigidity: It would be - if `OwnedSpendBundleConditions` implemented `Default`, but it doesn't

crates/chia-protocol/src/coin_record.rs
matt-o-how: You can compare with `chia-blockchain` implementation this is 1:1. As the comment says `timestamp` is the timestamp in the block where the coin is confirmed, if it is not set then the coin is not confirmed

arvidn: the `timestamp` field on `CoinRecord` is a bit unfortunate I think. It's a short cut that mostly exists for this function (I believe). It would be possible to look up the timestamp of the block at `confirmed_block_index`, but we store this along with every coin instead.
In this context, "confirmed" means "created" right? Does `CoinRecord` support referring to coins that have not been created? I'm not so sure.

In fact, if you look in `chia-blockchain`, this function, `coin_state()` is *only* used in tests. I'm tempted to suggest we should remove it (and possibly put it in `chia/_tests/wallet/cat_wallet/test_cat_wallet.py`)

arvidn: I can't think of the scenario where the timestamp would be zero. Even reward coins are ascribed a timestamp of the next transaction block. It would be good to understand where this check comes from

crates/chia-consensus/src/check_time_locks.rs
```diff
@@ -0,0 +1,348 @@
...
+        }
+    }
+
+    None
+}
+
+#[cfg(feature = "py-bindings")]
+#[pyfunction]
+#[pyo3(name = "check_time_locks")]
+pub fn py_check_time_locks(
```
arvidn, line 82: there should be a python test to exercise this function too. it's not trivial

crates/chia-consensus/src/check_time_locks.rs
```diff
@@ -0,0 +1,348 @@
...
+    removal_coin_records: &Bound<'_, PyDict>,
```
arvidn, line 83: can you just take a `HashMap<Bytes32, CoinRecord>` here, and let pyo3 perform the conversion from python?

arvidn (changes requested):
I think the `coin_state()` function probably should be removed

crates/chia-protocol/src/coin_record.rs
```diff
@@ -0,0 +1,49 @@
...
+    spent_block_index: u32,
+    coinbase: bool,
+    timestamp: u64, // Timestamp of the block at height confirmed_block_index
+}
+
+#[cfg(feature = "py-bindings")]
+#[pymethods]
+impl CoinRecord {
+    #[inline]
+    pub fn spent(&self) -> bool {
```
arvidn, line 25: you can make this a property by adding `#[pyo3(get)]`

arvidn: this would make this class much more similar to the current python class

crates/chia-protocol/src/coin_record.rs
```diff
@@ -0,0 +1,49 @@
...
+        self.spent_block_index > 0
+    }
+
+    #[inline]
```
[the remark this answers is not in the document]
matt-o-how: added `#[getter]` for this

crates/chia-consensus/src/check_time_locks.rs
matt-o-how: added

crates/chia-consensus/src/check_time_locks.rs
matt-o-how: Yes apparently. Implemented now.

crates/chia-consensus/src/check_time_locks.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    None
+}
+
+#[cfg(feature = "py-bindings")]
+#[pyfunction]
+#[pyo3(name = "check_time_locks")]
+#[allow(clippy::needless_pass_by_value)] // pyo3 prefers pass_by_value
+pub fn py_check_time_locks(
+    removal_coin_records: HashMap<chia_protocol::BytesImpl<32>, CoinRecord>,
```
arvidn, line 82: a nit, but we have a type alias for this called `Bytes32`, and you're already importing it at the top
```suggestion
    removal_coin_records: HashMap<Bytes32, CoinRecord>,
```

crates/chia-consensus/src/check_time_locks.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    fn dummy_spend_conditions(
+        coin_id: Bytes32,
+        height_relative: Option<u32>,
+        seconds_relative: Option<u64>,
+        before_height_relative: Option<u32>,
+        before_seconds_relative: Option<u64>,
+        birth_height: Option<u32>,
+        birth_seconds: Option<u64>,
+    ) -> OwnedSpendConditions {
```
arvidn, line 127: I think it would be better to implement the `Default` trait than having this function.

crates/chia-consensus/src/check_time_locks.rs
arvidn: if you implement `Default` for `OwnedSpendBundleConditions`, you can probably just take that as the `#[case]` parameter and construct them by just initializing the specific field(s) you need for each test. That would make it much clearer which constraint is being tested. I think you can do something like: `OwnedSpendBundleConditions { seconds_absolute: 100, ..Default::default()}`

crates/chia-consensus/src/check_time_locks.rs
```diff
@@ -0,0 +1,429 @@
...
+        #[case] timestamp: u64,
+        #[case] expected: Option<ErrorCode>,
+    ) {
+        let coin_id = Bytes32::from([3u8; 32]);
+        let coin_record = dummy_coin_record(50, 1000);
+
+        let mut spend = spend;
+        spend.coin_id = coin_id;
+        spend.birth_height = Some(50);
+        spend.birth_seconds = Some(1000);
```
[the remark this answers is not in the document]
matt-o-how: These are two different values. 
`birth_height` and `birth_seconds` vs `new_block_height` and `current_seconds`.
`check_time_locks()` does the following:
```
if new_block_height >= unspent.confirmed_block_index + spend.before_height_relative:
                return Err.ASSERT_BEFORE_HEIGHT_RELATIVE_FAILED
```

I can rename these variables to clarify this if you'd like. They're also confusingly named in `check_time_locks()` for legacy reasons as well.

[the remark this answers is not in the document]
matt-o-how: I removed them anyway

arvidn (approved):

PR #1168, opened 2025-07-23
CHIA-2072 create chia-fuzzing from clvm_rs utils

jack60612:


crates/chia-fuzzing/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+pub mod fuzzing_utils;
+pub mod make_tree;
+pub mod node_eq;
+pub mod pick_node;
+pub mod serialized_len;
```
Rigidity, line 5:
```suggestion
mod fuzzing_utils;
mod make_tree;
mod node_eq;
mod pick_node;
mod serialized_len;

pub use fuzzing_utils::*;
pub use make_tree:*;
pub use node_eq::*;
pub use pick_node::*;
pub use serialized_len::*;
```

crates/chia-fuzzing/src/make_tree.rs
```diff
@@ -0,0 +1,184 @@
...
+
+/// returns an arbitrary CLVM tree structure and the number of (unique) nodes
+/// it's made up of. That's both pairs and atoms.
+pub fn make_tree_limits(
+    a: &mut Allocator,
+    unstructured: &mut Unstructured<'_>,
+    mut max_nodes: i64,
+    reuse_nodes: bool,
+) -> (NodePtr, u32) {
+    let mut previous_nodes = Vec::<NodePtr>::new();
```
Rigidity, line 31: The type hint here is unnecessary, I'd just use `Vec::new()`

crates/chia-fuzzing/src/make_tree.rs
```diff
@@ -0,0 +1,184 @@
...
+    let mut value_stack = Vec::<NodePtr>::new();
+    let mut op_stack = vec![Op::SubTree];
+    // the number of Op::SubTree items on the op_stack
+    let mut sub_trees: i64 = 1;
+    let mut counter = 0;
+
+    while let Some(op) = op_stack.pop() {
+        match op {
+            Op::Pair(swap) => {
+                let left = value_stack.pop().expect("internal error, empty stack");
```
Rigidity, line 41: I know it's fuzzing code so we don't care about handling errors (ie panicking is fine), but to make the code simpler you can return an `anyhow::Result` anyway and use the `?` operator instead of using `expect` on `Result` types. At the call site this whole function can be `.expect`ed.

crates/chia-fuzzing/src/make_tree.rs
```diff
@@ -0,0 +1,184 @@
...
+                };
+                counter += 1;
+                value_stack.push(pair);
+                previous_nodes.push(pair);
+            }
+            Op::SubTree => {
+                sub_trees -= 1;
+                if unstructured.is_empty() {
+                    value_stack.push(NodePtr::NIL);
+                } else {
```
Rigidity, line 56: You can use an early `continue;` in the `if` to un-indent this block

crates/chia-fuzzing/src/make_tree.rs
```diff
@@ -0,0 +1,184 @@
...
+                            });
+                        }
+                    }
+                }
+            }
+        }
+    }
+    assert_eq!(value_stack.len(), 1);
+    (
+        *value_stack.last().expect("internal error, empty stack"),
```
Rigidity, line 134:
```suggestion
        value_stack.remove(0),
```

crates/chia-fuzzing/src/make_tree.rs
```diff
@@ -0,0 +1,184 @@
...
+
+            NodeType::Bytes => a
+                .new_atom(&unstructured.arbitrary::<Vec<u8>>().unwrap_or_default())
+                .unwrap(),
+
+            NodeType::Pair => {
+                let left = NodePtr::NIL;
+                let right = NodePtr::NIL;
+                a.new_pair(left, right).unwrap()
+            }
```
Rigidity, line 170: Would we ever want a list containing pairs of complex trees?

Rigidity (commented):
I think you might be able to use `.map_or(default, |value| ...)` in a few places to simplify `match`es

Rigidity (approved):

crates/chia-fuzzing/src/fuzzing_utils.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+#[allow(dead_code)]
+enum TreeOp {
+    SExp(NodePtr),
+    Cons(NodePtr),
+}
+
+#[allow(dead_code)]
+pub fn tree_hash(a: &Allocator, node: NodePtr) -> [u8; 32] {
```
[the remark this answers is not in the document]
arvidn: I think the reason to have a second implementation here is to compare it against the production implementation.

crates/chia-fuzzing/src/make_tree.rs
arvidn: @jack60612 I see you return `anyhow` from this function, but you didn't implement the suggestion of replacing `.expect()` with `?`. Was this deliberate?

crates/chia-fuzzing/src/serialized_len.rs
arvidn, line 1: I don't really think this is a fuzzing utility as much as just part of fuzzer for the serialization. Not a big deal though

PR #1176, opened 2025-07-28
Remove extract_n() from external uses

matt-o-how:
This PR identifies and removes the two places that use `extract_n()` outside of `run_block_generators()` and removes them in favour of `match_list!` and `destructure_list!`.

It also changes `extract_n` to be a private function to prevent people from using it again.

### DIFF
--- a/crates/chia-consensus/src/spendbundle_conditions.rs
+++ b/crates/chia-consensus/src/spendbundle_conditions.rs
@@ -216,12 +216,12 @@ mod tests {
     #[cfg(not(debug_assertions))]
     fn convert_block_to_bundle(generator: &[u8], block_refs: &[Vec<u8>]) -> SpendBundle {
         use crate::run_block_generator::setup_generator_args;
-        use chia_protocol::Coin;
-        use clvmr::op_utils::first;
-        use clvmr::serde::node_from_bytes_backrefs;
         use chia_protocol::BytesImpl;
+        use chia_protocol::Coin;
         use chia_protocol::Program;
         use clvm_traits::{destructure_list, match_list, FromClvm};
+        use clvmr::op_utils::first;
+        use clvmr::serde::node_from_bytes_backrefs;
 
         let mut a = make_allocator(MEMPOOL_MODE);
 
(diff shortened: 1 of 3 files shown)


crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -240,17 +241,19 @@ mod tests {
...
+                    .expect("parsing CLVM");
             spends.push(CoinSpend::new(
                 Coin::new(
-                    a.atom(parent_id).as_ref().try_into().expect("parent_id"),
-                    tree_hash(&a, puzzle).into(),
-                    a.number(amount).try_into().expect("amount"),
+                    parent_id.try_into().expect("parent_id"),
+                    clvm_utils::tree_hash_from_bytes(puzzle.as_ref())
+                        .expect("hash")
+                        .into(),
```
arvidn, line 252: it's not obvious that this is an improvement. turning a single line of code into 3.

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -240,17 +241,19 @@ mod tests {
         while let Some((spend, rest)) = a.next(all_spends) {
             all_spends = rest;
             // process the spend
-            let [parent_id, puzzle, amount, solution, _spend_level_extra] =
-                extract_n::<5>(&a, spend, ErrorCode::InvalidCondition).expect("extract_n");
-
+            let destructure_list!(parent_id, puzzle, amount, solution) =
+                <match_list!(BytesImpl<32>, Program, u64, Program)>::from_clvm(&a, spend)
+                    .expect("parsing CLVM");
```
arvidn, line 246: are you sure this has the same behavior? if `spend` is a list that has an extra element, will that still match this list?

crates/chia-consensus/src/spendbundle_conditions.rs
matt-o-how: Sure this one line is more verbose, but it actually saves similar casts on lines 253, 255, and 256. This only looks worse because `fmt` is splitting it across multiple lines

Rigidity: If you import tree_hash_from_bytes it'd probably go back to one line 😉

matt-o-how: UPDATE: I have, and it has - and this change is still a net positive given lines saved elsewhere

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -240,17 +243,17 @@ mod tests {
         while let Some((spend, rest)) = a.next(all_spends) {
             all_spends = rest;
             // process the spend
-            let [parent_id, puzzle, amount, solution, _spend_level_extra] =
-                extract_n::<5>(&a, spend, ErrorCode::InvalidCondition).expect("extract_n");
-
+            let destructure_tuple!(parent_id, puzzle, amount, solution, _) =
+                <match_tuple!(BytesImpl<32>, Program, u64, Program, NodePtr)>::from_clvm(&a, spend)
```
arvidn, line 247:
```suggestion
                <match_tuple!(BytesImpl<32>, NodePtr, u64, NodePtr, NodePtr)>::from_clvm(&a, spend)
```

This will prevent the serialization of the `NodePtr` tree into bytes, and be cheaper.

arvidn (approved):

PR #1186, opened 2025-08-01
correct the type stubs for coin_id, parent_id and puzzle_hash

arvidn:
correct type stub for fields in `SpendConditions`


jack60612 (approved):

jack60612 (dismissed):

wheel/python/chia_rs/chia_rs.pyi
```diff
@@ -318,17 +318,17 @@ class PrivateKey:
...
     seconds_relative: Optional[int]
     before_height_relative: Optional[int]
     before_seconds_relative: Optional[int]
     birth_height: Optional[int]
     birth_seconds: Optional[int]
-    create_coin: list[tuple[bytes, int, Optional[bytes]]]
+    create_coin: list[tuple[bytes32, int, Optional[bytes]]]
     agg_sig_me: list[tuple[G1Element, bytes]]
     agg_sig_parent: list[tuple[G1Element, bytes]]
     agg_sig_puzzle: list[tuple[G1Element, bytes]]
```
jack60612, line 334: Make lines 344-346 use bytes32

jack60612 (changes requested):

wheel/python/chia_rs/chia_rs.pyi
jack60612: I can't suggest changes because I'm on my phone

arvidn: the constructor is deliberately made to take `bytes` for initializing the object. It's implemented in `def transform()` in `generate_type_stubs.py`. Since the conversion from python -> rust already validates the length of 32, requiring the python side to also first cast to `bytes32` would be unnecessary

jack60612: i see, dismissing.

jack60612 (approved):

PR #1194, opened 2025-08-05
catchup: into main from long_lived/initial_datalayer  @ cf34dfd8cd7d01f1f3e49df4173506c5ebbd716b

altendky:
Source hash: cf34dfd8cd7d01f1f3e49df4173506c5ebbd716b
Remaining commits: 0


crates/chia-datalayer/src/merkle/blob.rs
```diff
@@ -0,0 +1,2539 @@
...
+    }
+
+    pub fn get_key_index(&self, key: KeyId) -> Result<TreeIndex, Error> {
+        self.block_status_cache
+            .get_index_by_key(key)
+            .copied()
+            .ok_or(Error::UnknownKey(key))
+    }
+
+    pub fn get_proof_of_inclusion(
```
arvidn, line 1152: an idea for a fuzzer:
generate a random tree, ensure that every node can successfully generate a proof-of-inclusion (and maybe even validate the proof)

crates/chia-datalayer/src/merkle/iterators.rs
```diff
@@ -0,0 +1,547 @@
...
+    }
+}
+
+pub struct ParentFirstIterator<'a> {
+    blob: &'a [u8],
+    deque: VecDeque<TreeIndex>,
+    already_queued: HashSet<TreeIndex>,
+}
+
+impl<'a> ParentFirstIterator<'a> {
```
arvidn, line 128: an idea for a fuzzer:
generate a random tree, iterate it with every kind of iterator, ensure that every node was visited exactly once.

tests/test_datalayer.py
arvidn, line 1: normally I would expect tests to be in the same file as the functions/structs. The main reason (in my mind) is that you can test things without making them `pub`. Tests that need a lot of infrastructure and that span multiple files may be warranted to have a separate test file. Was it deliberate to put the tests together here?

altendky:
> I think you'd want this patch as well, run the fuzzers for a moment in CI.

yes...  i would like that.  `:|`  thanks.

- [ ] https://github.com/Chia-Network/chia_rs/pull/1202

crates/chia-datalayer/src/merkle/blob.rs
altendky: - [ ] https://github.com/Chia-Network/chia_rs/pull/1205

PR #1205, opened 2025-08-06
more datalayer fuzzers

altendky:
### Draft For:
- [x] https://github.com/Chia-Network/chia_rs/pull/1202


arvidn (approved):
Just one nit, I'm sure you're aware of. If `get_lineage_blocks_with_indexes()` would return an iterator instead, the `collect()` call could be deferred and you would save some amount of work (heap allocations and copys). It would make the function a bit more complicated, so it may not be worth addressing now.

PR #1206, opened 2025-08-06
address new fuzzer find

altendky:


crates/chia-datalayer/src/merkle/iterators.rs
```diff
@@ -64,9 +64,32 @@ impl Iterator for LeftChildFirstIterator<'_> {
...
+                                node: Node::Leaf(_),
+                                ..
+                            }) => {
+                                return Some(Err(Error::LeafCannotBeParent()));
+                            }
+                            Err(Error::BlockIndexOutOfBounds(_)) => {
+                                return Some(Err(Error::ReferenceToUnknownParent()))
+                            }
+                            Err(e) => return Some(Err(e)),
+                        };
```
arvidn, line 90: I would expect clippy to point out that this `;` is unnecessary

crates/chia-datalayer/src/merkle/iterators.rs
```diff
@@ -64,9 +64,32 @@ impl Iterator for LeftChildFirstIterator<'_> {
...
+                    } else if item.index != self.from_index && !self.already_queued.contains(&index)
```
arvidn, line 91: This expression must always be true, right? `item.index != self.from_index`

arvidn, line 91:
```suggestion
                    } else if !self.already_queued.contains(&index)
```

arvidn (commented):
looks good. Some minor comments

PR #1212, opened 2025-08-11
catchup: into main from long_lived/initial_datalayer  @ dd0b815e234ffb62b4f68599ef8a93eaf86fc07a

altendky:
Source hash: dd0b815e234ffb62b4f68599ef8a93eaf86fc07a
Remaining commits: 0


crates/chia-datalayer/src/merkle/blob.rs
```diff
@@ -0,0 +1,2557 @@
...
+        &mut self,
+        mut keys_values_hashes: Vec<((KeyId, ValueId), Hash)>,
+    ) -> Result<(), Error> {
+        // OPT: perhaps go back to taking an iterator?
+        // OPT: would it be worthwhile to hold the entire blocks?
+        let mut indexes = vec![];
+
+        if self.block_status_cache.leaf_count() <= 1 {
+            for _ in 0..2 {
+                let Some(((key, value), hash)) = keys_values_hashes.pop() else {
```
arvidn, line 585: I take it the insertion order doesn't matter here, because in this special case, you add the last items 2 items first, and then start from the beginning on line 592

crates/chia-datalayer/src/merkle/blob.rs
```diff
@@ -0,0 +1,2557 @@
...
+                hash: internal_node_hash,
+                left: left.index,
+                right: right.index,
+            }),
+        };
+        self.insert_entry_to_blob(new_internal_node_index, &block)?;
+        self.update_parent(new_index, Some(new_internal_node_index))?;
+
+        // TODO: expect relates to comment at the beginning about assumptions about the tree etc
+        let old_leaf_parent = old_leaf.parent.0.expect("not handling this case");
```
arvidn, line 715: is it acceptable to panic here? the message looks like it's something that could happen

crates/chia-datalayer/src/merkle/test_util.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    small_blob.calculate_lazy_hashes().unwrap();
+    small_blob
+}
+
+pub fn generate_kvid(seed: i64) -> (KeyId, ValueId) {
+    let mut kv_ids: Vec<i64> = Vec::new();
+
+    for offset in 0..2 {
+        let seed_int = 2i64 * seed + offset;
```
arvidn, line 71: if `seed` is allowed to be a sufficiently large or sufficiently small value, this may overflow the integer. If that's deliberate, you should use `seed.wrapping_mul(2).wrapping_add(offset)`

crates/chia-datalayer/src/merkle/blob.rs
altendky: This is indeed a bit awkward and likely doesn't need to be this way.  But also yes, the order doesn't matter here.  If the caller wants to control the tree structure they need to use another insert mechanism.

crates/chia-datalayer/src/merkle/test_util.rs
altendky: looks like we're trying to spread the seed space out so we can get two unique results from one seed.  the key and the value.  i'll constrain the seed parameter to i32.

- [ ] https://github.com/Chia-Network/chia_rs/pull/1214

PR #1213, opened 2025-08-12
Dedup fingerprint

arvidn:

Fold `compute_puzzle_fingerprint()` into `validate_clvm_and_signature()` (used by the mempool to validate transactions). With a new flag, `COMPUTE_FINGERPRINT`, each spend eligible for dedup will have its conditions fingerprint computed and populated (in the `SpendConditions` object).

This will allow the mempool to avoid running dedup-puzzles again, and simplify the mempool since the `MempoolItems`'s fingerprint is populated from the start.

### DIFF
--- a/crates/chia-consensus/src/conditions.rs
+++ b/crates/chia-consensus/src/conditions.rs
@@ -922,9 +922,9 @@ pub(crate) fn parse_single_spend(
 }
 
 #[allow(clippy::too_many_arguments)]
-pub fn process_single_spend<V: SpendVisitor>(
+pub fn process_single_spend<'a, V: SpendVisitor>(
     a: &Allocator,
-    ret: &mut SpendBundleConditions,
+    ret: &'a mut SpendBundleConditions,
     state: &mut ParseState,
     parent_id: NodePtr,
     puzzle_hash: NodePtr,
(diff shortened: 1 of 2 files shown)


crates/chia-consensus/src/conditions.rs
```diff
@@ -754,6 +754,10 @@ pub struct SpendConditions {
 
     /// the cost of the conditions generated by this puzzle
     pub condition_cost: u64,
+
+    /// If this spend is eligible for fast-forward, this fingerprint is set to
```
AmineKhaldi, line 758: You mean dedup, right?

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -94,6 +93,13 @@ pub fn run_spendbundle(
             clvm_cost,
             constants,
         )?;
+
+        let Some(ref mut spend) = ret.spends.last_mut() else {
+            panic!("process_single_spend() failed to add a spend");
+        };
```
AmineKhaldi, line 99: Hmm this looks weird, `process_single_spend` already returns a `Result` so by reaching this line it's reasonable to assume we added a spend. Isn't it better to make its success path return the spend (instead of unit) and use it here?

arvidn: this statement *does* assume we added the spend (and panics otherwise). I'm hesitant to always make a copy of the spend just to use it rarely (i.e. `ELIGIBLE_FOR_DEDUP` when `COMPUTE_FINGERPRINT` is specified).

arvidn: we can't return a mutable reference from `process_single_spend()`, since that would violate borrowing rules. The object is already referenced by `ret.spends`

AmineKhaldi (approved):

PR #1214, opened 2025-08-12
avoid potential overflows when creating test data

altendky:


crates/chia-datalayer/src/merkle/test_util.rs
```diff
@@ -64,11 +64,11 @@ pub fn traversal_blob(mut small_blob: MerkleBlob) -> MerkleBlob {
     small_blob
 }
 
-pub fn generate_kvid(seed: i64) -> (KeyId, ValueId) {
+pub fn generate_kvid(seed: i32) -> (KeyId, ValueId) {
     let mut kv_ids: Vec<i64> = Vec::new();
 
     for offset in 0..2 {
-        let seed_int = 2i64 * seed + offset;
+        let seed_int = 2i64 * seed as i64 + offset;
```
arvidn, line 71: a minor comment: the `as` operator allows to silently truncate values. Since we don't intend that to happen here, a more robust way of saying this is `i64::from(seed)`. If `seed` would change type such that this conversion is no longer guaranteed to be lossless, it would trigger a compilation error.

arvidn (approved):

crates/chia-datalayer/src/merkle/test_util.rs
altendky:
```suggestion
        let seed_int = 2i64 * i64::from(seed) + offset;
```

PR #1228, opened 2025-08-20
stop using blspy

altendky:


crates/chia-bls/src/public_key.rs
```diff
@@ -494,15 +494,6 @@ mod tests {
...
     fn test_aggregate_pubkey() {
-        // from blspy import PrivateKey
-        // from blspy import AugSchemeMPL
-        // sk = PrivateKey.from_bytes(bytes.fromhex("52d75c4707e39595b27314547f9723e5530c01198af3fc5849d9a7af65631efb"))
-        // pk = sk.get_g1()
-        // pk + pk
-        // <G1Element b1b8033286299e7f238aede0d3fea48d133a1e233139085f72c102c2e6cc1f8a4ea64ed2838c10bbd2ef8f78ef271bf3>
-        // pk + pk + pk
-        // <G1Element a8bc2047d90c04a12e8c38050ec0feb4417b4d5689165cd2cea8a7903aad1778e36548a46d427b5ec571364515e456d6>
-
```
arvidn, line 505: This is a description of how the test values were determined. I think it makes sense to keep this, as well as the similar comments in the other tests. I prefer it over just having seemingly random numbers in the test with no explanation

arvidn (commented):
looks good, but I think we should keep the comments in the rust tests describing where the test values came from

PR #1257, opened 2025-09-19
Add costing to the treehash in run_generators (with a flag and new ConsensusConstants)

matt-o-how:


crates/chia-consensus/src/flags.rs
```diff
@@ -27,5 +27,9 @@ pub const COMPUTE_FINGERPRINT: u32 = 0x4_0000;
 /// processing them in in chia_rs. It is set to activate after hard fork 2.
 pub const COST_CONDITIONS: u32 = 0x80_0000;
 
+// This flag controls whether we are adding the cost of the sha_trees for
+// the puzzle reveals in a generator into the generators cost
+pub const COST_SHATREE: u32 = 0x100_000;
```
arvidn, line 32: maybe referring to just puzzle hash is clearer. This doesn't affect *all* tree hashing, right?

crates/chia-consensus/src/test_generators.rs
```diff
@@ -234,7 +234,7 @@ fn run_generator(#[case] name: &str) {
     // When making changes to print_conditions() enabling this will update the
     // test cases to match. Make sure to carefully review the diff before
     // landing an automatic update of the test case.
-    const UPDATE_TESTS: bool = false;
+    const UPDATE_TESTS: bool = true;
```
arvidn, line 237: left-over

crates/chia-consensus/src/test_generators.rs
```diff
@@ -331,7 +343,9 @@ fn run_generator(#[case] name: &str) {
                     // before the hard fork, the cost of running the genrator +
                     // puzzles should never be lower than after the hard-fork
                     // but it's likely higher.
-                    assert!(conditions.cost >= expected_cost);
+                    if flags & COST_SHATREE == 0 {
```
arvidn, line 346:
```suggestion
                    if (flags & COST_SHATREE) == 0 {
```

crates/chia-consensus/src/test_generators.rs
```diff
@@ -305,8 +311,14 @@ fn run_generator(#[case] name: &str) {
         };
 
         if UPDATE_TESTS {
-            if (flags & MEMPOOL_MODE) != 0 {
+            if flags & COST_SHATREE != 0 {
```
arvidn, line 314:
```suggestion
            if (flags & COST_SHATREE) != 0 {
```

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -307,6 +308,79 @@ pub fn tree_hash_cached(a: &Allocator, node: NodePtr, cache: &mut TreeCache) ->
     hashes[0]
 }
 
+fn subtract_cost(cost_left: &mut Cost, subtract: Cost) -> Result<(), usize> {
+    if subtract > *cost_left {
+        Err(0)
```
arvidn, line 313:
```suggestion
fn subtract_cost(cost_left: &mut Cost, subtract: Cost) -> Result<(), ()> {
    if subtract > *cost_left {
        Err(())
```

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -307,6 +308,79 @@ pub fn tree_hash_cached(a: &Allocator, node: NodePtr, cache: &mut TreeCache) ->
...
+}
+
+pub fn tree_hash_cached_costed(
+    a: &Allocator,
+    node: NodePtr,
+    cache: &mut TreeCache,
+    cost_left: &mut u64,
+    cost_per_call: u32,
+    cost_per_byte: u32,
+) -> Result<TreeHash, usize> {
```
arvidn, line 327:
```suggestion
) -> Result<TreeHash, ()> {
```
I don't think the `usize` carries any meaning, right?

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -307,6 +308,79 @@ pub fn tree_hash_cached(a: &Allocator, node: NodePtr, cache: &mut TreeCache) ->
...
+    cache.visit_tree(a, node);
+
+    let mut hashes = Vec::new();
+    let mut ops = vec![TreeOp::SExp(node)];
+
+    while let Some(op) = ops.pop() {
```
arvidn, line 333: did you consider making the existing implementation a generic/template so you could reuse it for both implementations? Having two implementations adds a lot of testing burden

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -307,6 +308,79 @@ pub fn tree_hash_cached(a: &Allocator, node: NodePtr, cache: &mut TreeCache) ->
...
+        subtract_cost(cost_left, cost_per_call.into())?;
+        match op {
+            TreeOp::SExp(node) => match a.node(node) {
+                NodeVisitor::Buffer(bytes) => {
+                    subtract_cost(cost_left, (cost_per_byte * bytes.len() as u32).into())?;
```
arvidn, line 338:
```suggestion
                    subtract_cost(cost_left, u64::from(cost_per_byte) * u64::from(bytes.len()))?;
```
the `as` operator isn't as safe as `::from()` since it will silently truncate values. `Cost` is u64, so `u32` doesn't seem like the right type

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -307,6 +308,79 @@ pub fn tree_hash_cached(a: &Allocator, node: NodePtr, cache: &mut TreeCache) ->
...
+                    let hash = tree_hash_atom(bytes);
+                    hashes.push(hash);
+                }
+                NodeVisitor::U32(val) => {
+                    subtract_cost(cost_left, (cost_per_byte * 4).into())?; // 4 bytes in u32
```
arvidn, line 343: this is not necessarily 4 bytes. There should be a unit test to cover this case, with short atoms. 1, 2, 3, 4 and 5 bytes.

In `clvm_rs` there's a function `len_for_value()` which can tell you how many bytes. But you can also ask the `Allocator`:
```
a.atom_len(node)
```

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -307,6 +308,79 @@ pub fn tree_hash_cached(a: &Allocator, node: NodePtr, cache: &mut TreeCache) ->
...
+    }
+}
+
+pub fn tree_hash_cached_costed(
+    a: &Allocator,
+    node: NodePtr,
+    cache: &mut TreeCache,
+    cost_left: &mut u64,
+    cost_per_call: u32,
+    cost_per_byte: u32,
```
arvidn, line 326:
```suggestion
    cost_per_call: Cost,
    cost_per_byte: Cost,
```

Alternatively, you can use `u64`. But costs are `u64`

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -307,6 +308,79 @@ pub fn tree_hash_cached(a: &Allocator, node: NodePtr, cache: &mut TreeCache) ->
...
+                NodeVisitor::U32(val) => {
+                    subtract_cost(cost_left, (cost_per_byte * 4).into())?; // 4 bytes in u32
+                    if (val as usize) < PRECOMPUTED_HASHES.len() {
+                        hashes.push(PRECOMPUTED_HASHES[val as usize]);
+                    } else {
+                        hashes.push(tree_hash_atom(a.atom(node).as_ref()));
+                    }
+                }
+                NodeVisitor::Pair(left, right) => {
+                    subtract_cost(cost_left, cost_per_byte.into())?;
```
arvidn, line 351: isn't this supposed to be the `cost_per_byte * 64`?

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -307,6 +308,79 @@ pub fn tree_hash_cached(a: &Allocator, node: NodePtr, cache: &mut TreeCache) ->
...
+                    if let Some(hash) = cache.get(node) {
+                        hashes.push(*hash);
```
arvidn, line 353: this case is not accounting cost correctly. You should have a test case for this, that ensures the same tree always has the same cost to tree-hash, regardless of cache hits.

crates/chia-consensus/src/test_generators.rs
```diff
@@ -255,8 +255,14 @@ fn run_generator(#[case] name: &str) {
...
-        Some((c, m)) => [c, m],
-        None => [expected, expected],
+        Some((c, m)) => match m.split_once("COSTED_SHA:\n") {
+            Some((d, n)) => [c, d, n],
+            None => [c, m, c],
+        },
+        None => match expected.split_once("COSTED_SHA:\n") {
+            Some((d, n)) => [d, d, n],
+            None => [expected, expected, expected],
+        },
```
arvidn, line 265: instead of duplicating all of the output, it might be a good idea to *just* have the alternative costs when this flag is set. Some tests have very large outputs that would be a shame to have to duplicate

matt-o-how: but if we are not expecting a difference and find one, then that *should* be flagged up, right?

arvidn: This flag is never expected to affect the output, right? It's only affecting cost. So the test cases could specify two costs, one with and one without this flag. That would be a lot more concise than to duplicate all of the output. It would also be easier to manually verify the test cases.

arvidn:
This change also needs to be incorporated in the mempool that builds the block. It needs to take into account the hashing cost of puzzles to avoid exceeding the max cost. But only after the hard fork.

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -75,7 +77,20 @@ pub fn run_spendbundle(
...
-        let buf = tree_hash(a, puz);
+        cache.visit_tree(a, puz);
+        let buf = if flags & COST_SHATREE != 0 {
+            tree_hash_cached_costed(
+                a,
+                puz,
+                &mut cache,
+                &mut cost_left,
+                constants.shatree_recurse_cost,
+                constants.shatree_cost_per_byte,
```
arvidn, line 88: I'm not sure we necessarily need to add these to `ConsensusConstants`. I don't think we put any other cost-related constants in there. Except cost_per_byte I suppose. But condition cost are not in there, nor any other clvm execution costs

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -298,314 +358,569 @@ pub fn tree_hash_cached(a: &Allocator, node: NodePtr, cache: &mut TreeCache) ->
...
-    let atom2_hash = {
-        let mut sha256 = Sha256::new();
-        sha256.update([1_u8]);
-        sha256.update([4, 5, 6]);
-        let atom2_hash = sha256.finalize();
-
-        assert_eq!(tree_hash(&a, atom2).as_ref(), atom2_hash.as_slice());
-        atom2_hash
-    };
+#[allow(clippy::result_unit_err)]
```
arvidn, line 381: here too, I think it's better to return `Option<>` instead of suppressing this warning

tests/test-generators.py
```diff
@@ -143,29 +144,56 @@ def validate_except_cost(output1: str, output2: str) -> None:
...
-                and mempool2.result is not None
-                and consensus2.result.cost != mempool2.result.cost
-            ):
-                print("\n\ncost when running in mempool mode differs from normal mode!")
-                failed = 1
+        if "STRICT:\n" in expected:
+            # split STRICT section
+            consensus_part, rest = expected.split("STRICT:\n", 1)
+
+            if "COSTED_SHA:\n" in rest:
```
arvidn, line 165: I have the same comment here. The test cases become much more difficult to read by duplicating the outputs like this. I really think you should just have separate outputs for the cost

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -162,6 +169,10 @@ enum TreeOp {
     SExp(NodePtr),
     Cons,
     ConsAddCache(NodePtr),
```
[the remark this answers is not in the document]
matt-o-how: Yes, for the non-costed path

crates/clvm-utils/src/tree_hash.rs
matt-o-how: Done

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -378,62 +485,71 @@ mod tests {
             5_000_000,
             &TEST_CONSTANTS,
         );
+        check_output(
```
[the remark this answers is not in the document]
matt-o-how: Done

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -379,61 +520,78 @@ mod tests {
...
-        }
+        // === COSTED_SHA case ===
+        if !costed_out.trim().is_empty() {
+            let mut a2_costed = make_allocator(COST_SHATREE);
+            let mut a1_costed = make_allocator(COST_SHATREE);
 
-        if output != expected {
-            print_diff(&output, expected);
-            panic!("mismatching condition output");
+            let conds_res = run_spendbundle(
```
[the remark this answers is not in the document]
matt-o-how: `get_conditions_from_spendbundle()` is calling `run_spendbundle()` but calling it directly allows us to set the flags manually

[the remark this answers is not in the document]
arvidn: line 515 should probably call `run_spendbundle()` as well then

crates/clvm-utils/src/tree_hash.rs
```diff
@@ -246,14 +260,57 @@ pub fn tree_hash(a: &Allocator, node: NodePtr) -> TreeHash {
...
-            TreeOp::ConsAddCache(_) => unreachable!(),
+            TreeOp::ConsAddCache(_) | TreeOp::ConsAddCacheCost(_, _) => unreachable!(),
         }
     }
 
     assert!(hashes.len() == 1);
     hashes[0]
 }
 
+pub fn tree_hash_costed(a: &Allocator, node: NodePtr, cost_left: &mut Cost) -> Option<TreeHash> {
```
arvidn, line 271: as far as I can tell, this function is only used by tests. Can we remove it?

crates/chia-consensus/src/spendbundle_conditions.rs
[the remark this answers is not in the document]
arvidn: I see. in that case, I think it would be easier to understand this test if line 515 also called `run_spendbundle()`

crates/chia-consensus/src/spendbundle_conditions.rs
```diff
@@ -342,23 +342,52 @@ mod tests {
             }
         };
 
-        if output != block_output {
-            print_diff(&output, &block_output);
-            panic!(
-            "run_block_generator2 produced a different result than get_conditions_from_spendbundle()"
-        );
+        let expected_trimmed = expected.trim();
+        let output_trimmed = output.trim();
```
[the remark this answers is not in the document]
matt-o-how: Done

crates/clvm-utils/src/tree_hash.rs
matt-o-how: I've moved it into the test mod

PR #1371, opened 2026-02-27
[CHIA-3854 CHIA-3823] Generator identity hard fork (rebased, P=3 fix)

richardkiss:
> **See also:** #1377 — alternative pure storage cost model (12000/0) for comparison.

## Summary

Implements the Generator Identity Hard Fork with a **split cost model** (`SIZE_COST_PER_BYTE=6000`, `SHA_COST_PER_UNIT=4500`):

- New `INTERNED_GENERATOR` consensus flag, activated at `hard_fork2_height`
- Generators are deserialized into an interned (canonical) tree so cost limits apply to the deduplicated structure independent of serialization
- New `generator_cost` module computes cost from `InternedTree` in one pass
- `run_block_generator2` dispatches to interned path when flag is set
- `run_spendbundle` uses interned cost for mempool validation
- Shared execution logic extracted into `run_block_generator_inner` to avoid duplicating the spend-parsing loop

## Cost Model: Split (6000/4500)

Uses two coefficients — size and SHA — to charge for both storage and computation:
- `size_cost = (atom_bytes + 2*atom_count + 3*pair_count) * 6000`
- `sha_cost = (sha_blocks + 8*sha_invocations) * 4500`

The pair coefficient P=3 is formally proven to ensure `size_cost >= serialized_byte_count`.

**See #1377 for the alternative pure storage model** (12000/0), which is simpler (one coefficient) but ~20% higher cost for typical generators.

## Test Plan

- [x] Unit tests pass (`cargo test -p chia-consensus` — 2252 tests)
- [x] `cargo fmt` / `cargo clippy` clean
- [ ] `analyze-chain` full mainnet regression (in progress)
- [ ] CI green

## Related

- Supersedes #1356
- Depends on #1367 (merged)
- Alternative: #1377 (pure storage model)

---

> [!NOTE]
> **High Risk**
> High risk because it changes consensus-critical generator/spendbundle cost accounting and validation behavior (including a new hard-fork-activated flag), which could affect block/mempool acceptance if the cost model or interning logic is wrong.
> 
> **Overview**
> Implements the generator-identity hard fork by introducing `ConsensusFlags::INTERNED_GENERATOR` (enabled at `hard_fork2_height`) and switching generator validation/costing to operate on an *interned/canonical* CLVM tree so limits and costs are independent of serialization/backrefs.
> 
[...]

PR #1377, opened 2026-03-04
[CHIA-3854] Alternative: pure storage cost model (12000/0) for generator identity HF

richardkiss:
> **See also:** #1371 — the split cost model (6000/4500), which is the other option.

## Summary

Alternative cost model for the generator identity hard fork. Instead of the split model in #1371 (which charges separately for size and SHA), this uses a **pure storage model**:

```
cost = (atom_bytes + 2*atom_count + 3*pair_count) * 12000
```

One constant, no SHA cost accounting. The SHA tree-hash computation is structurally bounded by the size component (worst-case ratio ≤ 3.33), so it doesn't need separate charging.

All other code (INTERNED_GENERATOR flag, interned tree path, run_block_generator_inner refactor) is identical to #1371. The only file that differs is `generator_cost.rs`, which is smaller here since the SHA calculation is removed entirely.

## Why this formula?

The hard fork switches generator identity from `SHA256(serialized_bytes)` to `SHA256_tree_hash(tree)` — content-addressable rather than serialization-dependent. This means cost can no longer be `len(serialized_bytes) * 12000` because different serializations of the same tree would have different costs. Instead, cost must be a function of the **interned tree structure**.

The formula `atom_bytes + 2*atom_count + 3*pair_count` counts the structural components of the deduplicated tree:

- **`atom_bytes`** — raw data stored on chain
- **`2 * atom_count`** — per-atom overhead (type byte + length encoding)
- **`3 * pair_count`** — per-pair overhead (P=3 is [formally proven](https://github.com/richardkiss/generator-identity-hf-analysis/blob/main/docs/GENERATOR_IDENTITY_HARDFORK.md) to be an upper bound on serialized byte count, accounting for pair opcodes and back-reference encoding overhead)

The multiplier 12000 matches the pre-HF2 `COST_PER_BYTE`, preserving the same cost scale.

Full analysis and derivation: [generator-identity-hf-analysis](https://github.com/richardkiss/generator-identity-hf-analysis/blob/main/docs/GENERATOR_IDENTITY_HARDFORK.md)

## Pure storage (this PR) vs split (#1371)

| | Pure storage (this PR) | Split (#1371) |
|---|---|---|
| **Formula** | `size * 12000` | `size * 6000 + sha * 4500` |
| **Coefficients** | 1 | 2 |
| **Typical generator cost** | ~20% higher | ~20% lower |
| **Effective TPS** | Lower | Higher |
| **DoS safety** | Structural bound (sha/size ≤ 3.33, worst-case ~37ms SHA CPU) | Explicit SHA charging |
| **Philosophy** | Charge for storage only | Charge for storage + computation |

### Pros of pure storage
- **Simpler:** one constant, no SHA block/invocation accounting
- **DoS safe:** structural bound guarantees SHA CPU is bounded by size budget (worst case ~37ms on 2012 Celeron)
- Returns to pre-HF2 philosophy: charge for what you store on chain
- `size_component` naturally correlates with SHA cost (pair-heavy trees have high SHA AND high size)

### Cons of pure storage
- ~20% higher base cost for typical generators → lower effective TPS
- Doesn't explicitly charge for SHA computation (relies on structural bound)

## Test plan
- [x] All existing tests pass (`cargo test -p chia-consensus`)
- [ ] `analyze-chain` full mainnet regression (in progress on #1371, same code path for existing blocks)

---

> [!NOTE]
> **High Risk**
> Changes consensus-critical generator cost accounting and decoding paths when `INTERNED_GENERATOR` is enabled, which can alter block/spendbundle acceptability at the cost limit. New interned-tree validation/costing logic introduces potential edge cases around decoding limits and allocator ownership transitions.
> 
> **Overview**
> Implements an alternative generator-identity hard fork path behind the new `ConsensusFlags::INTERNED_GENERATOR`, switching base generator costing from `len(serialized_bytes) * cost_per_byte` to a **pure storage model** computed from the *interned (deduplicated) CLVM tree*.
> 
> Adds `generator_cost::total_cost_from_tree()` and wires it into `run_block_generator2()` and mempool `run_spendbundle()` so both block generators and spendbundles can charge consistent, serialization-independent base cost; `solution_generator::build_generator()` is exposed for internal reuse to build the generator tree for interning.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9774629f95af4cc95d27e6957efdc1a0963698f8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/chia-consensus/src/flags.rs
+++ b/crates/chia-consensus/src/flags.rs
@@ -49,6 +49,11 @@ bitflags! {
 
         /// Limit the number of spends per block.
         const LIMIT_SPENDS = 0x200_0000;
+
+        /// After the generator-identity hard fork, generators must be validated from
+        /// the INTERNED (canonical) tree so atom/pair limits and cost apply to the same
+        /// structure independent of serialization.
+        const INTERNED_GENERATOR = 0x0800_0000;
     }
 }
 
--- a/crates/chia-consensus/src/generator_cost.rs
+++ b/crates/chia-consensus/src/generator_cost.rs
@@ -1,0 +1,71 @@
+//! Chia-specific generator cost calculation.
+//!
+//! Pure storage model: cost = (atom_bytes + 2*atoms + 3*pairs) * COST_PER_BYTE.
+//! SHA tree-hash cost is not charged separately — it is structurally bounded
+//! by the size component (worst-case ratio <= 3.33, ~37ms SHA CPU on a 2012
+//! Celeron). See PR #1371 for the alternative split model (6000/4500).
+
+use clvmr::serde::InternedTree;
+
+const COST_PER_BYTE: u64 = 12000;
+
+/// Compute total generator cost from an interned tree.
+///
+/// The size formula `atom_bytes + 2*atom_count + 3*pair_count` is proven to
+/// be an upper bound on the serialized byte count (P=3 accounts for pair
+/// opcodes and back-reference overhead).
+#[inline]
+pub fn total_cost_from_tree(tree: &InternedTree) -> u64 {
+    let atom_count = tree.atoms.len() as u64;
+    let pair_count = tree.pairs.len() as u64;
+
+    let mut atom_bytes: u64 = 0;
+    for &atom in &tree.atoms {
+        atom_bytes += tree.allocator.atom_len(atom) as u64;
+    }
+
+    (atom_bytes + 2 * atom_count + 3 * pair_count) * COST_PER_BYTE
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use clvmr::allocator::Allocator;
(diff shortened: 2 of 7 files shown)


arvidn (dismissed):
looks good! the branch needs to be rebased though.
Also, once this lands, it means `chia-blockchain` can't update to this new version of `chia_rs` until it can *produce* valid interned blocks.
That means `BlockTools` and block creation needs an update. You may want to plan for that and maybe expose some functionality in `chia_rs` to help.

The `BlockBuilder` class (here in `chia_rs`) currently relies on predicting the cost of a block as it compresses. It may need some way to track the new cost model as well.

arvidn (approved):

PR #1423, opened 2026-04-28
Streamable size limits

arvidn:
This adds an option to the streamable macro, to specify a limit on vectors in streamable types. It can short cut some validation by failing earlier for messages that are invalid.

The second commit employs the new option to set appropriate limits for some message types.

---

> [!NOTE]
> **Medium Risk**
> Changes deserialization behavior for `Streamable` structs by enforcing new per-field vector length limits, which can cause previously-accepted oversized network messages to be rejected. Risk is moderate because it touches core protocol parsing but is additive and guarded by explicit `#[chia(max_length = ...)]` annotations.
> 
> **Overview**
> Adds a new `#[chia(max_length = N)]` field attribute to the `Streamable` derive macro to enforce bounded parsing for `Vec<T>` and `Option<Vec<T>>`, failing early with `Error::SequenceTooLarge`.
> 
> Refactors vector parsing in `chia-traits` to route through a shared `parse_vec_with_max_length` helper and adds tests covering bounded `Vec`/`Option<Vec>` behavior.
> 
> Applies explicit maximum lengths to several full node and wallet protocol message fields (e.g., block/header responses, additions/removals queries, subscription update payloads, mempool notifications) to cap decoded list sizes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c79e84e9b7ef35b83e5d090d9c3f73f53f203f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/chia-protocol/src/wallet_protocol.rs
+++ b/crates/chia-protocol/src/wallet_protocol.rs
@@ -72,6 +72,7 @@ pub struct RejectHeaderRequest {
 pub struct RequestRemovals {
     height: u32,
     header_hash: Bytes32,
+    #[chia(max_length = 30000)]
     coin_names: Option<Vec<Bytes32>>,
 }
 
@@ -79,7 +80,9 @@ pub struct RequestRemovals {
 pub struct RespondRemovals {
     height: u32,
     header_hash: Bytes32,
+    #[chia(max_length = 30000)]
     coins: Vec<(Bytes32, Option<Coin>)>,
+    #[chia(max_length = 30000)]
     proofs: Option<Vec<(Bytes32, Bytes)>>,
 }
 
@@ -93,6 +96,7 @@ pub struct RejectRemovalsRequest {
 pub struct RequestAdditions {
     height: u32,
     header_hash: Option<Bytes32>,
+    #[chia(max_length = 30000)]
     puzzle_hashes: Option<Vec<Bytes32>>,
 }
 
@@ -100,7 +104,9 @@ pub struct RequestAdditions {
 pub struct RespondAdditions {
     height: u32,
     header_hash: Bytes32,
+    #[chia(max_length = 30000)]
     coins: Vec<(Bytes32, Vec<Coin>)>,
+    #[chia(max_length = 30000)]
     proofs: Option<Vec<(Bytes32, Bytes, Option<Bytes>)>>,
 }
 
@@ -114,6 +120,7 @@ pub struct RejectAdditionsRequest {
 pub struct RespondBlockHeaders {
     start_height: u32,
     end_height: u32,
+    #[chia(max_length = 64)]
     header_blocks: Vec<HeaderBlock>,
 }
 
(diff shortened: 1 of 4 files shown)


crates/chia-protocol/src/wallet_protocol.rs
```diff
@@ -245,6 +265,7 @@ pub struct CoinStateFilters {
 
 #[streamable(message)]
 pub struct RequestPuzzleState {
+    #[chia(max_length = 32700)]
```
Rigidity: If register_for_coin_updates and register_for_ph_updates can allow 1.6 million, so should these. This limit is insufficient for wallets like Sage to function. Many wallets have over this many puzzle hashes to subscribe to (each one is a wallet address).

crates/chia-protocol/src/wallet_protocol.rs
```diff
@@ -254,10 +275,12 @@ pub struct RequestPuzzleState {
 
 #[streamable(message)]
 pub struct RespondPuzzleState {
+    #[chia(max_length = 32700)]
```
Rigidity: Same as above

crates/chia-protocol/src/wallet_protocol.rs
```diff
@@ -254,10 +275,12 @@ pub struct RequestPuzzleState {
...
     puzzle_hashes: Vec<Bytes32>,
     height: u32,
     header_hash: Bytes32,
     is_finished: bool,
+    #[chia(max_length = 32700)]
```
Rigidity: Same as above

crates/chia-protocol/src/wallet_protocol.rs
```diff
@@ -268,6 +291,7 @@ pub struct RejectPuzzleState {
 
 #[streamable(message)]
 pub struct RequestCoinState {
+    #[chia(max_length = 32700)]
```
Rigidity: Same as above

crates/chia-protocol/src/wallet_protocol.rs
```diff
@@ -276,7 +300,9 @@ pub struct RequestCoinState {
 
 #[streamable(message)]
 pub struct RespondCoinState {
+    #[chia(max_length = 32700)]
```
Rigidity: Same as above

crates/chia-protocol/src/wallet_protocol.rs
```diff
@@ -276,7 +300,9 @@ pub struct RequestCoinState {
...
     coin_ids: Vec<Bytes32>,
+    #[chia(max_length = 32700)]
```
Rigidity: Same as above

crates/chia-traits/src/streamable.rs
```diff
@@ -124,19 +124,28 @@ impl<T: Streamable> Streamable for Vec<T> {
...
+    let len = u32::parse::<TRUSTED>(input)?;
+    if (len as usize) > max_length {
+        return Err(Error::SequenceTooLarge);
     }
+    let mut ret = if mem::size_of::<T>() == 0 {
+        Vec::<T>::new()
+    } else {
+        let limit = 2 * 1024 * 1024 / mem::size_of::<T>();
+        Vec::<T>::with_capacity(std::cmp::min(limit, len as usize))
+    };
```
Rigidity, line 144: This could be simplified to:
```suggestion
    let limit = 2 * 1024 * 1024 / mem::size_of::<T>();
    let mut ret = Vec::<T>::with_capacity(std::cmp::min(limit, len as usize));
```

Because https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity will only allocate if it's not a zero sized type already, and otherwise the capacity is usize::MAX.

Rigidity: Ah, never mind - if it's a zero size type then that would produce a division by zero :)
Being clever isn't worth a panic

crates/chia-protocol/src/wallet_protocol.rs
arvidn: the 1.6 million is derived from the 50 MiB message limit, which will not fit more 32 byte hashes.

Does Sage subscribe to all coins in a single request? This number was picked based on `CoinStore.MAX_PUZZLE_HASH_BATCH_SIZE`, see: https://github.com/Chia-Network/chia-blockchain/blob/main/chia/full_node/full_node_api.py#L1949

The limit is defined here: https://github.com/Chia-Network/chia-blockchain/blob/main/chia/full_node/coin_store.py#L444

It seems like we silently drop any item past this limit.

Rigidity (approved):

PR #1436, opened 2026-05-08
Add InternedBlockBuilder for INTERNED_GENERATOR cost model

richardkiss:
Do `diff -u crates/chia-consensus/src/build_compressed_block.rs  crates/chia-consensus/src/build_interned_block.rs` to see how the new file `build_interned_block.rs` is just a minimal diff from existing code.

-----------
## Summary

Add `InternedBlockBuilder`, a block generator builder for the post-HF2 `INTERNED_GENERATOR` cost model.

~~Unlike `BlockBuilder`, which estimates serialization cost incrementally using a sentinel-based approach, `InternedBlockBuilder` computes cost exactly by interning the full quoted generator tree after each `add_spend_bundles()` call and applying `total_cost_from_tree()`. Serialization happens once in `finalize()` via `node_to_bytes_backrefs`.~~

Like `BlockBuilder`, cost is tracked incrementally. Each spend's vbyte contribution is estimated by interning it in an isolated scratch allocator (`spend_vbytes`); by the triangle inequality this sum is a safe upper bound on the true interned cost of the combined tree. `cost()` returns that upper bound. `finalize()` builds the full quoted generator, serializes with `node_to_bytes_backrefs`, re-interns from the root to get the exact cost, and asserts it's within `max_block_cost_clvm`.

## Changes

- `crates/chia-consensus/src/build_interned_block.rs` (new): `InternedBlockBuilder` with Rust and Python (`#[pymethods]`) interface
- `wheel/src/api.rs`: register `InternedBlockBuilder` with `add_class`
- `wheel/python/chia_rs/chia_rs.pyi` + `generate_type_stubs.py`: type stubs

## Test plan

- [ ] `cargo test -p chia-consensus` passes
~~`from chia_rs import InternedBlockBuilder; b = InternedBlockBuilder(); b.cost()` returns 20~~
- [ ] `from chia_rs import InternedBlockBuilder; b = InternedBlockBuilder(constants); b.cost()` returns expected wrapper cost
- [ ] CI: `Check chia_rs.pyi` passes

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> Touches consensus-critical block cost limits and generator assembly; incorrect bounds could reject valid txs or overfill blocks, though tests assert parity with the INTERNED_GENERATOR validation path.
> 
> **Overview**
> Adds **`InternedBlockBuilder`**, a mempool-style block assembler for the post-HF2 **`INTERNED_GENERATOR`** cost model, alongside the existing compressed **`BlockBuilder`**.
> 
> It greedily packs prioritized spend bundles under `max_block_cost_clvm` using incremental accounting: callers pass **CLVM execution + conditions cost only** (no generator byte cost). Per-spend interned vbytes are estimated in a scratch allocator (`spend_vbytes`); the running sum is a **safe upper bound** via the triangle inequality. `cost()` exposes that bound; **`finalize()`** builds the quoted `(q . (spend_list))` generator, serializes with `node_to_bytes_backrefs`, re-interns from the root for the **exact** consensus cost (aligned with `run_block_generator2` + `INTERNED_GENERATOR`), and aggregates BLS signatures. Rejection uses the same skip cap and near-full heuristics as the compressed builder.
> 
> **Python / wheel:** `InternedBlockBuilder` is registered in `wheel/src/api.rs` with stubs in `chia_rs.pyi` and `generate_type_stubs.py` (`ConsensusConstants` at `__new__`; `finalize()` without extra constants). The module is exported from `chia-consensus` via `build_interned_block`.
> 
> **Tests** cover upper-bound ≥ exact cost, consensus cost parity, limits/skips, and an optional heavy integration test over `test-bundles`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 030aacb69165bc65a07fc5534076bb6d4d967206. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


richardkiss:
The `Generator performance (3.13)` failure is pre-existing flakiness unrelated to this PR: `single-coin-only-garbage.txt` ran at 10.09s against a limit of 10s on a loaded CI runner. All other checks pass.

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,203 @@
...
+    Done,
+}
+
+fn result(num_skipped: u32) -> BuildBlockResult {
+    if num_skipped > MAX_SKIPPED_ITEMS {
+        BuildBlockResult::Done
+    } else {
+        BuildBlockResult::KeepGoing
+    }
+}
```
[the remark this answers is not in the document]
richardkiss: Fixed in 7457a45c.

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,203 @@
...
+        let (added, result) = self.add_spend_bundles(
+            bundles.iter().map(|item| {
+                item.extract::<Bound<'_, SpendBundle>>()
+                    .expect("spend bundle")
+                    .get()
+                    .clone()
+            }),
+            cost,
+            constants,
+        )?;
```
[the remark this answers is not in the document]
richardkiss: Fixed in 7457a45c.

crates/chia-consensus/src/build_interned_block.rs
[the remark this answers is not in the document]
richardkiss: Fixed in 7457a45c.

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,203 @@
...
+            generator_cost: 0,
+            num_skipped: 0,
+        })
+    }
+
+    fn compute_generator_cost(allocator: &mut Allocator, spend_list: NodePtr) -> Result<u64> {
+        // Build (q . ((spend_list)))
+        let inner = allocator.new_pair(spend_list, allocator.nil())?;
+        let outer = allocator.new_pair(allocator.one(), inner)?;
+        let interned = intern_tree_limited(allocator, outer, usize::MAX)?;
```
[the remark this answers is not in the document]
richardkiss: Fixed in 7457a45c.

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,203 @@
...
+                let amount = a.new_number(spend.coin.amount.into())?;
+                let item = a.new_pair(amount, item)?;
+                let puzzle = node_from_bytes_backrefs(a, spend.puzzle_reveal.as_ref())?;
+                let item = a.new_pair(puzzle, item)?;
+                let parent_id = a.new_atom(&spend.coin.parent_coin_info)?;
+                let item = a.new_pair(parent_id, item)?;
+                self.spend_list = a.new_pair(item, self.spend_list)?;
+            }
+            cumulative_signature.aggregate(&bundle.borrow().aggregated_signature);
+        }
```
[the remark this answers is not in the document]
richardkiss: Fixed in 7457a45c.

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,475 @@
...
+                    generator.as_slice(),
+                    [],
+                    TEST_CONSTANTS.max_block_cost_clvm,
+                    MEMPOOL_MODE,
+                    &signature,
+                    None,
+                    &TEST_CONSTANTS,
+                )
+                .expect("run_block_generator2");
+                assert_eq!(conds.cost, cost);
```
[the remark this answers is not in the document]
richardkiss: Fixed in 7457a45c.

arvidn:
> The Generator performance (3.13) failure is pre-existing flakiness unrelated to this PR: single-coin-only-garbage.txt ran at 10.09s against a limit of 10s on a loaded CI runner. 

We run benchmarks on self-hosted runners that only run one job at a time to avoid being loaded. I don't believe this explanation.

I suppose this PR doesn't affect running generators though, does it?

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,460 @@
...
+        })
+    }
+
+    fn compute_generator_cost(
+        allocator: &mut Allocator,
+        spend_list: NodePtr,
+        cost_per_byte: u64,
+    ) -> Result<u64> {
+        // Build (q . ((spend_list)))
+        let inner = allocator.new_pair(spend_list, allocator.nil())?;
```
arvidn:
```suggestion
        let inner = allocator.new_pair(spend_list, NodePtr::NIL)?;
```

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,460 @@
...
+            generator_cost: 0,
+            num_skipped: 0,
+        })
+    }
+
+    fn compute_generator_cost(
+        allocator: &mut Allocator,
+        spend_list: NodePtr,
+        cost_per_byte: u64,
+    ) -> Result<u64> {
```
arvidn: I think it's unexpected that a function that *computes cost* allocates new nodes in the `Allocator`. I would expect the `Allocator` be be a reference to immutable.

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,460 @@
...
+        constants: &ConsensusConstants,
+    ) -> Result<(bool, BuildBlockResult)>
+    where
+        T: IntoIterator<Item = S>,
+        S: Borrow<SpendBundle>,
+    {
+        if self.generator_cost + self.block_cost + MIN_COST_THRESHOLD
+            > constants.max_block_cost_clvm
+        {
+            self.num_skipped += 1;
```
arvidn: technically, we don't need to increment `num_skipped` here. We're not skipping because of a property of the spend bundle, but the block being full

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,460 @@
...
+            &mut self.allocator,
+            self.spend_list,
+            constants.cost_per_byte,
+        )?;
+
+        if new_generator_cost + self.block_cost + cost > constants.max_block_cost_clvm {
+            // Restore: the allocator is not reset (dead nodes are acceptable).
+            self.spend_list = saved_spend_list;
+            self.num_skipped += 1;
+            return Ok((false, skip_result(self.num_skipped)));
```
arvidn: I think it might be good to restore the Allocator from a checkpoint from before trying to add these spend bundles too. I can imagine a worst-case of large puzzles that keep being rejected, accumulating wasted space in the allocator

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,460 @@
...
+        // self.spend_list untouched and consistent with the tracked costs.
+        let mut local_spend_list = saved_spend_list;
+        let mut cumulative_signature = Signature::default();
+        for bundle in bundles {
+            for spend in &bundle.borrow().coin_spends {
+                let solution = node_from_bytes_backrefs(a, spend.solution.as_ref())?;
+                let item = a.new_pair(solution, NodePtr::NIL)?;
+                let amount = a.new_number(spend.coin.amount.into())?;
+                let item = a.new_pair(amount, item)?;
+                let puzzle = node_from_bytes_backrefs(a, spend.puzzle_reveal.as_ref())?;
```
arvidn, line 177: is it safe to support backrefs in puzzles and solutions? I think there's a risk the full node would be vulnerable to a decompression-bomb if we do

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,460 @@
...
+            }
+            cumulative_signature.aggregate(&bundle.borrow().aggregated_signature);
+        }
+        self.spend_list = local_spend_list;
+
+        let new_generator_cost = Self::compute_generator_cost(
+            &mut self.allocator,
+            self.spend_list,
+            constants.cost_per_byte,
+        )?;
```
arvidn: This makes the cost of building the block O(n^2). It seems risky. I suppose computing the cost incrementally isn't trivial, since you would need to know about atom matches existing ones.

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,460 @@
...
+
+        if new_generator_cost + self.block_cost + cost > constants.max_block_cost_clvm {
+            // Restore: the allocator is not reset (dead nodes are acceptable).
+            self.spend_list = saved_spend_list;
```
arvidn: rather than updating and the restoring `self.spend_list`, could you just compute the cost over `local_spend_list` and only update `self.spend_list` if we pass the cost check? It would seem a bit simpler and robust.

crates/chia-consensus/src/build_interned_block.rs
richardkiss: Applied.

crates/chia-consensus/src/build_interned_block.rs
richardkiss: Fixed — removed the num_skipped increment on the block-full path.

crates/chia-consensus/src/build_interned_block.rs
richardkiss: The recent fix (commit 7457a45c) already does this — local_spend_list accumulates throughout the loop and self.spend_list is only updated after the loop completes. The saved_spend_list / restore path is now only used for the cost-limit rejection below.

crates/chia-consensus/src/build_interned_block.rs
richardkiss: Fixed in 0331bf2a — compute_generator_cost is gone. The upper-bound approach (raw_vbytes walk) is pure reads on existing allocator nodes; the exact intern call (now named exact_generator_cost) interns spend_list directly and adds the 11-vbyte wrapper constant, so no wrapper nodes are ever allocated.

(AI wrote this comment, I'm still catching up. -- RK)

crates/chia-consensus/src/build_interned_block.rs
richardkiss: Fixed in 0331bf2a — O(n²) eliminated. The fast path now uses a running raw_vbytes upper bound (O(spend_size) per add, no intern call). intern_tree_limited is only invoked when the upper bound approaches the block limit, and always in finalize().

crates/chia-consensus/src/build_interned_block.rs
richardkiss: node_from_bytes_backrefs can't produce a decompression bomb. When a backreference is encountered it resolves to an existing NodePtr via path traversal — no new node is allocated. Pair construction is always allocator.new_pair(left, right) on two existing NodePtrs. So the number of allocator nodes is bounded by the number of unique atoms and pairs in the serialized blob, which is at most proportional to its byte length. This is also what BlockBuilder uses for the same purpose.

Same behavior as BlockBuilder — no change needed here.

richardkiss: > `node_from_bytes_backrefs` can't produce a decompression bomb. Backreferences resolve to existing `NodePtr`s — no new nodes are allocated. Pair construction is always `new_pair(left, right)` on two existing pointers. The allocator growth is bounded by the number of unique atoms and pairs in the serialized blob, which is at most proportional to its byte length.
>
> This is also the same code path as `BlockBuilder`, which uses `node_from_bytes_backrefs` identically.

(AI-assisted, approved by me.)

crates/chia-consensus/src/build_interned_block.rs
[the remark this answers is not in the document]
richardkiss: Intentional — `BlockBuilder` (pre-HF2) and `BlockBuilder2026` (post-HF2, renamed in ced30dfb) are parallel implementations with separate lifetimes. `BlockBuilder` will be deleted after the hard fork activates. Sharing a common module would complicate that deletion, so each file defines its own copy. The duplication is load-bearing.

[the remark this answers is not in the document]
richardkiss: Intentional — `BlockBuilder` (pre-HF2) and `BlockBuilder2026` (post-HF2, renamed in ced30dfb) are parallel implementations with separate lifetimes. `BlockBuilder` will be deleted after the hard fork activates. Sharing a common module would complicate that deletion, so each file defines its own copy. The duplication is load-bearing.

crates/chia-consensus/src/build_interned_block.rs
arvidn: I'm pretty sure the full node explicitly does not support back-refs for puzzles and solutions to avoid decompression bombs. I think it's a mistake to support it here, as it can raise questions of what's correct. With the current protocol and features surrounding puzzles and solutions, serializations with back-refs should be rejected.

The problem isn't that *this* call would cause a decompression bomp, the problem would be that this code makes it look like perhaps there's a mistake in `chia-blockchain` to not support backrefs. If someone changes it in the future we'll have a decompression bomb in the get_puzzle_and_solution_for_coin RPC.

So, it's (most likely) true that it doesn't matter, because back-ref puzzles and solutions won't make it here anyway. But it *does* matter in the sense that it casts doubt on what the current protocol is.

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -37,18 +68,47 @@ impl InternedBlockBuilder {
...
+                }
+                SExp::Pair(left, right) => {
+                    pair_count += 1;
+                    stack.push(left);
+                    stack.push(right);
+                }
+            }
+        }
+        atom_bytes + 2 * atom_count + 3 * pair_count
+    }
```
arvidn: this looks like it should be a free function. i.e. since it doesn't depend on `self`, it might as well not be part of `BlockBuilder2026`

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -37,18 +68,47 @@ impl InternedBlockBuilder {
...
+    /// does when it calls `intern_tree_limited` on the deserialized program.
+    /// The two wrapper nodes are dead after this call but harmless.
+    fn exact_generator_cost(
         allocator: &mut Allocator,
         spend_list: NodePtr,
         cost_per_byte: u64,
     ) -> Result<u64> {
-        // Build (q . ((spend_list)))
-        let inner = allocator.new_pair(spend_list, allocator.nil())?;
+        let inner = allocator.new_pair(spend_list, NodePtr::NIL)?;
```
arvidn: you could checkpoint and restore the allocator here, so you don't "leak" the quote and pair

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -66,24 +126,26 @@ impl InternedBlockBuilder {
         T: IntoIterator<Item = S>,
         S: Borrow<SpendBundle>,
     {
-        if self.generator_cost + self.block_cost + MIN_COST_THRESHOLD
+        self.cost_per_byte = constants.cost_per_byte;
```
arvidn: is there a good reason to not pass this in to the constructor instead?

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,485 @@
...
+        T: IntoIterator<Item = S>,
+        S: Borrow<SpendBundle>,
+    {
+        self.cost_per_byte = constants.cost_per_byte;
+
+        if self.upper_bound_vbytes * constants.cost_per_byte + self.block_cost + MIN_COST_THRESHOLD
+            > constants.max_block_cost_clvm
+        {
+            // Block is full regardless of what bundle we try next.
+            return Ok((false, BuildBlockResult::Done));
```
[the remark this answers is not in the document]
arvidn: Isn't this the place where you would recompute `self.block_cost` with the more accurate de-duplication number?

crates/chia-consensus/src/build_interned_block.rs
arvidn: no, it didn't. You're still restoring `self.spend_list` here. `saved_spend_list` should be removed along with this line

crates/chia-consensus/src/build_interned_block.rs
arvidn: I see that the existing `BlockBuilder` does the same thing, which I don't think is right (but not a bug per-se)

crates/chia-consensus/src/build_interned_block.rs
arvidn: I see line 155 still say `self.num_skipped += 1;` I don't think it really matters if you prefer to keep it

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,709 @@
...
+    }
+
+    /// Compute the interned vbyte cost of a single spend in isolation.
+    /// Interns the spend tuple in a scratch allocator and measures its
+    /// interned_vbytes, which is an upper bound on this spend's contribution
+    /// when added to the full generator (triangle inequality).
+    fn spend_vbytes(spend: &chia_protocol::CoinSpend) -> Result<u64> {
+        let mut a = Allocator::new();
+        // solution
+        let solution = node_from_bytes_backrefs(&mut a, spend.solution.as_ref())?;
```
arvidn, line 111:
```suggestion
        let solution = node_from_bytes(&mut a, spend.solution.as_ref())?;
```

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,709 @@
...
+        let item = a.new_pair(solution, NodePtr::NIL)?;
+        // amount
+        let amount = a.new_number(spend.coin.amount.into())?;
+        let item = a.new_pair(amount, item)?;
+        // puzzle reveal
+        let puzzle = node_from_bytes_backrefs(&mut a, spend.puzzle_reveal.as_ref())?;
```
arvidn, line 117:
```suggestion
        let puzzle = node_from_bytes(&mut a, spend.puzzle_reveal.as_ref())?;
```

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,709 @@
...
+            .expect("add_spend_bundles");
+        assert!(added);
+
+        let upper_bound = builder.cost();
+        let (_, _, exact_total) = builder.finalize().expect("finalize");
+
+        assert!(
+            upper_bound >= exact_total,
+            "upper bound {upper_bound} should be >= exact {exact_total}"
+        );
```
arvidn: I think you should run the spend bundle and add up the total cost of the generator and ensure that it exactly matches the cost reported by the builder. This is the cost that we'll put in the block header, right? it has to match exactly.

just like `test_build_interned_block()` does.

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,709 @@
...
+
+    #[test]
+    fn test_generator_cost_accuracy() {
+        // Verify that the upper-bound estimate is always >= the exact cost,
+        // and that finalize() returns the correct exact cost.
+        let mut builder = InternedBlockBuilder::new(&TEST_CONSTANTS).expect("new builder");
+
+        let file = "../../test-bundles/e003f780f1bf036bfa3df7eed6b0e480c2dc3e9d6b1f8c3aeeb542e9da08e8d4.bundle";
+        if !std::path::Path::new(file).exists() {
+            return;
```
arvidn: this should fail the test, don't you think?

arvidn (commented):
I think it would be good to have a fuzzer as well, something like this: crates/chia-consensus/fuzz/fuzz_targets/solution-generator.rs

crates/chia-consensus/src/build_interned_block.rs
richardkiss: This is the same as in `build_compressed_block.py`.

crates/chia-consensus/src/build_interned_block.rs
richardkiss: Done.

crates/chia-consensus/src/build_interned_block.rs
richardkiss: This is outdated, resolving.

richardkiss:
> crates/chia-consensus/fuzz/fuzz_targets/solution-generator.rs

This is done now in `chia_rs/interned-block-builder/crates/chia-consensus/fuzz/fuzz_targets/interned-block-builder.rs`

crates/chia-consensus/src/build_interned_block.rs
richardkiss: This is fixed now (and it now points at a file that's checked in too, ha ha).

crates/chia-consensus/src/build_interned_block.rs
richardkiss: Done. See `crates/chia-consensus/src/build_interned_block/additional_tests.rs`

crates/chia-consensus/src/build_interned_block.rs
richardkiss: I've added a lot more tests here. One bundle's not enough. Now it does all 90 in test-bundles (skipping the ones with short names).

crates/chia-consensus/src/build_interned_block.rs
richardkiss: This is also copied direct from `build_compressed_block.py`. It's just decompressing a potentially back-ref serialization from the mempool, so no significant additional time or consensus concerns.

crates/chia-consensus/src/build_interned_block/additional_tests.rs
```diff
@@ -0,0 +1,342 @@
...
+        assert!(
+            result == BuildBlockResult::Done,
+            "should signal Done when close to limit"
+        );
+    } else {
+        assert!(
+            result == BuildBlockResult::Done,
+            "should be Done if bundle doesn't fit"
+        );
+    }
```
arvidn: This looks odd to me. this test is deterministic, right? only one of these branches are ever executed, as far as I can tell.

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -57,11 +54,11 @@ pub struct InternedBlockBuilder {
...
 
     // the cost of the block we've built up so far, not including the byte-cost.
-    // That's separated out into the byte_cost_ub member.
+    // That's seprated out into the byte_cost member.
     block_cost: u64,
 
-    // the upper-bound byte cost, so far, from per-spend isolated interning
-    byte_cost_ub: u64,
+    // the byte cost, so for, of what's in the generator (upper-bound estimate)
+    byte_cost: u64,
```
arvidn, line 61: as far as I can tell, `byte_cost_ub` changed name, but not semantics. right?

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,490 @@
...
+/// running sum is an upper bound on the true interned cost of all spends
+/// combined. finalize() computes the exact cost.
+#[cfg_attr(feature = "py-bindings", pyclass)]
+pub struct InternedBlockBuilder {
+    allocator: Allocator,
+    signature: Signature,
+    spend_list: NodePtr,
+
+    // the cost of the block we've built up so far, not including the byte-cost.
+    // That's seprated out into the byte_cost member.
```
arvidn:
```suggestion
    // That's separated out into the byte_cost member.
```

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -124,15 +121,15 @@ impl InternedBlockBuilder {
...
+        Ok(interned_vbytes(&interned) + COST_CONS)
     }
 
     /// add a batch of spend bundles to the generator. The cost for each bundle
     /// must be *only* the CLVM execution cost + the cost of the conditions.
     /// It must not include the byte cost of the bundle. The byte cost is
-    /// unpredictable as the generator is being compressed. The true byte cost
-    /// is computed by this function. Returns true if the bundles could be added
+    /// unpredictible as the generator is being / compressed. The true byte cost
+    /// is computed by this function. / returns true if the bundles could be added
```
arvidn: not really important, but these `/` seems accidentally inserted, also "unpredictable" was correctly spelled previously.

crates/chia-consensus/src/build_interned_block.rs
arvidn: I'll resolve this comment, it's not important, but it looks like you intended to remove this.

crates/chia-consensus/src/build_interned_block.rs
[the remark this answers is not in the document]
arvidn: this question has been left hanging. This seems like an important issue, if it isn't, please explain it here

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -276,15 +272,16 @@ impl InternedBlockBuilder {
     /// generate the block generator
     #[pyo3(name = "finalize")]
     pub fn py_finalize(&mut self) -> PyResult<(Vec<u8>, Signature, u64)> {
-        // PyO3 doesn't allow consuming self, so we swap in a dummy and finalize
-        // the original.
```
arvidn: this comment seems quite useful still

crates/chia-consensus/src/build_interned_block.rs
```diff
@@ -0,0 +1,490 @@
...
+        self.cost()
+    }
+
+    /// generate the block generator
+    #[pyo3(name = "finalize")]
+    pub fn py_finalize(&mut self) -> PyResult<(Vec<u8>, Signature, u64)> {
+        let mut temp = InternedBlockBuilder::new_with(self.cost_per_byte, self.max_block_cost);
+        std::mem::swap(self, &mut temp);
+        let (generator, sig, cost) = temp.finalize()?;
+        Ok((generator, sig, cost))
```
arvidn: I think the cursor comment is quite reasonable here.


```suggestion
        match temp.finalize() {
           Ok(x) => x,
           Err(err) => {
              std::mem::swap(self, &mut temp);
              Err(err)
           }
        }
```

richardkiss:
> > The Generator performance (3.13) failure is pre-existing flakiness unrelated to this PR: single-coin-only-garbage.txt ran at 10.09s against a limit of 10s on a loaded CI runner.
> 
> We run benchmarks on self-hosted runners that only run one job at a time to avoid being loaded. I don't believe this explanation.
> 
> I suppose this PR doesn't affect running generators though, does it?

It does not. This PR should not change consensus in any way, even if it were imported by chia-blockchain. It's all new, opt-in functionality.

crates/chia-consensus/src/build_interned_block.rs
richardkiss: I called it `_ub` for a while to emphasize it's "upper-boundedness", but reverted to make the diffs smaller.

richardkiss: The semantics are different though; it now keeps the "upper bound" cost, since it's much cheaper than keeping the exact cost. The exact cost is calculated at the end.

crates/chia-consensus/src/build_interned_block.rs
richardkiss: These problems were copied directly from `build_compressed_block.rs`. They're fixed here, but I think fixing them there is a bit out of scope. That file's days could be numbered anyone, as it could be removed after HF2.

crates/chia-consensus/src/build_interned_block.rs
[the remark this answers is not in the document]
richardkiss: Sorry, I missed this. Cursor is right, this feature is not implemented yet. I wanted to keep this PR as simple as I could so we have a fast path that will work well with mostly empty blocks. We can add this capability as an opt-in later. I figure this is the safest way: something simple that works, then adventurous people trying to optimize can try to improve throughput/fees by trying new ideas.

crates/chia-consensus/src/build_interned_block.rs
richardkiss: I merged this.

crates/chia-consensus/src/build_interned_block/additional_tests.rs
richardkiss: This test has been reworked to take better advantage of the determinism.

crates/chia-consensus/src/build_interned_block.rs
arvidn: it looked like it always had that semantics, and the last diff I reviewed didn't change the comment. it looks like you changed it back now though.

arvidn (approved):

PR #1437, opened 2026-05-09
Wire serde_2026 serialization into chia_rs

richardkiss:
## Summary

Wire the serde_2026 serialization format into chia_rs as a foundation for the post-HF2 `INTERNED_GENERATOR` cost model. This PR ships the producer/consumer surface only — nothing here changes consensus behavior until #1438 wires it into validation.

Depends on clvmr ≥ 0.18.0 (serde_2026 shipped to crates.io in June), which `main` already uses — **no Cargo.toml changes**.

## Changes

- `chia_consensus::serde_2026::node_from_bytes_auto`: deserialize CLVM bytes, auto-detecting classic / backrefs / serde_2026 by magic prefix. Callers supply the max blob size; the per-atom cap (1 MiB) is fixed. Accepts non-minimal encodings (`strict = false`) — rationale documented inline: post-HF2, identity and cost come from the interned tree, so overlong varints only hurt their sender.
- `chia_consensus::serde_2026::max_canonical_blob_size(max_cost, cost_per_byte)`: the provable upper bound on the canonical serde_2026 wire size of any within-cost generator (`max_cost / cost_per_byte + 5 + prefix`). The proof is in its doc comment; at mainnet constants the bound is 916,677 bytes. This is what callers should derive the `node_from_bytes_auto` size cap from — the limit is a property of the cost constants, not a hardcoded number. How much headroom to allow for non-minimal encodings is caller policy (input welcome).
- `solution_generator_2026()`: builds a generator using `serialize_2026`, at a named `SERDE_2026_COMPRESSION_LEVEL` (level 0; levels are producer-side only, all decodable by the one deserializer).
- Python bindings: `solution_generator_2026()` and `SERDE_2026_MAGIC_PREFIX`, stub generator updated.

## Tests

- Unit tests for auto-dispatch (all three formats), size-cap boundary, and `solution_generator_2026` round-trip through `node_from_bytes_auto`.
- The size bound is tested at its adversarial tight shapes (max-length atom — exactly tight, slack 0; >63 distinct atom lengths; >4096 unique pairs; doubling DAG) and shown general across arbitrary cost constants, from `(0, 7)` to `(u64::MAX, u64::MAX)` — not just mainnet's 11e9 / 12,000.
- New fuzz target `serde-2026-size-bound`: arbitrary trees *and* arbitrary cost constants per run; asserts the encoding bound, the derived cap for affordable trees, and round-trip identity. A deterministic seed generator (`gen_serde_2026_fuzz_seeds` example) produces corpus entries at 0.98–0.999 of the bound so mutation starts at the boundary. Fuzzing already earned its keep: it caught an integer overflow in the bound function at extreme constants (fixed with saturation).

## Depends on

#1436 (`InternedBlockBuilder`) — merged; this PR is based on current `main`.

## Follow-up (#1438 / FullBlock #1456)

Arvid's merged FullBlock format (#1456) adds v1 blocks that store the generator as a raw length-prefixed `Vec<u8>` (`transactions_generator_buffer`), giving a place to inspect bytes for `SERDE_2026_MAGIC_PREFIX` before HF2 and reject serde_2026 generators while `INTERNED_GENERATOR` is inactive — without parsing CLVM first. This PR does not wire that guard; #1438 adds `check_generator_quote` / consensus-path checks, applying the prefix check to both legacy `Program` bytes and the v1 raw buffer path.

## Test plan

- [x] fmt / clippy / full test suite pass locally
- [x] `from chia_rs import solution_generator_2026, SERDE_2026_MAGIC_PREFIX` works
- [x] `solution_generator_2026(...)` output starts with `SERDE_2026_MAGIC_PREFIX` (unit-tested)
- [x] Fuzz target: ~750k executions clean
- [ ] CI green on current head

---

> [!NOTE]
> **Medium Risk**
> New deserialization and provable size-bound logic touch generator wire format and pre-charge decoding limits; behavior is not yet on the consensus path but mistakes here would matter for the HF2 follow-up.
> 
> **Overview**
> Adds **serde_2026** as a producer/consumer surface ahead of post-HF2 interned-generator work, without changing live consensus validation in this PR.
> 
> New **`chia_consensus::serde_2026`** exposes **`node_from_bytes_auto`** (magic-prefix dispatch to classic/backrefs/serde_2026 with caller-supplied **`max_blob_size`**, non-minimal encodings allowed via **`strict = false`**), **`max_canonical_blob_size`** (cost-derived wire cap, ~916 677 bytes at mainnet constants), and **`SERDE_2026_COMPRESSION_LEVEL`**. **`solution_generator_2026`** builds generators with **`serialize_2026`**.
> 
> Python **`chia_rs`** gains **`solution_generator_2026`** and **`SERDE_2026_MAGIC_PREFIX`** (stubs updated). Coverage includes unit tests on tight encoding shapes and format dispatch, a **`serde-2026-size-bound`** fuzz target, and **`gen_serde_2026_fuzz_seeds`** to seed near-bound corpora.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2cd665e561a13ff3dfd3373066399e142f380924. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/chia-consensus/src/solution_generator.rs
```diff
@@ -106,6 +108,16 @@ where
...
 
+pub fn solution_generator_2026<BufRef, I>(spends: I) -> Result<Vec<u8>>
+where
+    BufRef: AsRef<[u8]>,
+    I: IntoIterator<Item = (Coin, BufRef, BufRef)>,
+{
+    let mut a = Allocator::new();
+    let generator = build_generator(&mut a, spends)?;
+    Ok(serialize_2026(&a, generator, 0)?)
+}
```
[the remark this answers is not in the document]
richardkiss: Done

crates/chia-consensus/src/solution_generator.rs
```diff
@@ -106,6 +108,16 @@ where
     Ok(node_to_bytes_backrefs(&a, generator)?)
 }
 
+pub fn solution_generator_2026<BufRef, I>(spends: I) -> Result<Vec<u8>>
+where
+    BufRef: AsRef<[u8]>,
+    I: IntoIterator<Item = (Coin, BufRef, BufRef)>,
+{
```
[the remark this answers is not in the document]
richardkiss: Done

wheel/generate_type_stubs.py
```diff
@@ -322,6 +322,7 @@ class _Unspec:
 
 def solution_generator(spends: Sequence[tuple[Coin, bytes, bytes]]) -> bytes: ...
 def solution_generator_backrefs(spends: Sequence[tuple[Coin, bytes, bytes]]) -> bytes: ...
+def solution_generator_2026(spends: Sequence[tuple[Coin, bytes, bytes]]) -> bytes: ...
```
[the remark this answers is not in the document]
richardkiss: This is done.

wheel/python/chia_rs/chia_rs.pyi
```diff
@@ -132,6 +133,7 @@ COMPUTE_FINGERPRINT: int = ...
 COST_CONDITIONS: int = ...
 SIMPLE_GENERATOR: int = ...
 LIMIT_SPENDS: int = ...
+SERDE_2026_MAGIC_PREFIX: bytes = ...
 DISABLE_OP: int = ...
```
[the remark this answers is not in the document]
richardkiss: Done.

crates/chia-consensus/src/serde_2026.rs
```diff
@@ -0,0 +1,301 @@
...
+///
+/// Combining: `wire_size <= max_cost / cost_per_byte + 5 + prefix_len`.
+///
+/// At mainnet constants (`max_cost` = 11e9, `cost_per_byte` = 12_000) this
+/// is 916_677 bytes, a little under 1 MiB.
+pub fn max_canonical_blob_size(max_cost: u64, cost_per_byte: u64) -> usize {
+    // Saturating: a clamped result is still a correct upper bound, and this
+    // keeps the function total for extreme (non-mainnet) constants.
+    ((max_cost / cost_per_byte) as usize).saturating_add(5 + SERDE_2026_MAGIC_PREFIX.len())
+}
```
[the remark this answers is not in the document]
richardkiss: Fixed in 80131ead: `cost_per_byte == 0` now returns `usize::MAX` — free bytes mean no size is over-cost, so the least upper bound is genuinely unbounded (returning 0 would reject blobs the constants permit). Covered by new unit cases `(11e9, 0)` and `(0, 0)`, and the fuzz harness no longer clamps `cost_per_byte` away from zero.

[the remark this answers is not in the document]
richardkiss: This is fixed.

crates/chia-consensus/examples/gen_serde_2026_fuzz_seeds.rs
```diff
@@ -0,0 +1,95 @@
...
+impl XorShift {
+    fn next(&mut self) -> u64 {
+        let mut x = self.0;
+        x ^= x << 13;
+        x ^= x >> 7;
+        x ^= x << 17;
+        self.0 = x;
+        x
+    }
+}
```
arvidn: it seems a bit unnecessary to implement your own RNG

crates/chia-consensus/src/serde_2026.rs
```diff
@@ -0,0 +1,312 @@
...
+/// dispatches to [`deserialize_2026`] with the per-atom consensus cap
+/// ([`CONSENSUS_MAX_ATOM_LEN`]). Otherwise falls back to
+/// [`node_from_bytes_backrefs`] (which also accepts plain classic).
+///
+/// `max_blob_size` bounds the total wire size accepted; blobs above it are
+/// rejected before any parsing. Callers should derive it from the network's
+/// cost constants via [`max_canonical_blob_size`] (any headroom multiplier
+/// on top — e.g. to tolerate non-minimal encodings, which `strict = false`
+/// otherwise admits — is caller policy).
+pub fn node_from_bytes_auto(
```
arvidn, line 96: do you have a use case for this in mind?
I expect that we'll switch to this encoding in the hard fork, as the mandatory way of encoding generators. There wouldn't be any need to add extra bytes identifying which encoding the block is using.

On the other hand, with the new consensus, how we store blocks in the database becomes an implementation detail, so these bytes are only part of the network protocol

crates/chia-consensus/src/serde_2026.rs
```diff
@@ -0,0 +1,312 @@
...
+use clvmr::allocator::{Allocator, NodePtr};
+use clvmr::error::{EvalErr, Result};
+use clvmr::serde::{SERDE_2026_MAGIC_PREFIX, deserialize_2026, node_from_bytes_backrefs};
+
+/// Per-atom byte cap used by chia consensus when deserializing CLVM blobs.
+///
+/// Matches the historical `clvm_rs` default. CLVM atoms above this size are
+/// uneconomical to construct under cost limits anyway; this is a defensive
+/// pre-allocation cap, not a hard consensus rule.
+pub const CONSENSUS_MAX_ATOM_LEN: usize = 1 << 20;
```
arvidn: this sounds like it should be part of `ConsensusConstants`, should it not?

crates/chia-consensus/src/solution_generator.rs
```diff
@@ -106,6 +109,16 @@ where
     Ok(node_to_bytes_backrefs(&a, generator)?)
 }
 
+pub fn solution_generator_2026<BufRef, I>(spends: I) -> Result<Vec<u8>>
```
arvidn, line 112: Is there a use case for this function? We already have the `InternedBlockBuilder`, which is how I would expect the mempool would build blocks. When and where would we use this?

crates/chia-consensus/src/serde_2026.rs
richardkiss: We can require 2026 serialization if that's what you're up for. It's nice to leave the magic header in place though so extraction tools don't have to add it back on their on (the clvm file format requires it for tooling, for graceful failure on legacy tools and easy-to-spot text to hint as to why the legacy tools are failing).

crates/chia-consensus/examples/gen_serde_2026_fuzz_seeds.rs
richardkiss: I'll switch to a seeded SmallRng.

crates/chia-consensus/src/serde_2026.rs
richardkiss: We can actually calculate this as a function of the consensus cost constants, using `max_canonical_blob_size()`. It returns a number `N` where any clvm program with cost under the max can be serialized as a 2026 blob in at most `N` bytes. And since atoms appear as literals in the serialization, an atom longer than `N` can't be part of any affordable program — so `N` doubles as a guaranteed upper bound on atom length. Done in 9a438ab3.

crates/chia-consensus/src/solution_generator.rs
richardkiss: The only potential consumer really is `block_tools`. If you think we can switch it over to using `InternedBlockBuilder`, I can take this out. Your call.

arvidn (approved):

PR #1438, opened 2026-05-09
[CHIA-3823] Add INTERNED_GENERATOR consensus path + serde_2026 prefix guard

richardkiss:
## Summary

Implement the `INTERNED_GENERATOR` consensus path in `run_block_generator2`.

We also tweak `check_generator_quote` to skip checking serde_2026-serialized generators because it's more difficult to tell if they start with a `q`. We still check in `check_generator_node`.

---

> [!NOTE]
> **High Risk**
> This is consensus-critical generator serialization and validation; a mismatch with network rules would fork or reject valid blocks.
> 
> **Overview**
> With **`INTERNED_GENERATOR`**, block generators are **serde_2026-only**: **`InternedBlockBuilder::finalize`** now writes **`serialize_2026`** instead of backrefs, and **`run_block_generator2`** / trusted-block coinspend helpers decode via **`node_from_bytes_2026`** with a **`max_canonical_blob_size`** cap, then charge byte cost from the **interned** tree instead of raw wire length.
> 
> **`check_generator_quote`** no longer inspects the `[0xff, 0x01]` prefix when the interned flag is set (serde_2026 bytes are not quote-shaped on the wire); **simple-generator “quoted” shape** is still enforced in **`check_generator_node`** after decode. Pre-fork behavior is unchanged: serde_2026 blobs fail parse or quote checks without the flag, and classic generators fail once **`INTERNED_GENERATOR`** is on.
> 
> Adds **`node_from_bytes_2026`** in **`serde_2026`**, fuzz round-trip coverage with **`INTERNED_GENERATOR`**, and unit tests for cross-encoding acceptance/rejection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 616e4863bf9d0933bff364912b6e303f55dba844. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/chia-consensus/src/run_block_generator.rs
+++ b/crates/chia-consensus/src/run_block_generator.rs
@@ -25,10 +25,7 @@ use clvmr::chia_dialect::ChiaDialect;
 use clvmr::cost::Cost;
 use clvmr::reduction::Reduction;
 use clvmr::run_program::run_program;
-use clvmr::serde::{
-    InternedTree, SERDE_2026_MAGIC_PREFIX, intern_tree_limited, node_from_bytes,
-    node_from_bytes_backrefs,
-};
+use clvmr::serde::{InternedTree, intern_tree_limited, node_from_bytes, node_from_bytes_backrefs};
 
 pub fn subtract_cost(cost_left: &mut Cost, subtract: Cost) -> Result<(), ValidationErr> {
     if subtract > *cost_left {
@@ -175,17 +172,6 @@ fn extract_n<const N: usize>(
 // this is required after the SIMPLE_GENERATOR fork is active
 #[inline]
 pub fn check_generator_quote(program: &[u8], flags: ConsensusFlags) -> Result<(), ValidationErr> {
-    // CRITICAL: Reject serde_2026 format before INTERNED_GENERATOR activates.
-    // Without this explicit guard, a serde_2026-prefixed generator would be
-    // accepted on the pre-HF2 chain during the window between code shipping
-    // and SIMPLE_GENERATOR (soft_fork9) activation.
-    if !flags.contains(ConsensusFlags::INTERNED_GENERATOR)
-        && program.starts_with(&SERDE_2026_MAGIC_PREFIX)
-    {
-        return Err(ValidationErr::Err(
-            ErrorCode::InvalidTransactionsGeneratorEncoding,
-        ));
-    }
     if flags.contains(ConsensusFlags::INTERNED_GENERATOR) {
         // serde_2026 blocks have a different serialization header
         return Ok(());
@@ -363,9 +349,13 @@ where
     check_generator_quote(generator.as_ref(), flags)?;
     let mut output = Vec::<CoinSpend>::new();
 
-    let max_blob_size =
-        max_canonical_blob_size(constants.max_block_cost_clvm, constants.cost_per_byte);
-    let program = node_from_bytes_auto(&mut a, generator, max_blob_size)?;
+    let program = if flags.contains(ConsensusFlags::INTERNED_GENERATOR) {
+        let max_blob_size =
+            max_canonical_blob_size(constants.max_block_cost_clvm, constants.cost_per_byte);
+        node_from_bytes_auto(&mut a, generator, max_blob_size)?
+    } else {
+        node_from_bytes_backrefs(&mut a, generator)?
+    };
     check_generator_node(&a, program, flags)?;
     let args = setup_generator_args(&mut a, refs, flags)?;
     let dialect = ChiaDialect::new(flags.to_clvm_flags());
@@ -464,9 +454,13 @@ where
     check_generator_quote(generator.as_ref(), flags)?;
     let mut output = Vec::<(CoinSpend, Vec<(u32, Vec<Vec<u8>>)>)>::new();
 
-    let max_blob_size =
-        max_canonical_blob_size(constants.max_block_cost_clvm, constants.cost_per_byte);
-    let program = node_from_bytes_auto(&mut a, generator, max_blob_size)?;
+    let program = if flags.contains(ConsensusFlags::INTERNED_GENERATOR) {
+        let max_blob_size =
+            max_canonical_blob_size(constants.max_block_cost_clvm, constants.cost_per_byte);
+        node_from_bytes_auto(&mut a, generator, max_blob_size)?
+    } else {
+        node_from_bytes_backrefs(&mut a, generator)?
+    };
     check_generator_node(&a, program, flags)?;
     let args = setup_generator_args(&mut a, refs, flags)?;
     let dialect = ChiaDialect::new(flags.to_clvm_flags());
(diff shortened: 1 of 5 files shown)


crates/chia-consensus/src/run_block_generator.rs
```diff
@@ -171,6 +175,14 @@ fn extract_n<const N: usize>(
 // this is required after the SIMPLE_GENERATOR fork is active
 #[inline]
 pub fn check_generator_quote(program: &[u8], flags: ConsensusFlags) -> Result<(), ValidationErr> {
+    if flags.contains(ConsensusFlags::INTERNED_GENERATOR)
+        && program.starts_with(&SERDE_2026_MAGIC_PREFIX)
```
arvidn: I got the impression that we wouldn't use this prefix in the full blocks on the blockchain, did I misunderstand?
It seems wasteful to include these bytes for every transaction block.
We wouldn't be storing them in the database at least, right?

crates/chia-consensus/src/run_block_generator.rs
```diff
@@ -223,7 +235,9 @@ where
 
     let (mut a, base_cost, program) = if flags.contains(ConsensusFlags::INTERNED_GENERATOR) {
         let mut decode_allocator = Allocator::new();
-        let program_node = node_from_bytes_backrefs(&mut decode_allocator, program)?;
+        let max_blob_size = max_canonical_blob_size(max_cost, constants.cost_per_byte);
+        let program_node = node_from_bytes_auto(&mut decode_allocator, program, max_blob_size)
+            .map_err(|_| ValidationErr::Err(ErrorCode::GeneratorRuntimeError))?;
```
arvidn: I think there's a better error code that "runtime" error.

crates/chia-consensus/src/run_block_generator.rs
```diff
@@ -223,7 +235,9 @@ where
 
     let (mut a, base_cost, program) = if flags.contains(ConsensusFlags::INTERNED_GENERATOR) {
         let mut decode_allocator = Allocator::new();
-        let program_node = node_from_bytes_backrefs(&mut decode_allocator, program)?;
+        let max_blob_size = max_canonical_blob_size(max_cost, constants.cost_per_byte);
+        let program_node = node_from_bytes_auto(&mut decode_allocator, program, max_blob_size)
```
arvidn: it really seems like a footgun to support both interned and original serialization when `INTERNED_GENERATOR` is specified. I think we should make it a consensus failure to use the original serialization after the hard fork activation.

https://github.com/Chia-Network/chia_rs/pull/1437/changes#r3589806831

arvidn (commented):
the `SIMPLE_GENERATOR` flag was part of the 2.7.0 soft fork, and has activated. Since we used generator programs in the initial version of block compression we can't require simple generators throughout the whole chain, so we still need this flag to validate early blocks.

We don't need to support (and maybe even should fail) the combination `INTERNED_GENERATOR` set and `SIMPLE_GENERATOR` not set. If there's any opportunity to simplify anything.

One crazy idea is that the interned serialization could have an implicit quote around the whole expression.

crates/chia-consensus/src/run_block_generator.rs
richardkiss: Okay, I'm fine with that. I'll change the PR to enforce it.

crates/chia-consensus/src/run_block_generator.rs
richardkiss: It is wasteful. That said, the generator is the only field in a block with a history of alternate encodings (backrefs, now this), and generator blobs often get extracted and passed around without height context. If one lands in a .bin file, `opd` and similar tools either work or choke immediately (if not upgraded to handle 2026) instead of misparsing.

It's 6 bytes per transaction block and never charged cost — the interned path charges from the deduplicated tree, so wire bytes are free. And right, we don't have to store them: a valid serde 2026 body can never begin with `0xfd` (the first magic byte), so the prefix strips and re-adds unambiguously.

What I'd propose: store the prefixed form for now, but make readers generous — accept the blob with or without the prefix. Since serde_2026 becomes obligatory anyway (your other comment, line #239ish), consensus sees exactly one encoding. If we really hate the six bytes, later we can just stop writing them (and even migrate) — no flag day needed.

crates/chia-consensus/src/run_block_generator.rs
richardkiss: I haven't had to debug stuff with these in the past, so I'll defer to you. Pick an existing error or name a new one and I'll use that. `ComplexGeneratorReceived` (148) is `INVALID_TRANSACTIONS_GENERATOR_ENCODING` in python – what do you think of that one?

richardkiss:
> the `SIMPLE_GENERATOR` flag was part of the 2.7.0 soft fork, and has activated. Since we used generator programs in the initial version of block compression we can't require simple generators throughout the whole chain, so we still need this flag to validate early blocks.
> 
> We don't need to support (and maybe even should fail) the combination `INTERNED_GENERATOR` set and `SIMPLE_GENERATOR` not set. If there's any opportunity to simplify anything.
> 
> One crazy idea is that the interned serialization could have an implicit quote around the whole expression.

We discussed this with Bram on the weekly architecture call a while back and he agreed to dropping the `(q . X)` wrapper — writing just `X` and so not running it at all, so the block is no longer a generator but carries a list of spends. 

I'd propose that as a separate change on top of this one: it deletes the quote checks for this path entirely and changes what the generator tree hash commits to, so it touches the builders too. At the same time we could rename the flag to something like `SERDE_2026_TX_LIST` since it's no longer a generator, and the interning is just how the cost is computed.

Or I can do it all in this change if you prefer, since the interaction is a bit intricate. Let me know.

I can make INTERNED_GENERATOR without SIMPLE_GENERATOR a hard failure in this PR.

crates/chia-consensus/src/run_block_generator.rs
arvidn: currently, this function is called on blocks we load from the database, so it makes it mandatory to store them. it's also called for full blocks and unfinished blocks we receive over the network, which probably is less of a problem. but there's not distinction between these two cases.

The case you're optimizing for seems much more fringe than all full nodes on the network.

crates/chia-consensus/src/run_block_generator.rs
arvidn: how about the error we use now? `Err(EvalErr::SerializationError)`

richardkiss: Done.

crates/chia-consensus/src/run_block_generator.rs
richardkiss: This is done now.

crates/chia-consensus/src/run_block_generator.rs
richardkiss: The only downside is six wasted bytes. Unless blocks remain mostly empty, this will represent a trivial amount of storage percentage-wise.

The goal is to have old parsers fast-fail on blobs they might otherwise try to parse. Without the header, they will almost certainly parse something, as nonsense, which will fail in validation down the line, far away from the real problem: the parsing point.

I think it makes sense to have this defensive protection, at least in the transition period around HF2. We can always remove the need for storing the prefix on disk later without affecting consensus since it doesn't change the tree hash.

arvidn (approved):

PR #1511, opened 2026-08-24
serde_2026 emission + trusted readers (minimal)

richardkiss:
This makes block building emit generators in serde_2026 format, and adds the readers that trusted (non-consensus) code needs to open them.

- `InternedBlockBuilder::finalize()` always emits serde_2026, so classic backrefs can be deprecated at HF2
- `solution_generator_2026()`, the serde_2026 counterpart of `solution_generator_backrefs()`
- `tree_hash_2026()` — hash a serde_2026 generator, caching common sub-trees. This is what the `generator_root` check uses for v1 blocks.
- `additions_and_removals()` dispatches on `INTERNED_GENERATOR` in the flags it already takes.
- `get_puzzle_and_solution_for_coin2`, `get_spends_for_trusted_block`, and `get_spends_for_trusted_block_with_conditions` now take the generator as a plain buffer, not a `Program`. If you're passing a `Program`, pass `bytes(program)` instead.
- `INTERNED_GENERATOR` is exposed to python, and `get_flags_for_height_and_constants` includes it at `hard_fork2_height`.

---

> [!NOTE]
> **High Risk**
> Changes consensus fork flags, generator encoding/decoding, and block-builder output at hard fork 2; incorrect dispatch or semantic mismatch would affect block validity and node behavior.
> 
> **Overview**
> Post-HF2 block building and consensus now treat **serde_2026** as the only generator wire format when **`INTERNED_GENERATOR`** is active: **`InternedBlockBuilder::finalize()`** always emits magic-prefixed serde_2026, and **`get_flags_for_height_and_constants`** turns on **`INTERNED_GENERATOR`** at **`hard_fork2_height`**.
> 
> **Parsing is explicit at call sites** instead of magic-prefix auto-dispatch: **`node_from_bytes_auto`** is removed in favor of **`node_from_bytes_2026`** (consensus size cap via **`max_canonical_blob_size`**) and **`node_from_bytes_2026_trusted`** (trusted reads after validation or local production). **`run_block_generator2`**, **`additions_and_removals`**, and the trusted **`get_coinspends_*`** paths choose classic backrefs vs serde_2026 from **`ConsensusFlags`**, not from sniffing bytes.
> 
> The Python wheel adds **`tree_hash_2026()`** (serde_2026 parse + DAG-aware **`tree_hash_cached`**), exposes **`INTERNED_GENERATOR`**, and changes generator helpers to accept raw **`ReadableBuffer`** bytes with flag-based decoding. Tests and fuzz targets cross-check interned vs classic generators on semantics and hash agreement.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a2de40266663d8a46750a73bcf60a2dc126f938. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/chia-consensus/src/build_interned_block/additional_tests.rs
```diff
@@ -329,3 +322,136 @@ fn test_byte_cost_tracking() {
...
+    }
+    (conds.spends, cost)
+}
+
+/// The builder's serde_2026 output round-trips through the
+/// INTERNED_GENERATOR consensus path and yields the same spends/conditions
+/// as an independently-built classic generator for the same bundles, run
+/// under classic rules.
+#[test]
+fn test_serde_2026_round_trip() {
```
arvidn, line 407: I don't think this is a round-trip test. it makes sure the original block builder and the new block builder both agree on the result.

Since there's no canonical serialization of serde 2026, I don't think it *does* round-trip. If we parse some serde 2026 block and then re-serualize it, we may end up with different bytes.

crates/chia-consensus/src/build_interned_block/additional_tests.rs
arvidn, line 1: I think some of these test would be better as fuzzers or proptests, to ensure fidelity between the original serialization and the new one, regardless of inputs.

crates/chia-consensus/src/build_interned_block/additional_tests.rs
```diff
@@ -329,3 +322,136 @@ fn test_byte_cost_tracking() {
...
+    let bundles = serde_2026_test_bundles();
+    let (generator_2026, _, _) = build_block(&bundles);
+    let (generator_classic, _) = build_classic_reference(&bundles);
+
+    // same parse as the wheel's tree_hash_2026()
+    let mut a = Allocator::new();
+    let node = node_from_bytes_2026_trusted(&mut a, &generator_2026)
+        .expect("node_from_bytes_2026_trusted");
+    let hash_2026 = tree_hash(&a, node);
+
```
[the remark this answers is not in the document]
arvidn: I think this is something worth looking into. a fuzzer here would help expose or disprove that an edge case can cause very expensive hashing

crates/chia-consensus/src/serde_2026.rs
```diff
@@ -301,31 +308,47 @@ mod tests {
...
 
-        // The size gate applies to non-serde_2026 formats too.
+        // The trusted reader parses the same serde_2026 blob with no cap to
+        // trip over...
+        let mut b = Allocator::new();
+        let parsed = node_from_bytes_2026_trusted(&mut b, &blob).unwrap();
+        assert_eq!(
+            node_to_bytes(&b, parsed).unwrap(),
+            node_to_bytes(&a, node).unwrap()
+        );
```
arvidn, line 341: there is a function called `node_eq_two()` in `clvm-fuzzing` that does this a bit more efficiently. It mostly matters with large or arbitrary inputs though, like fuzzers.

crates/chia-consensus/src/serde_2026.rs
```diff
@@ -301,31 +308,47 @@ mod tests {
...
         let mut b = Allocator::new();
-        assert!(matches!(
-            node_from_bytes_auto(&mut b, &classic, classic.len() - 1),
-            Err(EvalErr::SerializationError)
-        ));
+        let parsed = node_from_bytes_backrefs(&mut b, &classic).unwrap();
+        assert_eq!(
+            node_to_bytes(&b, parsed).unwrap(),
+            node_to_bytes(&a, node).unwrap()
+        );
```
arvidn, line 352: here too: `node_eq_two()`

wheel/src/api.rs
```diff
@@ -227,6 +247,46 @@ pub fn get_puzzle_and_solution_for_coin<'a>(
...
+                    NodePtr::NIL,
+                    "coin not found".to_string(),
+                )),
+                Err(ValidationErr::Eval(e)) => Err(e),
+                Ok(pair) => Ok(pair),
+            }
+        })
+        .map_err(map_pyerr)?;
+
+    // keep serializing normally, until wallets support backrefs
```
arvidn, line 281: I think we could probably drop this comment now

wheel/src/api.rs
```diff
@@ -227,6 +247,46 @@ pub fn get_puzzle_and_solution_for_coin<'a>(
...
+    Ok((
+        node_to_bytes(allocator, puzzle).map_err(map_pyerr)?.into(),
+        node_to_bytes(allocator, solution)
+            .map_err(map_pyerr)?
+            .into(),
```
arvidn, line 286: I think, technically, the serialization could happen while the GIL is released as well.

crates/chia-consensus/src/build_interned_block/additional_tests.rs
richardkiss: Oh yeah, it's wrongly named. I'll change it.

crates/chia-consensus/src/build_interned_block/additional_tests.rs
richardkiss: I'll adapt `fuzz_targets/interned-block-builder.rs` to also build classic generators and make sure the spends match.

crates/chia-consensus/src/build_interned_block/additional_tests.rs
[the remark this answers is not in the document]
richardkiss: Both fixed: the test now uses `tree_hash_cached` like `tree_hash_2026()` does, and `serde-2026-size-bound` now hashes whatever it parses so the fuzzer's timeout catches slow inputs.

crates/chia-consensus/src/serde_2026.rs
richardkiss: Okay, I'll switch to that.

wheel/src/api.rs
richardkiss: Will do.

wheel/src/api.rs
richardkiss: Okay, moving it up.

crates/chia-consensus/src/build_interned_block/additional_tests.rs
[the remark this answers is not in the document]
arvidn: I just realize now that this was in a test.. so probably didn't matter :/

wheel/src/api.rs
```diff
@@ -253,30 +313,57 @@ pub fn get_puzzle_and_solution_for_coin2<'a>(
...
+    let refs = block_refs.to_list()?.into_iter().map(|b| {
+        let buf = b
+            .extract::<PyBuffer<u8>>()
+            .expect("block_refs should be a sequence of buffers");
+        py_to_slice::<'a>(buf)
+    });
+
+    let generator =
+        node_from_bytes_2026_trusted(&mut allocator, generator_slice).map_err(map_pyerr)?;
+    let args = setup_generator_args(&mut allocator, refs, flags)?;
```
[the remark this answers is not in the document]
arvidn: this seem to suggest `py_to_slice()` is buggy. I'm pretty sure I expected the python object to stay alive after passing it to this function. It should probably be passed by reference.

crates/chia-consensus/fuzz/fuzz_targets/interned-block-builder.rs
```diff
@@ -92,6 +112,45 @@ fuzz_target!(|spends: Vec<CoinSpend>| -> Corpus {
...
+                normalized_spends(&interned_a, conds),
+                normalized_spends(&classic_a, classic_conds),
+                "interned and classic generators disagree on spends for the same spend bundle"
+            );
+        }
+        Err(e) => {
+            panic!(
+                "classic reference generator errored where the interned generator succeeded: {e:?}"
+            );
+        }
```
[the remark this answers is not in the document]
arvidn: this should probably be addressed. I imagine cursor can generate a fuzzer input that triggers this to test it. I think it's safe to assume that serde 2026 will never have higher cost, right? so this could be a special case in the fuzzer

arvidn:
@cursor review

wheel/src/api.rs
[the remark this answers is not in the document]
richardkiss: I've fixed the corresponding problem here too.

arvidn (approved):
I think we should also make the change to not require the block generator to be quoted, and just directly interpret it as a list of spends, i.e. skipping the `run_program()` call entirely. That could either be done in this PR or in another one. But I think we should include the change before cutting a release to move this into `chia-blockchain` (since that will trigger some extra work with re-generating the test chains).

PR #1515, opened 2026-09-09
Bump clvmr 0.20.0

arvidn:
also expose new function `run_chia_program_with_timeout()`

---

> [!NOTE]
> **High Risk**
> Changes mempool CLVM validation semantics (timeouts, new error codes, stricter serialization) on a security-critical path; downstream callers must pass the new timeout argument and handle error code 152/148 shifts.
> 
> **Overview**
> Upgrades **clvmr** and **clvm-fuzzing** to **0.20.0** and wires in clvmr’s timed program execution for mempool spend validation.
> 
> **`run_spendbundle`** now takes an optional wall-clock **`timeout`**; when set, each puzzle runs via **`run_program_with_timeout`** against a shared deadline (checked roughly every 1M cost). **`validate_clvm_and_signature`** requires a **`timeout`** and forwards it; the Python **`validate_clvm_and_signature`** API adds **`timeout: float`** (default infinity), with invalid values rejected and overflow clamped to **`Duration::MAX`**.
> 
> Validation errors are aligned with **chia-blockchain**: **`InvalidTransactionsGeneratorEncoding`** (148) for non-canonical CLVM bytes from clvmr, **`Timeout`** (152), and **`ComplexGeneratorReceived`** renumbered to 153. Call sites and tests pass **`None`** / **`Duration::MAX`** / **`60.0`** as appropriate; new tests cover timeout and non-canonical puzzle encoding. One **`solution_generator`** length edge-case fixture was dropped, consistent with stricter canonical serialization in 0.20.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ba3f3afdb2a07b0f076374977fe820379d464c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/wheel/python/chia_rs/chia_rs.pyi
+++ b/wheel/python/chia_rs/chia_rs.pyi
@@ -63,6 +63,7 @@ def validate_clvm_and_signature(
     max_cost: int,
     constants: ConsensusConstants,
     flags: int,
+    timeout: float = float("inf"),
 ) -> tuple[SpendBundleConditions, list[tuple[bytes32, GTElement]], float]: ...
 
 def get_conditions_from_spendbundle(
--- a/wheel/python/chia_rs/chia_rs.pyi
+++ b/wheel/python/chia_rs/chia_rs.pyi
@@ -151,9 +151,6 @@ NO_UNKNOWN_OPS: int = ...
 def run_chia_program(
     program: bytes, args: bytes, max_cost: int, flags: int
 ) -> tuple[int, LazyNode]: ...
-def run_chia_program_with_timeout(
-    program: bytes, args: bytes, max_cost: int, flags: int, timeout: float
-) -> tuple[int, LazyNode]: ...
 
 @final
 class LazyNode:
--- a/wheel/python/chia_rs/chia_rs.pyi
+++ b/wheel/python/chia_rs/chia_rs.pyi
@@ -150,6 +150,9 @@ NO_UNKNOWN_OPS: int = ...
 def run_chia_program(
     program: bytes, args: bytes, max_cost: int, flags: int
 ) -> tuple[int, LazyNode]: ...
+def run_chia_program_with_timeout(
+    program: bytes, args: bytes, max_cost: int, flags: int, timeout: float
+) -> tuple[int, LazyNode]: ...
 
 @final
 class LazyNode:
--- a/wheel/src/run_program.rs
+++ b/wheel/src/run_program.rs
@@ -62,6 +62,9 @@ pub fn run_chia_program(
 /// `timeout` is in seconds. The timeout is checked roughly every 1_000_000 cost
 /// units; programs that finish before the first check are unaffected. If the
 /// elapsed time exceeds `timeout`, raises `ValueError("timeout")`.
+///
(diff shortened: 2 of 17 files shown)


wheel/src/run_program.rs
```diff
@@ -54,3 +56,41 @@ pub fn run_chia_program(
...
+    max_cost: Cost,
+    flags: ConsensusFlags,
+    timeout: f64,
+) -> PyResult<(Cost, LazyNode)> {
+    if !(timeout.is_finite() && timeout >= 0.0) {
+        return Err(PyValueError::new_err(
+            "timeout must be a non-negative finite number of seconds",
+        ));
+    }
+    let timeout = Duration::from_secs_f64(timeout);
```
AmineKhaldi: Can we use https://doc.rust-lang.org/std/time/struct.Duration.html#method.try_from_secs_f64 here?

arvidn:
@cursor review

wheel/src/run_program.rs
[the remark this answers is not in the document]
arvidn: fixed

arvidn:
@cursor review

wheel/src/run_program.rs
```diff
@@ -54,3 +56,46 @@ pub fn run_chia_program(
...
+    timeout: f64,
+) -> PyResult<(Cost, LazyNode)> {
+    let timeout = if timeout.is_nan() || timeout < 0.0 {
+        return Err(PyValueError::new_err(
+            "timeout must be a non-negative number of seconds",
+        ));
+    } else {
+        // Overflow and +inf are clamped to Duration::MAX instead of panicking.
+        Duration::try_from_secs_f64(timeout).unwrap_or(Duration::MAX)
+    };
```
AmineKhaldi: Considering it wouldn't make sense (to me) to allow any of these cases, I would simply do something like `let timeout = Duration::try_from_secs_f64(timeout).map_err(|_| PyValueError::new_err("invalid timeout value"))?;` here.

What do you think?

arvidn: I'm not sure. I can imagine timeout values being computed, in which case this is the most reasonable way to interpret floating point values.

wheel/python/chia_rs/chia_rs.pyi
```diff
@@ -150,6 +151,9 @@ NO_UNKNOWN_OPS: int = ...
 def run_chia_program(
     program: bytes, args: bytes, max_cost: int, flags: int
 ) -> tuple[int, LazyNode]: ...
+def run_chia_program_with_timeout(
+    program: bytes, args: bytes, max_cost: int, flags: int, timeout: float
+) -> tuple[int, LazyNode]: ...
```
AmineKhaldi: Considering `validate_clvm_and_signature` accepts `timeout` now, does it still make sense to expose `run_chia_program_with_timeout` instead of just having `run_chia_program` accept timeout in a similar way?

arvidn:
@cursor review

AmineKhaldi (approved):

PR #1519, opened 2026-09-15
py_to_slice should not destruct the python object

arvidn:
> [!NOTE]
> **Medium Risk**
> Changes the Python/Rust buffer boundary on consensus block-generator paths; behavior should match but incorrect lifetimes could previously cause subtle use-after-free with memoryviews.
> 
> **Overview**
> Fixes Python buffer handling in the wheel bindings so **`memoryview`** and other exported buffers stay valid for the whole native call.
> 
> **`py_to_slice`** now takes **`&PyBuffer<u8>`** instead of owning the buffer, avoiding an early drop that could invalidate slice views. A new **`extract_buffers`** helper collects **`block_refs`** into a **`Vec<PyBuffer<u8>>`** that is kept alive while **`&[u8]`** slices are built and passed into block-generator and related APIs (`run_block_generator`, `get_spends_for_trusted_block`, `get_puzzle_and_solution_for_coin2`, etc.). Call sites were updated to use this shared path and some unnecessary lifetime/`PySequenceMethods` imports were trimmed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc07f41439ce0df775626d08253213b5428c6c49. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


arvidn:
@cursor review

arvidn:
@cursor review

richardkiss:
Claude found a potential improvement. He helped me write this comment.

`py_to_slice`'s `'a` isn't actually tied to `buf`'s lifetime, so this works because every call site is disciplined about holding the buffers until after use, not because the compiler checks it. Tying it directly would get the invariant enforced for free:

```rust
pub fn py_to_slice<'a>(buf: &'a PyBuffer<u8>) -> &'a [u8] {
    assert!(buf.is_c_contiguous(), "buffer must be contiguous");
    unsafe { std::slice::from_raw_parts(buf.buf_ptr() as *const u8, buf.len_bytes()) }
}

pub fn extract_buffers(block_refs: &Bound<'_, PySequence>) -> PyResult<Vec<PyBuffer<u8>>> {
    Ok(block_refs
        .to_list()?
        .into_iter()
        .map(|b| b.extract::<PyBuffer<u8>>().expect("block_refs must be a sequence of buffers"))
        .collect())
}
```

Each call site becomes:
```rust
let block_ref_buffers = extract_buffers(block_refs)?;
let refs: Vec<&[u8]> = block_ref_buffers.iter().map(py_to_slice).collect();

let result = py.detach(|| native_additions_and_removals(program, refs, flags, constants));
drop(block_ref_buffers);  // now optional — the compiler would catch it if moved earlier
result
```

The reason `extract_buffer_slices` couldn't do this is it returns both the owned `Vec<PyBuffer<u8>>` and slices into it from one function — the self-referential struct problem, probably why it reached for the unconstrained lifetime instead. Building the slices at the call site sidesteps that. I checked the ordering is actually enforced this way (tried dropping `buffers` before building `refs`, got `E0505: cannot move out of 'buffers' because it is borrowed`).

richardkiss (approved):

